PR #4, opened 2022-08-03
Donating ebpf based instrumentation

edeNFed:
This PR contains a copy of all the files in the https://github.com/keyval-dev/opentelemetry-go-instrumentation repository


tigrannajaryan:
Please make sure to follow the [attribution guidelines](https://github.com/open-telemetry/community/blob/main/CONTRIBUTING.md#copyright-notices). All source code files should carry the corresponding copyright notice. If you copied any files from other open-source projects that they need to carry the attribution notice. 

Also, is ebpf-based/vmlinux.h generated or it is a copy from somewhere else? If generated please include generation instruction, if copied it needs an attribution.

edeNFed:
Added. I followed the same licenses that cilium (another eBPF based CNCF project) use for their Go code / eBPF programs.

ebpf-based/Makefile
```diff
@@ -0,0 +1,21 @@
...
+BPF_INCLUDE += -I${REPODIR}/include/libbpf
+BPF_INCLUDE+= -I${REPODIR}/include
+
+.PHONY: generate
+generate: export CFLAGS := $(BPF_INCLUDE)
+generate:
+	go mod tidy
+	go generate ./...
+
+.PHONY: build
```
dineshg13, line 15: When i try to build the code , am getting an error 
```
CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o kv-go-instrumentation cli/main.go
# github.com/keyval-dev/opentelemetry-go-instrumentation/pkg/opentelemetry
pkg/opentelemetry/controller.go:119:24: undefined: otlptracegrpc
pkg/opentelemetry/controller.go:120:3: undefined: otlptracegrpc
make: *** [Makefile:17: build] Error 2
```
Can we also make a note that we need to build this on linux (it might be obvious few folks , but initially i tried it on mac which failed with a different error ). 

Also can we rename the binary from ` kv-go-instrumentation` to ` opentelemetry-go-instrumentation`

edeNFed: You can build it on Mac/Windows using the provided Dockerfile by running:
```
make docker-build IMG=otel-go-auto-inst:v0.1
```
I will document it better and change the binary name.

tigrannajaryan:
> Added. I followed the same licenses that cilium (another eBPF based CNCF project) use for their Go code / eBPF programs.

@edeNFed Thank you. To help with the review, can you please tell which files are derived/copied from elsewhere and which files are authored by yourself?

For files that are derived/copied we will need to know their original licenses, since not every license is allowed.

For files that you are the author are you the sole author or you are contributing this on behalf of other authors / on behalf of a company?

edeNFed:
@tigrannajaryan The files in `include/libbpf` directory are copied from the linux kernel: https://github.com/torvalds/linux/tree/master/tools/lib/bpf
All the other files are written by myself and are donated on behalf of Keyval.
Please let me know if there are still attribution issues.

@open-telemetry/go-instrumentation-approvers would love to get a review for this PR.

pellared:
> @tigrannajaryan The files in `include/libbpf` directory are copied from the linux kernel: https://github.com/torvalds/linux/tree/master/tools/lib/bpf All the other files are written by myself and are donated on behalf of Keyval. Please let me know if there are still attribution issues.
> 
> @open-telemetry/go-instrumentation-approvers would love to get a review for this PR.

The `libbpf` is dual-licensed. IANAL but I have no idea how we should say that we are "choosing" the `BSD-2-Clause` as `LGPL-2.1` may be problematic, because of:

> Each time you redistribute the Library (or any work based on the Library), the recipient automatically receives a license from the original licensor to copy, distribute, link with or modify the Library subject to these terms and conditions.

Any ideas?

edeNFed:
Those files are needed for easy linkage during compilation. I am not sure if this makes the decision easier but I guess we can download them during the compilation script instead of having a copy in this repo.

Anyway, I think there should be a way to use those files as other eBPF based projects in the CNCF (cilium for example) are also dependent on them.

pellared:
> Those files are needed for easy linkage during compilation. I am not sure if this makes the decision easier but I guess we can download them during the compilation script instead of having a copy in this repo.

It is not about "storing" the code but the fact that the library is linked.

Reference: https://www.gnu.org/licenses/old-licenses/lgpl-2.1.en.html

> Anyway, I think there should be a way to use those files as other eBPF based projects in the CNCF (cilium for example) are also dependent on them.

That does not guarantee that these repositories are OK from legal perspective 😉 

For instance, AFAIK, MIT requires adding a copywrite notice / EULA.

pellared:
My "perfectionist ego" forced me to create https://github.com/open-telemetry/community/issues/1280 to track

I think we should consider deferring the licensing "issue" I brought (we can create an issue for it).

At this point of time I think it is fine to use `libbpf` using the "precedence rule".

tigrannajaryan:
BSD is also explicitly allowed by Otel provided that the notice is retained which it is is, so we are good: https://github.com/open-telemetry/community/blob/main/CONTRIBUTING.md#copyright-notices

include/utils.h
```diff
@@ -0,0 +1,77 @@
...
+    }
+}
+
+static __always_inline void copy_byte_arrays(unsigned char *src, unsigned char *dst, u32 size)
+{
+    for (int i = 0; i < size; i++)
+    {
+        dst[i] = src[i];
+    }
+}
```
damemi, line 77: nit: seems like these header files are all missing a newline at eof

cli/main.go
```diff
@@ -0,0 +1,90 @@
...
+	"syscall"
+
+	"github.com/open-telemetry/opentelemetry-go-instrumentation/pkg/errors"
+	"github.com/open-telemetry/opentelemetry-go-instrumentation/pkg/instrumentors"
+	"github.com/open-telemetry/opentelemetry-go-instrumentation/pkg/log"
+	"github.com/open-telemetry/opentelemetry-go-instrumentation/pkg/opentelemetry"
+	"github.com/open-telemetry/opentelemetry-go-instrumentation/pkg/process"
+)
+
+func main() {
```
damemi, line 30: follow-up: help/usage text would be useful

pkg/process/args.go
```diff
@@ -0,0 +1,47 @@
...
+
+func (t *TargetArgs) Validate() error {
+	if t.ExePath == "" {
+		return errors.New("target binary path not specified")
+	}
+
+	return nil
+}
+
+func ParseTargetArgs() *TargetArgs {
```
damemi, line 38: any consideration for command line flags? Or at least a centralized struct for storing the various required arguments (process exe, collector endpoint, service name)

pkg/instrumentors/manager.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+}
+
+func registerInstrumentors(m *instrumentorsManager) error {
+	insts := []Instrumentor{
+		grpc.New(),
+		grpcServer.New(),
+		httpServer.New(),
+		gorillaMux.New(),
+	}
```
damemi, line 104: I like how modular this is, it looks like it will make adding support for new libraries fairly easy in the future. wonder if this could eventually be made customizable rather than hard-coded for people to support their own libraries

pkg/inject/injector.go
```diff
@@ -0,0 +1,118 @@
...
+	varsMap["is_registers_abi"] = i.isRegAbi
+	if initAlloc {
+		varsMap["total_cpus"] = i.TotalCPUs
+		varsMap["start_addr"] = i.StartAddr
+		varsMap["end_addr"] = i.EndAddr
+	}
+}
+
+func (i *Injector) getFieldOffset(libName string, libVersion string, structName string, fieldName string) (uint64, bool) {
+	for _, l := range i.data.Data {
```
damemi, line 103: From the docs:

> Luckily for us, there is a way to analyze the target binary and extract the required offsets, by using DWARF. The DWARF debug information is generated by the compiler and is stored inside the binary.
> Notice that one of our design goals is to support stripped Go binaries - meaning binaries that do not contain debug information. In order to support stripped binaries and to create a stable instrumentation, we created a library called [offsets-tracker](https://github.com/keyval-dev/offsets-tracker).

does this mean that in non-stripped binaries, we could retrieve, store, and access the offset data without needing to do this loop?

I also wonder if there are ways to optimize this data structure, maybe something we could look into if it saves a significant amount of time

damemi (approved):
@edeNFed, this is a really cool project. Separating each library the way you have done gives it a neat modular-manager design that makes this easily extensible in the future.

I left some comments for future ideas but I am +1 to merging this if there's nothing blocking us (mainly because I want to get my hands dirty and start contributing myself!)

pkg/inject/injector.go
edeNFed: Yes, I agree. this loop can probably be skipped for non-stripped binaries.
Also +1 for creating a better data structure.

pkg/process/args.go
edeNFed: Good idea, the current implementation tries to follow: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md

include/libbpf/bpf_endian.h
```diff
@@ -0,0 +1,99 @@
+/* SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause) */
```
pellared, line 1: Please copy the license files listed below to `include/libbpf`:

- https://github.com/libbpf/libbpf/blob/master/LICENSE
- https://github.com/libbpf/libbpf/blob/master/LICENSE.BSD-2-Clause
- https://github.com/libbpf/libbpf/blob/master/LICENSE.LGPL-2.1

pellared (changes requested):
Leaving just one comment to make sure that this donation is safe from the legal perspective.

pellared:
I am approving as I think that this PR is too big to process it in a regular way. I suggest merging it as it is. People can always look at this PR and add comments. We may address comments as separate PRs

dineshg13 (approved):
As discussed, we should be good to merge this .

pdelewski (approved):

PR #6, opened 2022-08-17
Donating autotel, a  source-based instrumentation

pdelewski:


dmathieu:
I don't think we can just copy donations like this.
If we want this repository to be maintainable, and its usage to be understandable by newcomers, donations need to be properly integrated as a whole, not each as a subdirectory.

pdelewski:
> I don't think we can just copy donations like this. If we want this repository to be maintainable, and its usage to be understandable by newcomers, donations need to be properly integrated as a whole, not each as a subdirectory.

Generally I agree with your argument, however both tools, I mean autotel and ebpf one, represent separate approaches. We can store them in one directory, but I'm not sure if that will change much. Do you have any ideas, suggestions?

pdelewski:
To integrate them better, we could extract some common stuff from both repos and store them in one common directory (for instance lib) and then store other unrelated functionalities in separate subdirectories and make two targets

dineshg13:
@pdelewski is review required on this PR , isn't the tool going to be merged into `https://github.com/open-telemetry/opentelemetry-go-contrib`

pdelewski:
> @pdelewski is review required on this PR , isn't the tool going to be merged into `https://github.com/open-telemetry/opentelemetry-go-contrib`

That's right, we can close this one

PR #11, opened 2022-10-19
Add GitHub action to check links

MrAlias:


.github/workflows/links.yml
```diff
@@ -0,0 +1,21 @@
...
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+    - uses: lycheeverse/lychee-action@v1.5.1
+      with:
+        args: >-
+          -v -n "*.md" "**/*.md"
+          --exclude "http://localhost*"
+    - name: fail for link errors
+      run: exit ${{ steps.lychee.outputs.exit_code }}
```
pellared, line 21:
```suggestion
        fail: true
```

see: https://github.com/lycheeverse/lychee-action#passing-arguments

.github/workflows/links.yml
```diff
@@ -0,0 +1,21 @@
...
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+
+jobs:
+  check-links:
+    runs-on: ubuntu-20.04
+    steps:
+    - uses: actions/checkout@v3
```
pellared, line 12: Do we want to pin the minor and patch version?

MrAlias: This is how it is done [elsewhere](https://github.com/open-telemetry/opentelemetry-go/blob/587437ba5b784c4ca242a8374758d0ac37fa6bd8/.github/workflows/links.yml#L15). I think it is find to pin to a major version here. All we are looking for is to ensure backwards compatibility and have the thing run.

.github/workflows/links.yml
```diff
@@ -0,0 +1,20 @@
...
+      with:
+        fetch-depth: 0
+    - uses: lycheeverse/lychee-action@v1.5.1
```
dineshg13, line 15: (This is more of observation , then an action for change) 
opentelemetry-collector-contrib  is using https://github.com/tcort/markdown-link-check , https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/eb42062694639daeed9a6b099c1d34cd8a7d9278/.github/workflows/check-links.yaml#L37 

opentelemetry-demo app is using https://github.com/open-telemetry/opentelemetry-demo/blob/d44100d0390a227475928e38239db8cc628215cc/.github/workflows/checks.yml#L51

opentelemetry-go is using `lycheeverse/lychee-action`  https://github.com/open-telemetry/opentelemetry-go/blob/11339775566920f5caa82866eb0b6b291ede9df9/.github/workflows/links.yml#L19 

I don't have experience with all of them, but it would have been better if we had consistency among projects.

.github/workflows/links.yml
pellared: I experienced a few times issues that using newer non-major GitHub Action versions were causing issues (e.g. flaky builds) and using a not pinned can make it harder to find out that the issue was related to a bug introduced in a GitHub Action.  From my experience pinning dependencies can sometimes save hours when retrospecting a problem.

We have dependabot configured for GitHub Actions so pinning a version should not be a problem while we will have more "control" on the build pipeline.

On the other hand, I have never encountered an issue like that with `actions/checkout` 😉 Still at least the maintainers get a chance to know what are the new features/bugfixes thanks to the dependabot notifications.

 This comment is not a blocker for me so feel free to resolve if you have other opinion.

.github/workflows/links.yml
dineshg13: yup , agree `lychee` looks good

dineshg13 (approved):

pellared (approved):

.github/workflows/links.yml
pellared:
```suggestion
    - uses: actions/checkout@v3.1.0
```

if you decide to pin the version 😉

PR #16, opened 2022-11-16
Commit generated bpf2go files

damemi:
Not sure if these were intentionally excluded, but this commits the generated bpf2go files.

Without these, the Instrumentors code shows up as broken in an IDE, so it will be annoying to have to `make generate` whenever working on them. Committing the generated files also allows us to track changes and put presubmits in place to enforce re-generating them when necessary.

We probably don't need to/shouldn't commit the compiled `*.o` files, so maybe we can add those to .gitignore.

Open to ideas, or please correct me if there is a reason we shouldn't commit these.


dineshg13:
Could you add command that you ran to generate these files ?

damemi:
@dineshg13 sure, I updated the commit message but this was just `make generate` from the project root.

dmathieu:
How about setting up a GH Action workflow which would try generating the files, and fail if something changes on disk (meaning the files need to be regenerated)?

damemi:
@dmathieu I can add that as part of this. Usually something like that just runs a `make generate` then `git diff` against the code in the PR, is that what you have in mind too?

dmathieu:
Yes, that's what I have in mind :)

edeNFed:
I like this idea.
Notice that the `*.o` files will be different depending on the CPU architecture. Currently, we only support x86 so that's not a problem, but we will probably need to think about it when adding ARM support.

edeNFed (approved):

pellared:
> @dmathieu I can add that as part of this. Usually something like that just runs a `make generate` then `git diff` against the code in the PR, is that what you have in mind too?

I tend to make `git diff` as a last step of a build pipeline just in case some other tool is also not making changes like `misspell -w` or `golangci-lint run --fix`.

pellared (approved):

damemi:
Based on @edeNFed's comment I removed the `.o` files and added them to gitignore. Even though it doesn't matter at the moment, taking care of that now will save headaches in the future (and it's not required to fix the IDE errors, which was my main goal)

@pellared is this good to merge? I have `The base branch restricts merging to authorized users.`. I'll leave the CI changes for @dineshg13

pellared (approved):

pellared:
>  I have The base branch restricts merging to authorized users.

@open-telemetry/go-instrumentaiton-maintainers can merge it 😉

PR #22, opened 2023-02-04
Replace bpf_ktime_get_boot_ns calls

edeNFed:
As discussed in the last SIG meeting, using `bpf_ktime_get_boot_ns()` requires a pretty new version of the kernel.
This PR replaces any invocation of `bpf_ktime_get_boot_ns()` with `bpf_ktime_get_ns()` that exists in old kernel versions.


dineshg13 (approved):
Can you please add from which Kernel version this method is supported?

pdelewski:
> Can you please add from which Kernel version this method is supported?

I think it would be good to mention reason of replacement, in the commit message or comment

MrAlias (approved):

PR #24, opened 2023-02-04
Add offsets-tracker code

edeNFed:
This PR adds the code from https://github.com/keyval-dev/offsets-tracker repo to a subdirectory called `offsets-tracker`


offsets-tracker/schema/schema.go
```diff
@@ -0,0 +1,23 @@
+package schema
+
+const FileName = "/tmp/offset_results.json"
```
dineshg13, line 3: Can we filepath configurable ?

offsets-tracker/main.go
```diff
@@ -0,0 +1,92 @@
...
+	if err != nil {
+		log.Fatalf("error in parsing version constraint: %v\n", err)
+	}
+
+	stdLibOffsets, err := target.New("go").
+		FindVersionsBy(target.GoDevFileVersionsStrategy).
+		DownloadBinaryBy(target.DownloadPreCompiledBinaryFetchStrategy).
+		VersionConstraint(&minimunGoVersion).
+		FindOffsets([]*binary.DataMember{
+			{
```
pdelewski, line 23: Would not be better to move such configuration to some json or so?

offsets-tracker/cache/cache.go
```diff
@@ -0,0 +1,70 @@
...
+	f, err := os.Open(schema.FileName)
+	if err != nil {
+		fmt.Println("could not find existing offset file, cache will be empty")
+		return nil
+	}
+
+	defer f.Close()
+	data, err := ioutil.ReadAll(f)
+	if err != nil {
+		log.Fatalf("error parsing existing offsets file: %v\n", err)
```
pdelewski, line 27: Is it ok to proceed further in case of error? (should not we return nil in this case) and the same below

offsets-tracker/target/target.go
```diff
@@ -0,0 +1,168 @@
...
+		if err == binary.ErrOffsetsNotFound {
+			fmt.Printf("%s: could not find offsets for version %s\n", t.name, v)
+		} else {
+			result.ResultsByVersion = append(result.ResultsByVersion, &VersionedResult{
+				Version:    v,
+				OffsetData: res,
+			})
+		}
+
+		f.Close()
```
pdelewski, line 117: Why not use defer statement?

offsets-tracker/downloader/go_wrapper.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	_ "embed"
+	"fmt"
+	"github.com/keyval-dev/offsets-tracker/utils"
+	"io/fs"
+	"io/ioutil"
+	"path"
+)
+
+const appName = "testApp"
```
damemi, line 12: this conflicts with the [module name](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/24/files#diff-1be04fea3040593d2f2a359560ac8dea215e83f826aaee271db55f61b0c026b5R1) (`testapp`) and I got this error running it:
```
2023/02/07 18:04:58 error while fetching offsets: open /tmp/testApp1757356936/testApp: no such file or directory
exit status 1

$ ls /tmp/testApp1757356936/
go.mod   go.sum   main.go  testapp
```

dineshg13: I am getting the same error.

edeNFed:
I think I resolved all the issues.
Also, I added a Dockerfile so you can now run the tool via:
```bash
docker build -t dev/offset-tracker:dev .
docker run -it --rm dev/offset-tracker:dev --output ./offsets.json
```

@pdelewski I totally agree with changing the configuration to some JSON instead of code in the main function, I will do it as a separate PR.

Please let me know if there are additional open issues.

pdelewski (approved):

dineshg13 (approved):
LGTM. Am able to generate `offset_resutlts.json`. As discussed in SIG, we will make further improvements in a separate PRs.

Issue #17, opened 2022-11-16
panic: cant find keyval map


damemi:
### Describe the bug

Running `otel-go-instrumentation` on a simple binary produces the following failure:
```
{"level":"info","ts":1668622016.422307,"caller":"cli/main.go:37","msg":"starting Go OpenTelemetry Agent ..."}
{"level":"info","ts":1668622016.422393,"caller":"opentelemetry/controller.go:107","msg":"Establishing connection to OpenTelemetry collector ..."}
{"level":"info","ts":1668622018.4382257,"caller":"process/discover.go:57","msg":"found process","pid":157218}
panic: cant find keyval map

goroutine 1 [running]:
github.com/open-telemetry/opentelemetry-go-instrumentation/pkg/process.(*processAnalyzer).findKeyvalMmap(0xb48c60?, 0xc0001328c0?)
	/usr/local/google/home/mikedame/go/src/github.com/open-telemetry/opentelemetry-go-instrumentation/pkg/process/analyze.go:92 +0x14f
github.com/open-telemetry/opentelemetry-go-instrumentation/pkg/process.(*processAnalyzer).Analyze(0xc000482120?, 0x26622, 0x2?)
	/usr/local/google/home/mikedame/go/src/github.com/open-telemetry/opentelemetry-go-instrumentation/pkg/process/analyze.go:118 +0x1a9
main.main()
	/usr/local/google/home/mikedame/go/src/github.com/open-telemetry/opentelemetry-go-instrumentation/cli/main.go:74 +0x37f
```

### Environment

- sample app built with go 1.18 (using `golang:1.18` docker container)

### To Reproduce

I am building the following code ([from here](https://gowebexamples.com/routes-using-gorilla-mux/):
```go
package main

import (
	"fmt"
	"log"
	"net/http"

	"github.com/gorilla/mux"
)

func main() {
	r := mux.NewRouter()

	r.HandleFunc("/books/{title}/page/{page}", func(w http.ResponseWriter, r *http.Request) {
		vars := mux.Vars(r)
		title := vars["title"]
		page := vars["page"]

		fmt.Fprintf(w, "You've requested the book: %s on page %s\n", title, page)
	})

	log.Fatal(http.ListenAndServe(":8080", r))
}
```

Save, run `go mod init` and `go tidy` then build with `go build -o gm main.go`.

Run with `./gm &`

Start a local OTEL collector at `localhost:4317`

[...]

PR #32, opened 2023-03-01
Remove dependency to KeyVal launcher when it is not needed

mariomac:
The GRPC instrumentors require to expose a writable memory map, so the [KeyVal launcher](https://github.com/keyval-dev/launcher) is needed.

However, for HTTP instrumentation don't require it, so the auto-instrumenter process should not fail if that mmap is not available. This will simplify the deployment in many scenarios.

The PR includes an example deployment. However it won't still provide metrics because the HTTP instrumenter needs to be fixed (will do it in another PR).

This should help with issue #17


pkg/process/analyze.go
```diff
@@ -86,10 +85,14 @@ func (a *processAnalyzer) findKeyvalMmap(pid int) (uintptr, uintptr) {
...
 			log.Logger.Info("found addr of keyval map", "addr", m.StartAddr)
-			return m.StartAddr, m.EndAddr
+			return &AllocationDetails{
+				StartAddr: uint64(m.StartAddr),
+				EndAddr:   uint64(m.EndAddr),
+			}
 		}
 	}
-	panic(errors.New("cant find keyval map"))
+	log.Logger.V(1).Info("did not find any address for keyval map. Some services couldn't be instrumented")
```
edeNFed, line 94: I think a better message would be something like `could not allocate remote memory, automatic context propagation will be disabled`

edeNFed (approved):
lgtm

deployments/grafana-demo/README.md
```diff
@@ -0,0 +1,23 @@
+# How to deploy
```
MrAlias, line 1: Can we split this demo example to its own PR?

mariomac: Done! Here is the PR: https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/37

MrAlias (approved):

Issue #31, opened 2023-02-28
Publish an official OpenTelemetry image


TylerHelmuth:
**Is your feature request related to a problem? Please describe.**
I am looking for an official OpenTelemetry image to use instead of `keyval/otel-go-agent:latest` but none exists yet.

**Describe the solution you'd like**
This repository publishes an image via Github/Dockerhub like other OpenTelemetry owned images.

**Describe alternatives you've considered**

**Additional context**
I feel a Community-owned image is required for https://github.com/open-telemetry/opentelemetry-operator/issues/908

TylerHelmuth:
This is something I am invested in and can help as needed.

edeNFed:
Opened: https://github.com/open-telemetry/community/issues/1377

PR #33, opened 2023-03-01
[admin] Add release workflow

TylerHelmuth:
Adds a new `release.yaml` workflow that is triggered on the push of new tags.  Shamelessly stolen and modified from https://github.com/keyval-dev/opentelemetry-go-instrumentation/blob/master/.github/workflows/release.yml.

I am woefully underqualified to name this image, but I went with `otel/autoinstrumentation-go` for this PR.  I reviewed the [existing otel images on DockerHub](https://hub.docker.com/u/otel), the [images published by the operator repo](https://github.com/orgs/open-telemetry/packages?repo_name=opentelemetry-operator), and the the java agent .jar name (`opentelemetry-javaagent.jar`).  Someone with authority should name this image because we don't want to rename it in the future.

Tested via my fork.  Published images can be found here:
- https://hub.docker.com/r/thelmuth34/autoinstrumentation-go/tags
- https://github.com/TylerHelmuth/opentelemetry-go-instrumentation/pkgs/container/opentelemetry-go-instrumentation%2Fautoinstrumentation-go

Closes https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/31.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,45 @@
+name: Release
+on:
+  push:
+    tags:
+      - 'v*.*.*'
+jobs:
+  build-and-push-image:
+    env:
+      IMAGE_NAME: autoinstrumentation-go
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+      - name: Log in to Docker Hub
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+      - name: Log in to the Container registry
+        uses: docker/login-action@v2
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Extract metadata (tags, labels) for Docker
+        id: meta
+        uses: docker/metadata-action@v4
+        with:
+          images: |
+            otel/${{ env.IMAGE_NAME }}
+            ghcr.io/${{ github.repository }}/${{ env.IMAGE_NAME }}
+      - name: Build and push
+        uses: docker/build-push-action@v4
+        with:
+          push: true
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
+          platforms: linux/amd64
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -13,7 +13,7 @@ jobs:
       packages: write
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v3
+        uses: actions/checkout@v3.3.0
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v2
       - name: Set up QEMU


edeNFed:
Can we start with just amd64 support?
ARM support requires many code changes that need to be merged before we can actually build container images for ARM processors. (already have it in the keyval repo, I will open a separate PR for that soon).

TylerHelmuth:
| Can we start with just amd64 support?

Updated

MikeGoldsmith (approved):
Looks good to me 👍🏻

edeNFed (approved):
lgtm

TylerHelmuth:
Added an agenda item to next week's go auto-instrumentation SIG meeting to finalize acceptance from the maintainers on publishing an image and image name.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,27 @@
...
+      - name: Login to DockerHub
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+      - name: Build and push
+        uses: docker/build-push-action@v2
+        with:
+          push: true
+          tags: otel/opentelemetry-go-agent:${{ steps.vars.outputs.tag }}
```
pellared, line 26: I suggest avoiding the "Agent" name as it brings confusion. More https://github.com/open-telemetry/opentelemetry.io/issues/1689

can we use `auto` or maybe even `auto-instrumentation` instead?

pellared, line 26: It may be good to use https://github.com/docker/metadata-action

Example usage: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/main/.github/workflows/build_and_push_image.yml

Also you can see https://docs.github.com/en/actions/publishing-packages/publishing-docker-images

.github/workflows/release.yaml
```diff
@@ -0,0 +1,27 @@
...
+    steps:
+      - name: Set env
+        id: vars
+        run: echo ::set-output name=tag::${GITHUB_REF#refs/*/}
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+      - name: Login to DockerHub
+        uses: docker/login-action@v2
```
pellared, line 18: I think we can simply push it to GitHub registry

pellared (commented):
I suggest testing it on a fork before doing any merge.

.github/workflows/release.yaml
TylerHelmuth: If the community like including `auto`, we could mimic the operator's style and use `otel/autoinstrumentation-go`.

.github/workflows/release.yaml
TylerHelmuth: I wouldn't excluding pushing to GitHub registry, but there is precedence for using DockerHub for OTel images.  It seems the community is using both locations (operator only used GitHub, demo and collector publish to both).  As long as I have an image to use I'll be happy so I don't care where it lives.  If starting with only Github is easier then that works for me.

pellared: I just have no idea if the secrets are available.

TylerHelmuth: https://github.com/open-telemetry/community/issues/1377 is completed so the secrets in `DOCKER_USERNAME` and `DOCKER_PASSWORD` should be available to us if we chose to use them.

.github/workflows/release.yaml
TylerHelmuth: Where is the .Net image being pushed?  I don't see it in GitHub or DockerHub.

TylerHelmuth: Updated name to `autoinstrumentation-go`.

TylerHelmuth: That action is amazing, thanks for pointing it out.

.github/workflows/release.yaml
TylerHelmuth: Updated to push to both.

TylerHelmuth:
@pellared thank you for the useful examples.  I have update the action to push to both ghcr.io and DockerHub as directed by https://docs.github.com/en/actions/publishing-packages/publishing-docker-images#publishing-images-to-docker-hub-and-github-packages.  I was able to test this using my fork, images can be found here:
- https://hub.docker.com/r/thelmuth34/autoinstrumentation-go/tags
- https://github.com/TylerHelmuth/opentelemetry-go-instrumentation/pkgs/container/opentelemetry-go-instrumentation%2Fautoinstrumentation-go

@MikeGoldsmith @edeNFed @pellared please re-review.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,45 @@
...
+  build-and-push-image:
+    env:
+      IMAGE_NAME: autoinstrumentation-go
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
```
pellared, line 16: Can we pin all GitHub Actions' verions?\
E.g. `actions/checkout@v3.3.0`.
This would follow the same pattern as in https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/main/.github/workflows/links.yml

Dependabot is making a PRs to bump and we would mitigate the possible risk if a new version of a GitHub Action would contain a bug.

pellared (approved):

.github/workflows/release.yaml
MrAlias, line 16:
```suggestion
        uses: actions/checkout@v3.3.0
```

PR #34, opened 2023-03-02
Fix net/http instrumentation for register-based ABI

mariomac:
Makes the stdlib net/http instrumentor to work on register-based ABIs.

Previously, it used the request context pointer as a key for the requests map. With register-based ABI, you can get the arguments in the start uprobe but not in the return uprobes, because the CPU registers might have changed at this point of the execution.

This PR changes the pointer to the context by a pointer to the actual Goroutine that executes the handle, which is guaranteed to be always in the CPU register 14 according to the [Go ABI specification](https://go.googlesource.com/go/+/refs/heads/dev.regabi/src/cmd/compile/internal-abi.md#amd64-architecture).

I have also observed that, at least in recent versions of Go, eBPF probes attached to `net/http.(*ServeMux).ServeHTTP` are not triggered, so I added a new `net/http.HandlerFunc.ServeHTTP` function reference and make sure that hooks are added to all the functions.


edeNFed (approved):

edeNFed:
lgtm

pkg/instrumentors/bpf/net/http/server/probe.go
```diff
@@ -102,41 +97,56 @@ func (h *httpServerInstrumentor) Load(ctx *context.InstrumentorContext) error {
...
+	offset, err := ctx.TargetDetails.GetFunctionOffset(funcName)
+	if err != nil {
+		logger.V(1).Info("could not find function start offset. Skipping",
+			"error", err.Error())
+		return
+	}
+	retOffsets, err := ctx.TargetDetails.GetFunctionReturns(funcName)
+	if err != nil {
+		logger.V(1).Info("could not find function end offsets. Skipping",
+			"error", err.Error())
```
MrAlias, line 124:
```suggestion
		logger.Error(err, "could not find function end offsets. Skipping")
```

pkg/instrumentors/bpf/net/http/server/probe.go
```diff
@@ -102,41 +97,56 @@ func (h *httpServerInstrumentor) Load(ctx *context.InstrumentorContext) error {
...
+
+	return nil
+}
+
+func (h *httpServerInstrumentor) registerProbes(ctx *context.InstrumentorContext, funcName string) {
+	logger := log.Logger.WithName("net/http-instrumentor").WithValues("function", funcName)
+	offset, err := ctx.TargetDetails.GetFunctionOffset(funcName)
+	if err != nil {
+		logger.V(1).Info("could not find function start offset. Skipping",
+			"error", err.Error())
```
MrAlias, line 118:
```suggestion
		logger.Error(err, "could not find function start offset. Skipping")
```

pkg/instrumentors/bpf/net/http/server/probe.go
```diff
@@ -102,41 +97,56 @@ func (h *httpServerInstrumentor) Load(ctx *context.InstrumentorContext) error {
...
+	h.uprobes = append(h.uprobes, up)
 
 	for _, ret := range retOffsets {
 		retProbe, err := ctx.Executable.Uprobe("", h.bpfObjects.UprobeServerMuxServeHTTP_Returns, &link.UprobeOptions{
 			Offset: ret,
 		})
 		if err != nil {
-			return err
+			logger.V(1).Info("could not insert return uprobe. Skipping",
+				"error", err.Error())
```
MrAlias, line 145:
```suggestion
			logger.Error(err, "could not insert return uprobe. Skipping")
```

MrAlias (approved):

damemi:
@mariomac would this change be relevant to the gorillamux instrumentor too?

PR #44, opened 2023-03-21
[chore] Add licenses/notices for all dependencies

TylerHelmuth:
This PR is a best effort attempt at ensuring our repository includes all the Licenses/Notices of the dependencies we use in our distribution.  The Licenses/Notices are gathered via [go-licenses](https://github.com/google/go-licenses) which allows for a repeatable workflow and a way to check the LICENSES with CI. 

The licenses were gathered using `go-licesnes save cli --save_path LICESNES`.  Since the tool cannot handle non-go licenses, the libbpf license and code is copied from `include` into the `LICESNES` folder.


Makefile
```diff
@@ -23,3 +23,26 @@ docker-build:
 .PHONY: offsets
 offsets:
 	cd offsets-tracker; OFFSETS_OUTPUT_FILE="../pkg/inject/offset_results.json" go run main.go
+
+.PHONY: install-tools
+install-tools:
+	go install github.com/google/go-licenses@latest
```
pellared, line 29: Can we use [the standard `go.mod` + `tools.go` pattern](https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module) to pin the tooling dependencies.

You can check how it is done in other Go-based OTel repositories:
- https://github.com/open-telemetry/opentelemetry-go-build-tools
- https://github.com/open-telemetry/opentelemetry-go
- https://github.com/open-telemetry/opentelemetry-collector

We should also add an entry in `dependabot.yml` to have automated PRs for bumping the dependencies.

Makefile
```diff
@@ -23,3 +23,26 @@ docker-build:
...
+	go-licenses save ./cli --save_path temp; \
+	cp -R ./include/libbpf ./temp; \
+    if diff temp LICENSES > /dev/null; then \
+      echo "Passed"; \
+      rm -rf temp; \
+    else \
+      echo "LICENSES directory must be updated. Run make update-licenses"; \
+      rm -rf temp; \
+      exit 1; \
+    fi; \
```
pellared, line 48: We can simply run `git diff` after `update-licences` in CI to validate if there are no missing changes.

Something like (not tested!):

```make
.PHONY: gitdiff
gitdiff:
          git status
          git diff
          test -z "$(shell git status --porcelain)"
```

We should also have a GitHub workflow to make sure it is up to date.

pellared (commented):
I think it would be good to mention in the main `README.md` (in License section) that third-party licenses and copywrite notices are stored under `LICENSES` directory.

Makefile
```diff
@@ -23,3 +23,26 @@ docker-build:
...
+
+.PHONY: install-tools
+install-tools:
+	go install github.com/google/go-licenses@latest
+
+.PHONY: update-licenses
+update-licenses: install-tools
+	rm -rf LICENSES; \
+	go-licenses save ./cli/ --save_path LICENSES; \
+	cp -R ./include/libbpf ./LICENSES
```
pellared, line 35: Why not?

```suggestion
	rm -rf LICENSES
	go-licenses save ./cli/ --save_path LICENSES
	cp -R ./include/libbpf ./LICENSES
```

Otherwise if `go-licenses save ./cli/ --save_path LICENSES` crashes then this will silently pass

Makefile
pellared: I am not sure if I follow. You can have a commit that bumps the dependency (via dependabot) and then you add commit that updates the licenses (if they changed) or you execute `make update-licences` after adding a new dependency. You can also do both in one commit.

Anyway. I do not have a strong opinion. The good thing about my proposal that it would be also compatible with other "code generation" tools e.g. `go generate`.

Makefile
TylerHelmuth: I did noticed this repo wasn't doing this yet.  As this is the first tool I can add these building blocks as part of this PR.

Makefile
TylerHelmuth: You're right, I was mixing the generation and verification into one step.  With your method you could bump the dependency, then run `make updated-licenses`, commit all of that, and then run `make verify-licenses`, which would run `update-licenses` and `git-diff`, erroring if there are any changes.

So I think it comes down to how much the failed error messages matters vs compatibility with other tools like `go generate`.

TylerHelmuth: Testing this out I have instantly run into `git diff` seeing the `.DS_Store` file on my mac :/

Makefile
TylerHelmuth: Added

Makefile
pellared: > Testing this out I have instantly run into `git diff` seeing the `.DS_Store` file on my mac :/

side note: I encourage you to add `.DS_Store` to you global git ignore 😉 See: https://gist.github.com/subfuzion/db7f57fff2fb6998a16c

internal/tools/go.mod
pellared, line 1: Can you add dependabot config for this?

Makefile
TylerHelmuth: That is an excellent idea, thanks for sharing; implemented immediately.  But I don't think we can rely on every contributor having done that.

internal/tools/go.mod
TylerHelmuth: Added.

Makefile
TylerHelmuth: Somehow I missed your second comment 👍

pellared (approved):

TylerHelmuth:
Pinging @open-telemetry/go-instrumentaiton-maintainers

LICENSES/github.com/hashicorp/go-version/README.md
MrAlias, line 1: This pulled in the complete repository. Was this intentional?

LICENSES/libbpf/bpf.h
MrAlias, line 1: Similar here. No sure this code should live in this repository here.

LICENSES/github.com/hashicorp/go-version/README.md
TylerHelmuth: The `go-licenses` tool made that decision, so I assume it has something to do with the `Mozilla Public License 2.0` that repo uses.

LICENSES/libbpf/bpf.h
TylerHelmuth: This one was done manually by the `update-licenses` make target.  We keep the code in `include/libbpf`, and I wasn't sure what was needed to be compliant, so I wrote the target to copy the whole folder.  If you think only the three LICENSE* files need included I can update the target.

LICENSES/github.com/hashicorp/go-version/README.md
MrAlias: Gotcha. I didn't realize it was MPL2

PR #45, opened 2023-03-23
Reformat offsets_result.json for brevity and better human visibility

mariomac:
This PR proposes a change in the format of the `offsets_restuls.json` files:

* Does not store all the versions from all the libraries but only the versions that change a given offset through its predecessor.
* It replaces some arrays by maps so you can lookup directly structs and fields without having to iterate.

The arguable advantages:

* Shorter length of the file (from 6K lines to 182).
* Resulting file is more "human-friendly". Being simpler to visually locate offsets and changes across versions.
* Most times will work when there is a new version of a library and the offsets haven't been yet updated. Most of the times the offset will remain the same, and the current version of Go or a Library will match to the newest library in the file.


offsets-tracker/go.mod
```diff
@@ -1,4 +1,4 @@
-module github.com/keyval-dev/offsets-tracker
+module github.com/open-telemetry/offsets-tracker
 
 go 1.17
```
dineshg13, line 3: outside of this PR, can we bump this to latest version?

MrAlias:
@mariomac any chance you can sync with main, or would you want to recreate this?

mariomac:
@MrAlias done!

damemi:
@mariomac offsets check failed, looks like it's partially due to https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/84 but also some go 1.20.2/1.20.3 diffs. Can you try running `make docker-offsets`?

mariomac:
@damemi Ooops... sorry I forgot it!

MrAlias (approved):

MikeGoldsmith (approved):

Issue #39, opened 2023-03-13
Add CI with a sample app to check that changes are valid


damemi:
We should add CI with a sample app to check that changes (such as the generated offsets) are valid. I don't see an issue for that yet so making one

_Originally posted by @damemi in https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/38#pullrequestreview-1337034592_

damemi:
I'd like to work on this if you can assign to me

PR #46, opened 2023-03-24
Add fixture-based e2e tests with sample apps

damemi:
Fixes https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/39

This adds an e2e test matrix (largely based on [collector-contrib's tests](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/a45376db04289be6e0ca98276d26d8b504d9cad5/.github/workflows/e2e-tests.yml)) that does the following:

1. Builds the auto-instrumentation agent, launcher (to be removed with https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/40), and sample apps using each supported library
2. Spins up a kind cluster
3. Starts an otel collector with trace receiver and file exporter
4. Starts the sample app for that library
5. The sample app calls itself to generate trace data
6. Copies the file exporter output to a local fixture
7. Checks for changes to the current fixture, ignoring/removing timestamps

I used a k8s cluster instead of just running the binaries for a couple reasons: First, I wanted to just use a collector image from upstream rather than building it ourselves. Second, k8s gives some nice orchestration around starting up and stopping these processes that works well as a Github action. I couldn't find a good way to build local binaries and have those run in the background, while moving on to the next Workflow step, and checking that everything had completed successfully.

The steps are basically copied to a new `make fixture-$LIBRARY` target that lets you generate these fixtures. Eg, `make fixture-nethttp` will run all the steps from above. Failures in these tests could indicate a legitimate update to the fixtures, so updating them locally will be useful.

I added sample apps for gorillamux and net/http. If someone else wants to add a gRPC app, I think the first 2 serve as a good example. New libraries should be able to drop in similarly.

### DIFF
--- a/.github/workflows/kind.yml
+++ b/.github/workflows/kind.yml
@@ -1,0 +1,83 @@
+name: e2e-tests
+
+on:
+  push:
+    branches:
+      - main
+    tags:
+      - 'v[0-9]+.[0-9]+.[0-9]+*'
+  pull_request:
+
+jobs:
+  kubernetes-test:
+    strategy: 
+      matrix:
+        k8s-version: ["v1.26.0"]
+        library: ["gorillamux", "nethttp"]
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v3
+      - name: Setup Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: 1.19   
+      - name: Build auto-instrumentation
+        timeout-minutes: 5
+        run: |
+          IMG=otel-go-instrumentation:latest make docker-build
+      - name: Build launcher #TODO: Remove launcher steps once pr/40 merges
+        run: |
+          git clone https://github.com/keyval-dev/launcher.git
+          cd launcher
+          docker build -t kv-launcher:latest .
+      - name: Build sample app
+        run: |
+          cd test/e2e/${{ matrix.library }}
+          docker build -t sample-app:latest .
+      - name: Set up Helm
+        uses: azure/setup-helm@v3.5
+        with:
+          version: v3.9.0
+      - name: Create kind cluster
+        uses: helm/kind-action@v1.5.0
+        with:
+          node_image: kindest/node:${{ matrix.k8s-version }}
+          kubectl_version: ${{ matrix.k8s-version }}
+      - name: Check kind
+        run: |
+          kubectl cluster-info --context kind-chart-testing
+          kubectl get node
+          docker ps -a --filter label=io.x-k8s.kind.cluster=chart-testing
+      - name: Kind load images
+        run: |
+          kind load docker-image otel-go-instrumentation --name chart-testing
+          kind load docker-image kv-launcher --name chart-testing
+          kind load docker-image sample-app --name chart-testing
+      - name: Add Dependencies
+        shell: bash
+        run: |
+            helm repo add open-telemetry https://open-telemetry.github.io/opentelemetry-helm-charts
+      - uses: actions/checkout@v3
+        with:
+          repository: 'open-telemetry/opentelemetry-helm-charts'
+          path: opentelemetry-helm-charts
+      - name: Helm install collector
+        run: helm install test -f .github/workflows/e2e/k8s/collector-helm-values.yml opentelemetry-helm-charts/charts/opentelemetry-collector
+      - name: check collector status
+        run: |
+          kubectl wait --for=condition=Ready --timeout=60s pod/test-opentelemetry-collector-0
+      - name: start sample job
+        run: |
+          kubectl -n default create -f .github/workflows/e2e/k8s/sample-job.yml
+      - name: check job status
+        run: |
+          kubectl wait --for=condition=Complete --timeout=60s job/sample-job
+      - name: copy telemetry trace output
+        run: |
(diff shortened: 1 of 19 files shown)


test/e2e/gorillamux/traces.json
```diff
@@ -0,0 +1,52 @@
...
+                  "value": {
+                    "stringValue": ""
+                  }
+                }
+              ],
+              "kind": 2,
+              "parentSpanId": "",
+              "spanId": "",
+              "status": {},
+              "traceId": ""
```
damemi, line 45: I was surprised that spanId and traceId were empty here. ping @edeNFed does that look right?

Makefile
```diff
@@ -58,3 +58,34 @@ verify-licenses: | $(GOLICENSES)
...
+.PHONY: fixture-gorillamux
+fixture-gorillamux:
+	LIBRARY=gorillamux $(MAKE) fixture
+
+.PHONY: fixture-nethttp
+fixture-nethttp:
+	LIBRARY=nethttp $(MAKE) fixture
+
+.PHONY: fixture
+fixture:
```
MrAlias, line 71: This will add a `fixture` target that shouldn't be called (`LIBRARY` will not be set if it is directly called). Instead, we could compose the built target

```suggestion
.PHONY: fixture-nethttp fixture-gorillamux
fixture-nethttp: fixtures/nethttp
fixture-gorillamux: fixtures/gorillamux
fixture/%: LIBRARY=$*
fixture/%:
```

.github/workflows/kind.yml
```diff
@@ -0,0 +1,83 @@
...
+        k8s-version: ["v1.26.0"]
+        library: ["gorillamux", "nethttp"]
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v3
+      - name: Setup Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: 1.19
```
MrAlias, line 24:
```suggestion
          go-version: "1.20"
```

test/e2e/gorillamux/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM golang:1.19
```
MrAlias, line 1:
```suggestion
FROM golang:1.20
```

test/e2e/nethttp/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM golang:1.19
```
MrAlias, line 1:
```suggestion
FROM golang:1.20
```

test/e2e/gorillamux/traces.json
```diff
@@ -0,0 +1,52 @@
...
+          "scope": {
+            "name": "github.com/gorilla/mux"
+          },
+          "spans": [
+            {
+              "attributes": [
+                {
+                  "key": "http.method",
+                  "value": {
+                    "stringValue": ""
```
MrAlias, line 31: It also seems weird that these values are empty (not "Get" for example).

pellared (commented):
This is a comment rather for future improvment. I do not want to block this PR as we need to start with something 😉 

Based on my experience in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation (where I am a maintainer) I find testing against a  "test fake collector" makes the tests more maintainable:

- spinning up a a `httptest` server or by setting up a gRPC server in tests is faster than running a real collector [image]
- you get more flexibility in assertion logic


Here you can find an example implementation and usage of a "test fake collector" in Go: https://github.com/signalfx/splunk-otel-go/blob/main/distro/otel_test.go

damemi:
@pellared that's a great suggestion, but I agree on not blocking this PR right now (especially because I want to unblock #40). I'll make a new issue from your comment to track that for a refactor in the future.

Right now I just want to figure out the empty labels from above which I'm trying a couple things for. Maybe an issue with context propagation from the request?

test/e2e/gorillamux/traces.json
damemi: Ah I didn't enable a propagator. Setting `OTEL_PROPAGATORS` on the sample app fills these in (and the request fields too).

test/e2e/gorillamux/traces.json
damemi: https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/46#discussion_r1152142583

test/e2e/gorillamux/traces.json
damemi: hm but this empty span is still here

damemi:
It looks like https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/55 broke auto-instrumentation due to a change in [v0.9.1](https://github.com/cilium/ebpf/releases/tag/v0.9.1) that updated `UprobeOptions.Offset` to be relative, with `Address` now being absolute. Updated this in https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/46/commits/dddbb7e637537506c241d9988d43674c4019be51 and running e2es locally now seem to work again, hopefully they pass in CI too

damemi:
They're green! 🎉 🎉

test/e2e/gorillamux/traces.json
damemi: I think this empty span is a bug in the auto-instrumentation, but that should be a separate issue to track instead of fix in this PR

MrAlias (approved):

edeNFed (approved):

PR #70, opened 2023-04-05
Generate 1.20.3 offsets and use docker for offsets CI

damemi:
go 1.20.3 was just released, and with that our offsets needed to be updated. This caused the `path` field to be empty, such as in https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/68#issuecomment-1497571408

This PR adds an offsets generation presubmit check to give us more of a signal into these kinds of failures. We already have the automated weekly check, but with more frequent changes and mid-week releases of Go it would be good to have this on a PR presubmit too.

This also uses the docker `golang` image for generating these offsets (in both the presubmit and automated update). This makes sure we are running and checking against the same version of go, because while the `golang:1.20` image uses `go1.20.3`, the setup-go github action still uses `1.20.2`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,6 +40,10 @@ docker-build:
 offsets:
 	cd offsets-tracker; OFFSETS_OUTPUT_FILE="../pkg/inject/offset_results.json" go run main.go
 
+.PHONY: docker-offsets
+docker-offsets:
+	docker run --rm -v $(shell pwd):/app golang:1.20 /bin/sh -c "cd ../app && make offsets"
+
 .PHONY: update-licenses
 update-licenses: | $(GOLICENSES)
 	rm -rf LICENSES
--- a/pkg/inject/offset_results.json
+++ b/pkg/inject/offset_results.json
@@ -5,6 +5,1464 @@
       "data_members": [
         {
           "struct": "net/http.Request",
+          "field_name": "Method",
+          "offsets": [
+            {
+              "offset": 0,
+              "version": "1.20.3"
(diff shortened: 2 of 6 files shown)


TylerHelmuth (approved):

pkg/inject/offset_results.json
edeNFed, line 1: @damemi  Maybe it's worth excluding the actual offset changes, and letting the GitHub workflow create a new PR? Might be a good chance to validate that it works properly.

Makefile
```diff
@@ -40,6 +40,10 @@ docker-build:
 offsets:
 	cd offsets-tracker; OFFSETS_OUTPUT_FILE="../pkg/inject/offset_results.json" go run main.go
 
+.PHONY: docker-offsets
+docker-offsets:
+	docker run --rm -v $(shell pwd):/app golang:1.20 /bin/sh -c "cd ../app && make offsets"
```
MrAlias, line 45: This looks like it is going to be brittle given when a new version of Go comes out this will need to be manually bumped to allow the new generation, right?

MrAlias (approved):

Makefile
damemi: Yeah, but that's true for our other workflows that use go (and our dockerfiles/go.mod).

If we want one less place to have to update with new minor version releases, we could use `golang:latest` here. But then there will be skew if the `latest` image updates before we update our own builds. So pinning to the latest minor version we've tested on seems reasonable.

PR #91, opened 2023-04-24
Add net/http client instrumentor

oxeye-gal:
Hi,

This PR attempts to add instrumentation for net/http client.

We do it by writing a fake tophash value (as calculating the correct one will be harder to achieve) to the first headers map bucket and writing the "traceparent" header key and value into memory.

This was tested with Go versions 1.12-1.19.

Our test environment consisted of instrumented Python application that sends a request to instrumented Go http server which in turn sends an http request to an additional instrumented Go http server.

The produced trace looks like the following:

213374473-d266bf1f-49f5-4f8c-8e0d-53ed51d80201

We would love to get your feedback on this.
Thanks!

PR #86, opened 2023-04-19
Fix missing spans in gorillamux instrumentation

JamieDanielson:
## Which problem is this PR solving?

- closes #81 

## Short description of the changes

~The [first commit](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/86/commits/b2d441b0474713f75f55aff0294bd2f2bb1929a4) updates the expected output of a trace for nethttp instrumentation, removing the empty second span. The next set of commits adjusts the http server instrumentation, ultimately just adjusting one of the changes in #34 that used an array of two function names for `ServeHttp` and keeping only the newer `"net/http.HandlerFunc.ServeHTTP"`.~
- Another [commit](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/86/commits/c84373173e5d576ce5c69c3bada09b4d85c5d5fb) updates the expected output of a trace for gorillamux instrumentation, updating the empty gorilla/mux span to include the http info.
- And finally, the gorillamux instrumentation was updated to match the nethttp/server changes made in #34
- Updated [per PR feedback](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/86#discussion_r1188868104): Method Length has been set to 7 instead of 6 to include items like OPTIONS. This has been adjusted for gin-gonic and nethttp server instrumentation as well.

I'm not really sure what the trace for gorillamux should look like; the same info exists for two spans, but one is named for net/http and is for gorillamux. This doesn't address that. The trace was already emitting two spans, with the gorillamux span being empty - and this PR populates the gorillamux span.

**Note**: This has been updated as several of the changes made were ported into #82 . Since then, the gorilla/mux spans were empty. This PR still has the effect of populating the gorilla/mux spans.

PR #92, opened 2023-04-24
[feat] Adding context propagation to net/http server instrumentation

oxeye-gal:
Hi,

This PR attempts to add context propagation to the net/http server instrumentation.

We do it by iterating over the buckets of the incoming request header map and trying to find the "traceparent" header.

This was tested with Go versions 1.12-1.19.

Our test environment consisted of instrumented Python application that sends a request to instrumented Go http server.

The produced trace looks like the following:

213373358-c184e90f-597c-4ca0-ba3f-52365fc89770

We would love to get your feedback on this.
Thanks!


damemi (dismissed):
This looks like you need to run `make generate` based on the [failure](https://github.com/open-telemetry/opentelemetry-go-instrumentation/actions/runs/4783444662/jobs/8520732823?pr=92#step:6:58)

Also the e2e failure is interesting, because it seems that this change removed the redundant empty span that @JamieDanielson was trying to fix in https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/86 (see [test output](https://github.com/open-telemetry/opentelemetry-go-instrumentation/actions/runs/4783444659/jobs/8520732629?pr=92#step:17:33))

damemi:
> also the redundant empty span removed is due to the removal of the net/http.(*ServeMux).ServeHTTP symbol from the functions to instrument as also mentioned by https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/86

Ah I see. Is that necessary for context propagation to work in this change?

JamieDanielson:
> > also the redundant empty span removed is due to the removal of the net/http.(*ServeMux).ServeHTTP symbol from the functions to instrument as also mentioned by #86
> 
> Ah I see. Is that necessary for context propagation to work in this change?

Is it cleaner to keep both function names in this PR so the output of the test remains the same, and I can update the other PR after this is merged in? That way we separate concerns, because I don't _think_ the extra empty span has an impact on this one (though I could be mistaken, I've been working on a few different branches of branches lately).

damemi:
> Is it cleaner to keep both function names in this PR so the output of the test remains the same, and I can update the other PR after this is merged in? That way we separate concerns

Yeah that's what I was wondering. if it's relevant to this PR, we can keep it, but otherwise @oxeye-gal can you remove that change so we don't duplicate the work?

MikeGoldsmith:
Hey @oxeye-gal - this PR needs rebasing after we merged ARM support on #82. I’m happy to help do that if you would like us to.

MikeGoldsmith (approved):
Looks great, thanks for the contribution @oxeye-gal. I'll look at updating #91 next 👍🏻

oxeye-gal:
@MikeGoldsmith seems like the build fails again due to the added Request.Header offset, what is the correct way to do it? thanks again for the help!

MikeGoldsmith:
After you've updated the offsets, you'll need to update the offsets.json. This can be done using `make docker-offsets`.

oxeye-gal:
I have added the new field to offsets-tracker, I hope this fixes it @MikeGoldsmith

MikeGoldsmith:
We're having hard time with this one 👎🏻  I'll give it a go too

include/go_types.h
```diff
@@ -44,6 +44,13 @@ struct go_iface
     void *data;
 };
 
+struct map_bucket {
+    char tophash[8];
+    struct go_string keys[8];
+    struct go_slice values[8];
+    void *overflow;
```
pdelewski, line 51: Is it used somewhere?

pdelewski, line 51: Is it used somewhere?

oxeye-gal: You mean the whole map_bucket struct or specifically the overflow field?

The map_bucket struct is used by the context propagation code, and even if the overflow field is not directly used by our code it is still important to keep it as part of the map_bucket struct as we are iterating through the buckets of the map, and if remove this field we will end up reading wrong values from memory.

pdelewski: I meant `overflow` field. Could you clarify why 

> we will end up reading wrong values from memory.

?

pkg/instrumentors/bpf/net/http/server/bpf/probe.bpf.c
```diff
@@ -50,6 +69,84 @@ volatile const u64 method_ptr_pos;
...
+
+    for (u64 j=0; j<MAX_BUCKETS; j++) {
+        if(j >= bucket_count) {
+            break;
+        }
+        res = bpf_probe_read(map_value, sizeof(struct map_bucket), header_buckets + (j * sizeof(struct map_bucket)));
+        if(res < 0) {
+            continue;
+        }
+        for (u64 i=0; i<8; i++) {
```
pdelewski, line 114: Would not make sense to extract `8` as constant? and use it also in struct

include/go_types.h
oxeye-gal: Sure, I used the following resource to understand the Go map implementation:

https://github.com/cch123/golang-notes/blob/master/map.md

As you can see the map bucket struct has an overflow field, if we omit it from the map_bucket struct line 110 in the probe.bpf.c file which attempts to read the next bucket will not read it correctly as the loop advances.

MikeGoldsmith:
@open-telemetry/go-instrumentation-approvers this is ready for another review 👍🏻

pkg/instrumentors/bpf/net/http/server/bpf/probe.bpf.c
MikeGoldsmith: @pdelewski I added a specific TODO when consolidating the HTTP probes to ensure const values are used for max loop iterations. Resolving this comment so we can get this merged then refactor.

edeNFed (approved):
lgtm

MrAlias (approved):

PR #94, opened 2023-04-24
custom user-agent (& expanding OTEL_EXPORTER_OTLP_* env var support)

robbkidd:
Instead of the Go autoinst agent appearing to telemetry data receivers as only the version of grpc-go against which it was built, this adds an identifier to the User-Agent header via the dial options parameter support within otlptracegrpc.

To help troubleshooting data transmission, runtime operating system and CPU architecture is included in the user-agent within a parenthetical comment in accordance with [the header's specification](https://www.rfc-editor.org/rfc/rfc7231#section-5.5.3).†

This replaces the low-level build up of a gRPC client connection with a reuse of the OTel Go SDK's grpc trace client constructor. A happy byproduct of using the SDK's constructor is that all of the OTEL_EXPORTER_OTLP_* environment variables related to exporting traces ought to be supported now, obviating the need to maintain code here to retrieve, interpret, and handle them.

TODO:
- [ ] confirm and link to a few existing issues these changes may solve
- [ ] tests. tests would be good.

† [RFC 7231](https://www.rfc-editor.org/rfc/rfc7231#section-5.5.3) though more readable documentation is available in [the MDN Web Docs about the User-Agent header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent).


pkg/opentelemetry/controller.go
```diff
@@ -38,6 +40,16 @@ const (
...
 )
 
+var (
+	// releaseVersion = ??? // TODO: reference version of the build to include in outward comms
+	// start of this autoinstrumentation's exporter User-Agent header, e.g. ""OTel-Go-Auto-Instrumentation/1.2.3"
+	baseUserAgent = fmt.Sprintf("OTel-Go-Auto-Instrumentation") // TODO: include the releaseVersion
+	// Information about the runtime environment for inclusion in User-Agent
+	runtimeInfo = fmt.Sprintf("%s (%s/%s)", strings.Replace(runtime.Version(), "go", "go/", 1), runtime.GOOS, runtime.GOARCH)
+	// The default User-Agent when no additions have been given
+	autoinstUserAgent = fmt.Sprintf("%s %s", baseUserAgent, runtimeInfo)
```
robbkidd, line 50: An example of what the user-agent output looks like from an Ubuntu AMD64 host running this branch at this commit:

```
OTel-Go-Auto-Instrumentation go/1.18.1 (linux/amd64) grpc-go/1.54.0
```

The `baseUserAgent` string was an initial stab at an ID string. Totally happy with a better identifier if someone's got one.

MikeGoldsmith (approved):

MikeGoldsmith:
This looks great, thanks @robbkidd 👍🏻

pkg/opentelemetry/controller.go
robbkidd: With the addition of `releaseVersion`, the user agent now looks like:

```
OTel-Go-Auto-Instrumentation/v0.1.0-alpha go/1.18.1 (linux/amd64) grpc-go/1.54.0
```

pkg/opentelemetry/controller.go
```diff
@@ -98,26 +104,20 @@ func NewController() (*Controller, error) {
...
-	conn, err := grpc.DialContext(timeoutContext, endpoint, grpc.WithInsecure(), grpc.WithBlock())
-	if err != nil {
-		log.Logger.Error(err, "unable to connect to OpenTelemetry collector", "addr", endpoint)
-		return nil, err
-	}
-
-	traceExporter, err := otlptracegrpc.New(ctx,
-		otlptracegrpc.WithGRPCConn(conn),
+	log.Logger.V(0).Info("Establishing connection to OTLP receiver ...")
+	otlpTraceClient := otlptracegrpc.NewClient(
```
robbkidd, line 115: :tada: This constructor does all the heavy lifting of looking up and parsing environment variables for the exporter config. 🎉

edeNFed (approved):

edeNFed:
Looks great!

I think it is worth adding to this PR the modification of `releaseVersion` variable as part of the release flow.
This should be fairly simple to do using `ldflags`

damemi:
I checked out your branch and ran `make fixture-nethttp` locally, the auto-instrumentation container appears to be failing to connect to the collector:

```
$ kubectl logs pod/sample-job-6xwd5 -c auto-instrumentation
{"level":"info","ts":1682422873.1658528,"caller":"cli/main.go:37","msg":"starting Go OpenTelemetry Agent ..."}
{"level":"info","ts":1682422873.1659956,"caller":"opentelemetry/controller.go:114","msg":"Establishing connection to OTLP receiver ..."}
{"level":"error","ts":1682422873.1662185,"caller":"opentelemetry/controller.go:120","msg":"unable to connect to OTLP endpoint","error":"failed to build resolver: passthrough: received empty target in Build()","stacktrace":"go.opentelemetry.io/auto/pkg/opentelemetry.NewController\n\t/app/pkg/opentelemetry/controller.go:120\nmain.main\n\t/app/cli/main.go:45\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:250"}
{"level":"error","ts":1682422873.1662562,"caller":"cli/main.go:47","msg":"unable to create OpenTelemetry controller","error":"failed to build resolver: passthrough: received empty target in Build()","stacktrace":"main.main\n\t/app/cli/main.go:47\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:250"}
```

robbkidd:
> I think it is worth adding to this PR the modification of releaseVersion variable as part of the release flow. This should be fairly simple to do using `ldflags`

I was only starting to venture into making changes to the build or release workflows. I'm game to add that, but I'm not quite clear on the source of truth for a particular build's version.

Is it the `versions.yaml`? I'm not familiar with what tooling uses that as a release config file.

In some of the pipelines I've worked with, we'll determine a build version based on an identifiably-a-release tag or the distance from one (somewhat like [the Collector does in its Makefile](https://github.com/open-telemetry/opentelemetry-collector/blob/076960f38374de3d176237ae269d029c8ef2f258/Makefile#L21-L22)). Is that a pattern this project would want to adopt?

For example, `git describe --always --match "v[0-9]*" HEAD` would produce a releaseVersion for this branch as `v0.1.0-alpha-8-g3e442b7`: currently sitting on commit 3e442b7 which is `8` commits ahead of the closest tag that matches what a version tag starts with `v0.1.0-alpha`.

robbkidd:
> Looks great!

😅 Y'all are OK with the creation of the unit test suite†? And that's the unit tests were added to the existing generate workflow?


† … that has only one test that probably needs changing if the source of truth for a version is not the `versions.yaml` file

robbkidd:
> appears to be failing to connect to the collector

Delegating the OTLP env var interpretation to the SDK means the endpoints need to have the protocol scheme included in the URL. (I believe.) Trying that in [ab37628](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/94/commits/ab37628995e46c8749cb70d531f5534341060891).

Another option (which is Working On My Machine™) would be to set `OTEL_EXPORTER_OTLP_INSECURE` to `true` for any instrument container, but I'd expect that to be determined when parsing an `http://` scheme from an endpoint config value.

robbkidd:
Rebased to latest main. If all passes, I believe this is ready for merge with the work to automate the release version lookup during build tracked in another issue.

MrAlias (commented):
Can you add a `version.go` similar to this: https://github.com/open-telemetry/opentelemetry-go/blob/86f325839ca329997672fbae8b054e7b6181b4d2/version.go  (with a [`version_test.go`](https://github.com/open-telemetry/opentelemetry-go/blob/86f325839ca329997672fbae8b054e7b6181b4d2/version_test.go)). Our release tooling will update versions found there.

Makefile
```diff
@@ -22,6 +22,18 @@ $(TOOLS)/go-licenses: PACKAGE=github.com/google/go-licenses
...
 tools: $(GOLICENSES)
 
+.PHONY: test
+#: run the tests!
+test:
+ifeq (, $(shell which gotestsum))
+	@echo " ***"
+	@echo "Running with standard go test because gotestsum was not found on PATH. Consider installing gotestsum for friendlier test output!"
+	@echo " ***"
+	go test -race -v ./...
```
MrAlias, line 32: This will not work across module boundaries. Please take a look at this: https://github.com/open-telemetry/opentelemetry-go/blob/86f325839ca329997672fbae8b054e7b6181b4d2/Makefile#L105-L119

Makefile
```diff
@@ -22,6 +22,18 @@ $(TOOLS)/go-licenses: PACKAGE=github.com/google/go-licenses
...
+else
+	gotestsum --junitfile unit-tests.xml --format testname -- -race ./...
```
MrAlias, line 34: Why is this being added?

test/e2e/gorillamux/traces.json
```diff
@@ -9,6 +9,12 @@
               "stringValue": "sample-app"
             }
           },
+          {
+            "key": "telemetry.auto.version",
+            "value": {
+              "stringValue": "v0.1.0-alpha"
```
MrAlias, line 15: This is going to be brittle. Is there anyway we can set this up so we don't need to change this every release?

Makefile
robbkidd: Convenience and possible future for outputting test results somewhere other than the GitHub Action job console. It's a lot and beyond the scope of getting a user-agent implemented, so I'm happy to remove this.

Makefile
robbkidd: Well, shoot. OK. Will look at that and adjust. (And I'll remove the gotestsum addendum below.)

robbkidd: I see the beginnings of the "[things with go.mod in them](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/105/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R5)" being added over in #105 which would make the basis for the multiple directories in which to run `go test`. I am ready to coordinate changes however y'all like.

robbkidd: [The Makefile trick](https://github.com/open-telemetry/opentelemetry-go/blob/86f325839ca329997672fbae8b054e7b6181b4d2/Makefile#L105-L119)† for factorying up test/ make targets for each module seems to not do anything for the module appearing in the current (repo root) directory.

```make
ALL_GO_MOD_DIRS := $(shell find . -type f -name 'go.mod' ! -path '$(TOOLS_MOD_DIR)/*' ! -path './LICENSES/*' -exec dirname {} \; | sort)
GO_MOD_TEST_TARGETS := $(ALL_GO_MOD_DIRS:%=test/%)

test: $(GO_MOD_TEST_TARGETS)
test/%: DIR=$*
test/%:
	cd $(DIR) && go test
```

[remake](https://remake.readthedocs.io/en/latest/) shows that the target for the root `go.mod` evaluates to `test/.` (which makes sense):

```
$ remake --targets | grep test
        test
        test/.
        test/./offsets-tracker
        test/./test/e2e/gorillamux
        test/./test/e2e/nethttp
```

But make ignores that target when actually run.

```
$ make test
cd ./offsets-tracker && go test
?       go.opentelemetry.io/auto/offsets-tracker        [no test files]
cd ./test/e2e/gorillamux && go test
?       go.opentelemetry.io/auto/test/e2e/gorillamux    [no test files]
cd ./test/e2e/nethttp && go test
?       go.opentelemetry.io/auto/test/e2e/nethttp       [no test files]
```

† Which I have mad respect for personally, having been accused of Too Much Make in other projects.

Makefile
robbkidd: Removed the check for gotestsum and run-if-present in 335f20d.

Makefile
robbkidd: Added that ☝🏻 variation of the test targets to the Makefile in 335f20d to give us more to look at and review.

version.go
```diff
@@ -0,0 +1,20 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package auto
+
+// Go autoinstrumentation release version
```
MrAlias, line 17:
```suggestion
// Version is the current release version of OpenTelemetry Go auto-instrumentation in use.
```

Makefile
```diff
@@ -27,6 +27,14 @@ $(TOOLS)/go-licenses: PACKAGE=github.com/google/go-licenses
 .PHONY: tools
 tools: $(GOLICENSES)
 
+ALL_GO_MODS := $(shell find . -type f -name 'go.mod' ! -path '$(TOOLS_MOD_DIR)/*' ! -path './LICENSES/*' | sort)
+GO_MODS_TO_TEST := $(ALL_GO_MODS:%=test/%)
+
+test: $(GO_MODS_TO_TEST)
```
MrAlias, line 33:
```suggestion
.PHONY: test
test: $(GO_MODS_TO_TEST)
```

MrAlias (approved):
Overall looks good, just minor issues. Thanks for the contribution!

robbkidd:
In thinking through this morning about the version number reflected in code, one disadvantage of the multimod-managed hard-coded value is that receivers of telemetry cannot distinguish uses of release builds versus dev builds. That's one benefit of the version being determined at build time, usually with ldflags set to the output of git describe.

As a vendor receiving telemetry from thousands of independent customer sources, Honeycomb has found it incredibly useful to be able to see more nuanced version information in user agents.

MrAlias:
@robbkidd can you sync with `main` so we can merge this?

PR #102, opened 2023-04-26
Fix source code license headers

MrAlias:
- Add a make target to check this
- Add a GitHub action to check this

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -6,6 +6,9 @@ on:
   pull_request:
   workflow_dispatch:
 
+env:
+  GO_VERSION: "1.20"
+
 jobs:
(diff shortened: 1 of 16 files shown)


damemi (approved):

.github/workflows/checks.yml
```diff
@@ -19,6 +22,31 @@ jobs:
           -v -n "*.md" "**/*.md"
           --exclude "http://localhost*"
         fail: true
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install Go
+        uses: actions/setup-go@v4
```
pellared, line 29: Can we pin the GitHub Action versions?

pellared (approved):

PR #105, opened 2023-04-26
Add go-mod-tidy target and action

MrAlias:


MikeGoldsmith (approved):

.github/workflows/checks.yml
```diff
@@ -46,7 +46,9 @@ jobs:
           path: ~/.tools
           key: ${{ runner.os }}-${{ env.cache-name }}-${{ hashFiles('./internal/tools/**') }}
       - name: Run linters
-        run: make license-header-check
+        run: make license-header-check go-mod-tidy
```
pellared, line 49: How about simplifying to the following?

```suggestion
        run: make precommit
```

pellared (approved):

.github/workflows/checks.yml
MrAlias: This is going to be different than `precommit`, see https://github.com/open-telemetry/opentelemetry-go/blob/86f325839ca329997672fbae8b054e7b6181b4d2/.github/workflows/ci.yml#L47

pellared: I see. How about encapsulating it to a `ci` Make target?

pellared: Make sense. Resolving.

PR #108, opened 2023-04-26
Add golangci-lint make target

MrAlias:
This does not address the large number of lint errors in the project. That is left for subsequent PRs.


.golangci.yml
```diff
@@ -0,0 +1,246 @@
+# See https://github.com/golangci/golangci-lint#config-file
+run:
+  issues-exit-code: 1 #Default
```
pellared, line 3:
```suggestion
  issues-exit-code: 0 # TODO: Change to 1 as part of: [Github issue that fixes the issues]
```

pellared: I created https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/109

MrAlias: Why?

pellared: So that:
- we could simply run `precommit` in GH Workflow
- `precommit` does not fail until we fix the issues

pellared: It may be not needed if https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/110 is done first and there are not more issues.

MrAlias: I would rather add the GitHub action once we fix all the lint instead of having the CLI return with `0`.

pellared: OK. I changed the description of https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/109

pellared (approved):

MikeGoldsmith (approved):

PR #110, opened 2023-04-26
[chore] Lint fixes for pkg/opentelemetry

MrAlias:
Related to https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/108


pkg/opentelemetry/id_generator.go
```diff
@@ -25,14 +25,16 @@ type eBPFSourceIDGenerator struct{}
...
 
-func NewEbpfSourceIDGenerator() *eBPFSourceIDGenerator {
+func newEbpfSourceIDGenerator() *eBPFSourceIDGenerator {
 	return &eBPFSourceIDGenerator{}
 }
 
-func ContextWithEbpfEvent(ctx context.Context, event events.Event) context.Context {
-	return context.WithValue(ctx, ebpfEventKey{}, event)
+// ContextWithEbpfEvent returns a copy of parent in which event is stored.
+func ContextWithEbpfEvent(parent context.Context, event events.Event) context.Context {
```
pellared, line 33: _Can be addressed as a separate issue if there are more places with the same problem_

```golang
// ContextWithEBPFEventreturns a copy of parent in which event is stored.
func ContextWithEBPFEvent(parent context.Context, event events.Event) context.Context {
```

Reason: https://github.com/golang/go/wiki/CodeReviewComments#initialisms

pellared (approved):

pkg/opentelemetry/id_generator.go
MrAlias: I'm all for the rename, but would want that to be its own PR. This is targeted at fixing the existing lint issues.

MikeGoldsmith (approved):

PR #113, opened 2023-04-26
[chore] Lint fixes for pkg/log

MrAlias:
Related to #108


pdelewski (approved):

pkg/log/logger.go
```diff
@@ -20,8 +20,10 @@ import (
 	"go.uber.org/zap"
 )
 
+// Logger is an interface to an abstract logging implementation.
```
pellared, line 23:
```suggestion
// Logger is a global logger used for internal logging.
```

MikeGoldsmith (approved):

PR #118, opened 2023-04-27
Fix context propagation across goroutines

edeNFed:
## What's changed
1. Added 4 utility functions that should be used by instrumentors:
- `get_consistent_key`: This function provides a key that works across any version of Go. This key should be used in the eBPF map created by the instrumentor to record the end timestamp at the end of the probed function.
- `start_tracking_span(void *contextContext, struct span_context *sc)` - This function should be used in order to perform in-process context propagation, meaning passing context between different instrumentations in the same applications (see below test case for example).
- `stop_tracking_span(struct span_context *sc)` - This function marks the passed span context as no more relevant for context propagation. This function should usually be called when recording the end timestamp of a span.
- `get_parent_span_context(void *ctx)` - Use this function in order to get the parent span context if available. This function receives the address of `context.Context` object.

2. Modified existing instrumentors to use those functions when reading/writing to eBPF maps and to performing context propagation between instrumentors.

3. Modified gRPC (client) instrumentation to inject headers by probing `loopyWriter.headerHandler` instead of `http2Client createHeaderFields`: In Go >= 1.17, the return value is passed by registers and not by the stack. Currently, eBPF is not able to modify registers' values, only memory. Changing the uprobe to attach to `loopyWriter.headerHandler` allow us to add context-propagation headers even on Go versions above 1.17.

## Validation
I used the following code as a test case:
```go
func main() {
	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
		con, err := grpc.Dial("grpc-server.default:50051", grpc.WithTransportCredentials(insecure.NewCredentials()))
		if err != nil {
			panic(err)
		}

		stopCh := make(chan struct{})
		go func() {
			healthClient := healthpb.NewHealthClient(con)
			resp, err := healthClient.Check(r.Context(), &healthpb.HealthCheckRequest{})
			if err != nil {
				panic(err)
			}

			fmt.Fprintf(w, "gRPC server status: %s", resp.Status)
			close(stopCh)
		}()

		<-stopCh
	})

	fmt.Println("HTTP server started on port 8080")
	err := http.ListenAndServe(":8080", nil)
	if err != nil {
		panic(err)
	}
}
```

Which produces the following trace (after the applied fix)
Screenshot 2023-05-22 at 1 11 33 PM

**Tested on Go 1.15 and 1.20, both on arm64 and amd64**


edeNFed:
Blocked by: #82

MrAlias:
Related to https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/86

edeNFed:
@open-telemetry/go-instrumentation-approvers this is ready for review, thanks! 🙏

include/go_context.h
```diff
@@ -16,22 +16,71 @@
...
+    struct span_context *parent_sc = bpf_map_lookup_elem(&tracked_spans, &parent_ctx);
+    if (parent_sc == NULL)
+    {
+        return NULL;
+    }
+
+    return parent_sc;
+}
+
+static __always_inline void track_running_span(void *contextContext, struct span_context *sc) {
```
MikeGoldsmith, line 72: For consistency with `stop_tracking_span`, could we rename to `start_tracking_span`?

include/go_context.h
```diff
@@ -16,22 +16,71 @@
...
+    bpf_map_update_elem(&tracked_spans, &contextContext, sc, BPF_ANY);
+    bpf_map_update_elem(&tracked_spans_by_sc, sc, &contextContext, BPF_ANY);
+}
```
MikeGoldsmith, line 75:
```suggestion
static __always_inline void track_running_span(void *ctx, struct span_context *sc) {
    bpf_map_update_elem(&tracked_spans, &ctx, sc, BPF_ANY);
    bpf_map_update_elem(&tracked_spans_by_sc, sc, &ctx, BPF_ANY);
}
```

pkg/instrumentors/bpf/google/golang/org/grpc/bpf/probe.bpf.c
```diff
@@ -111,84 +141,78 @@ int uprobe_ClientConn_Invoke_Returns(struct pt_regs *ctx)
...
-
+    bpf_map_delete_elem(&grpc_events, &key);
+    stop_tracking_span(&grpcReq.sc);
     return 0;
 }
 
-// func (t *http2Client) createHeaderFields(ctx context.Context, callHdr *CallHdr) ([]hpack.HeaderField, error)
-SEC("uprobe/Http2Client_createHeaderFields")
-int uprobe_Http2Client_CreateHeaderFields(struct pt_regs *ctx)
+// func (l *loopyWriter) headerHandler(h *headerFrame) error
```
MikeGoldsmith, line 156: Is it intentional to monitor `loopyWriter_headerHandler` instead of `Http2Client_createHeaderFields`? Do we need to still monitor `http2Client_NewStream` too?

Can you say anymore about why it's better to use this instead?

MikeGoldsmith (commented):
This looks really good, thanks @edeNFed.

I've left some comments and suggestions.

It's worth noting #91 conflicts with this. Whichever PR lands first, will require the other to be rebased.

pkg/instrumentors/bpf/google/golang/org/grpc/bpf/probe.bpf.c
edeNFed: Added description to PR

edeNFed:
@open-telemetry/go-instrumentation-approvers This is ready for another review
Thank you 🙏

MrAlias (approved):

pkg/instrumentors/bpf/net/http/client/bpf/probe.bpf.c
```diff
@@ -54,8 +53,10 @@ static __always_inline long inject_header(void* headers_ptr, struct span_context
 
     // Currently only maps with less than 8 keys are supported for injection
     if (map_keyvalue_count >= 8) {
+        bpf_printk("Map size is bigger than 8, skipping context propagation");
```
MikeGoldsmith, line 56: My understanding is `bpf_printk` has non-negligible performance impact. Maybe we should review where we use them, and have some sort of flag to enable / disable? There are a few 

I'll create a ticket to track from this comment.

MikeGoldsmith (approved):
Looks great, thanks @edeNFed 👍🏻

PR #121, opened 2023-04-28
[chore] update e2e tests to use BATS assertions

robbkidd:
## Short description of the changes

- Use [bats](https://bats-core.readthedocs.io/en/stable/) for e2e testing
- Modify step in workflow to use BATS to make more precise assertions about the unredacted trace data
- Moves the redaction logic for the traces.json file to the BATS test helpers and asserts no-git-difference in the test suites
- Add utilities test helper for more clarity (specifically includes named functions that are easier to reason about)

This can be run locally as well. You will get a `traces-orig.json` file for each test run, and from there you can further work with the tests by running `bats verify.bats` in the directory if you install bats locally.

Note I added a TODO for gorilla/mux spans coming out with a scope name of net/http, probably as a result of being produced by the net/http probe. This is a PR for updating tests, so I propose fixing the probe source is out of scope and  should be addressed in a separate PR.

## How to verify that this has the expected result

e2e tests pass with expected results. Running locally with `make fixture-nethttp` and the others has expected results, and also creates a clean `traces.json` with expected output.

To see an example of a fail locally, run one of the e2e tests (e.g. `make fixture-gin`), change directory to `test/e2e/gin`, change the local `traces-org.json` (maybe change the real span ID to `"nope"`), and run `bats verify.bats`.

e2e-fail-gin

### DIFF
--- a/.github/workflows/kind.yml
+++ b/.github/workflows/kind.yml
@@ -22,6 +22,8 @@ jobs:
         uses: actions/setup-go@v4
         with:
           go-version: "1.20"
+      - name: Setup BATS
+        uses: mig4/setup-bats@v1
       - name: Build auto-instrumentation
         run: |
           IMG=otel-go-instrumentation:latest make docker-build
--- a/.github/workflows/kind.yml
+++ b/.github/workflows/kind.yml
@@ -73,4 +73,4 @@ jobs:
           kubectl cp -c filecp default/test-opentelemetry-collector-0:tmp/trace.json ./test/e2e/${{ matrix.library }}/traces.json
       - name: verify output
         run: |
-          bats ./test/e2e/$(LIBRARY)/verify.bats
+          bats ./test/e2e/${{ matrix.library }}/verify.bats
--- a/.github/workflows/kind.yml
+++ b/.github/workflows/kind.yml
@@ -70,7 +70,8 @@ jobs:
           kubectl wait --for=condition=Complete --timeout=60s job/sample-job
       - name: copy telemetry trace output
         run: |
-          kubectl cp -c filecp default/test-opentelemetry-collector-0:tmp/trace.json ./test/e2e/${{ matrix.library }}/traces.json
-      - name: verify output
+          kubectl cp -c filecp default/test-opentelemetry-collector-0:tmp/trace.json ./test/e2e/${{ matrix.library }}/traces-orig.json
+          rm -f ./test/e2e/${{ matrix.library }}/traces.json
+      - name: verify output and redact to traces.json
         run: |
           bats ./test/e2e/${{ matrix.library }}/verify.bats
--- a/.github/workflows/kind.yml
+++ b/.github/workflows/kind.yml
@@ -68,9 +68,7 @@ jobs:
           kubectl wait --for=condition=Complete --timeout=60s job/sample-job
       - name: copy telemetry trace output
         run: |
-          kubectl cp -c filecp default/test-opentelemetry-collector-0:tmp/trace.json ./test/e2e/${{ matrix.library }}/traces.json.tmp
-          jq 'del(.resourceSpans[].scopeSpans[].spans[].endTimeUnixNano, .resourceSpans[].scopeSpans[].spans[].startTimeUnixNano) | .resourceSpans[].scopeSpans[].spans[].spanId|= (if . != "" then "xxxxx" else . end) | .resourceSpans[].scopeSpans[].spans[].traceId|= (if . != "" then "xxxxx" else . end) | .resourceSpans[].scopeSpans|=sort_by(.scope.name)' ./test/e2e/${{ matrix.library }}/traces.json.tmp | jq --sort-keys . > ./test/e2e/${{ matrix.library }}/traces.json
-          rm ./test/e2e/${{ matrix.library }}/traces.json.tmp
+          kubectl cp -c filecp default/test-opentelemetry-collector-0:tmp/trace.json ./test/e2e/${{ matrix.library }}/traces.json
       - name: verify output
         run: |
-          make check-clean-work-tree
+          bats ./test/e2e/$(LIBRARY)/verify.bats
--- a/Makefile
+++ b/Makefile
@@ -136,9 +136,8 @@ fixtures/%:
 	kubectl wait --for=condition=Ready --timeout=60s pod/test-opentelemetry-collector-0
 	kubectl -n default create -f .github/workflows/e2e/k8s/sample-job.yml
 	kubectl wait --for=condition=Complete --timeout=60s job/sample-job
-	kubectl cp -c filecp default/test-opentelemetry-collector-0:tmp/trace.json ./test/e2e/$(LIBRARY)/traces.json.tmp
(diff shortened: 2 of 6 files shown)


.github/workflows/kind.yml
```diff
@@ -68,9 +68,7 @@ jobs:
           kubectl wait --for=condition=Complete --timeout=60s job/sample-job
       - name: copy telemetry trace output
         run: |
-          kubectl cp -c filecp default/test-opentelemetry-collector-0:tmp/trace.json ./test/e2e/${{ matrix.library }}/traces.json.tmp
-          jq 'del(.resourceSpans[].scopeSpans[].spans[].endTimeUnixNano, .resourceSpans[].scopeSpans[].spans[].startTimeUnixNano) | .resourceSpans[].scopeSpans[].spans[].spanId|= (if . != "" then "xxxxx" else . end) | .resourceSpans[].scopeSpans[].spans[].traceId|= (if . != "" then "xxxxx" else . end) | .resourceSpans[].scopeSpans|=sort_by(.scope.name)' ./test/e2e/${{ matrix.library }}/traces.json.tmp | jq --sort-keys . > ./test/e2e/${{ matrix.library }}/traces.json
```
damemi, line 72: I am +1 any solution that looks better than my jq monstrosity here hah

robbkidd: In fairness to your jq monstrosity, @damemi, this experiment breaks things up into a bunch of small jq monsters that form up into bigger jq monsters with bash and bats glue. It's like a different flavor of poison: we get to choose the side-effects.

robbkidd:
First learning from the experiment: `bats` is not already on the runner.

JamieDanielson:
The second note I added above "2) http span names do not match latest conventions" is fixed with #143 . I can update that after this gets merged or vice versa.

JamieDanielson:
The first note I added above "1) gorilla/mux spans are actually using the net/http probe" is partially fixed with #86 . There is still an extra net/http span there, but at least gorilla/mux scope will show up as expected.

MikeGoldsmith (approved):
I think this is great and we should take this PR as is, then update the tests in the other two PRs to use BATS.

robbkidd:
What do y'all think about removing the `tests/e2e/*/traces.json` files and adding them to .gitignore with this PR? With specific test assertions about expected state of distinct portions of the JSON, the files don't need to be tracked by version control to support e2e test runs.

MikeGoldsmith:
Does removing the traces.json still allow a jq comparision check? The BATS assertions are good for determining whether something expect is present but we can't check for unexpected data, eg additional span attributes. The jq comparison allows us to check the complete shape of the generated telemetry, with a few redacted values.

Should we care about extra attributes if the ones we expect to be present are?

robbkidd:
Alrightie!

Latest change:

+ brings back the redaction jq command as a function within the BATS utilities
+ one of the tests in each library's e2e suite is to perform the JSON redaction and assert no `git diff`
+ removes the `traces.json` file before running the tests (and before the redaction-jq generates it again) to protect the e2e checks from verifying the already-known-good state that is checked in to version control

I've held off adding an e2e suite for grpc because that instrumentation doesn't pass the assertions that the others do, yet. I reckon that work can be in a follow-up PR.

.github/workflows/kind.yml
robbkidd: Today I Learned that `jq` has no cares about arbitrary whitespace, so [the monster is BACK](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/121/files#diff-93a3e09daea561b2dfc5c4c85465654bbf4470a640c139a97ae0bc638586f625R40-R55) but after a bath, some hair combing, and some ribbons put on!

robbkidd:
Would that addition be to `.github/workflows/kind.yaml` like so?

```diff
# …
  jobs:
    kubernetes-test:
+     concurrency:
+       group: ${{ github.ref }}
+       cancel-in-progress: true
      strategy:
        matrix:
          k8s-version: ["v1.26.0"]
          library: ["gorillamux", "nethttp", "gin"]
# … 
```

And if I understand that correctly, it would cancel all of these e2e tests if any faster check against the same commit fails, right?

Can the CI config change happen on a separate PR? Or does it block acceptance of this one?

MasonLegere:
The [addition would be at the workflow level](https://docs.github.com/en/actions/using-jobs/using-concurrency) as opposed to the job and added on [line 10](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/121/files#diff-0083d2437ae7b209c8d66a16f1d8bbebf33c3ce0885bd0ae9a223b39c5dc343cR10). The impact is pretty small and only applies in the case where a commit is pushed quickly after another. The new commit would trigger all the new e2e tests to re-run where most likely the runners from the previous commit should be cancelled to save runner minutes if they're still active. 

It's pretty small and might not come up frequently but I can add it as a separate PR if there is some interest

robbkidd:
It sounds like a good idea. I would support it† in a separate PR so that the conversation around adding short-circuits to CI can be distinct from the conversation here about an overhaul of end-to-end test assertions.

† Disclosure: I am a new contributor to this project, so my support is of questionable value.

robbkidd:
@open-telemetry/go-instrumentation-approvers I believe this is ready and worthy of review. 🙇🏻 💌 💞

JamieDanielson (approved):

edeNFed (approved):
Looks great! 
I think we have enough approvers, can we proceed to merge this?

MikeGoldsmith:
@open-telemetry/go-instrumentation-approvers This is ready for more eyes before we can merge.

Makefile
```diff
@@ -136,9 +136,9 @@ fixtures/%:
 	kubectl wait --for=condition=Ready --timeout=60s pod/test-opentelemetry-collector-0
 	kubectl -n default create -f .github/workflows/e2e/k8s/sample-job.yml
 	kubectl wait --for=condition=Complete --timeout=60s job/sample-job
-	kubectl cp -c filecp default/test-opentelemetry-collector-0:tmp/trace.json ./test/e2e/$(LIBRARY)/traces.json.tmp
-	jq 'del(.resourceSpans[].scopeSpans[].spans[].endTimeUnixNano, .resourceSpans[].scopeSpans[].spans[].startTimeUnixNano) | .resourceSpans[].scopeSpans[].spans[].spanId|= (if . != "" then "xxxxx" else . end) | .resourceSpans[].scopeSpans[].spans[].traceId|= (if . != "" then "xxxxx" else . end) | .resourceSpans[].scopeSpans|=sort_by(.scope.name)' ./test/e2e/$(LIBRARY)/traces.json.tmp | jq --sort-keys . > ./test/e2e/$(LIBRARY)/traces.json
-	rm ./test/e2e/$(LIBRARY)/traces.json.tmp
+	kubectl cp -c filecp default/test-opentelemetry-collector-0:tmp/trace.json ./test/e2e/$(LIBRARY)/traces-orig.json
+	rm -f ./test/e2e/$(LIBRARY)/traces.json
+	bats ./test/e2e/$(LIBRARY)/verify.bats
```
MrAlias, line 141: Can we use a docker image to run this so we don't require a dependency on bats for local development?

This can be a follow up task.

MrAlias: https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/166

Issue #78, opened 2023-04-14
gRPC CreateHeaderFields probe fails with permission denied


vreynolds:
### Describe the bug

Running a sample gRPC app with go auto-instrumentation causes the app to crashloop, instrumentation fails with a permission error.

### Environment

 - OS: osx ventura 13.3.1, k8s 1.25.4
 - Go Version: 1.20
 - Version: [e.g. 1295520, v0.1.0]

### To Reproduce

1. Using the [example](https://github.com/JamieDanielson/grpc-go/tree/jamie.grpc-example/examples/helloworld)
2. Create the deployments in k8s cluster following readme
3. Notice the server and client pods are crashlooping
4. Check the logs for an instrumentation pod and observe the logs:

```
2023-04-14 19:40:06 {"level":"error","ts":1681515606.0069542,"caller":"cli/main.go:88","msg":"error while running instrumentors","error":"field UprobeHttp2ClientCreateHeaderFields: program uprobe_Http2Client_CreateHeaderFields: load program: permission denied: 979: (71) r7 = *(u8 *)(r7 +0): R0=inv(id=51) R1_w=inv(id=0,umax_value=65535,var_off=(0x0; 0xffff)) R2_w=inv(id=0,umax_value=255,var_off=(0x0; 0xff)) R3_w=inv(id=0,umax_value=16777215,var_off=(0x0; 0xffffff)) R4_w=invP3 R5_w=fp-204 R6_w=fp-381 R7_w=map_value(id=0,off=0,ks=4,vs=16,umax_value=15,var_off=(0x0; 0xf)) R8_w=map_value(id=0,off=0,ks=4,vs=16,umax_value=15,var_off=(0x0; 0xf)) R9=inv(id=0,umax_value=1099511627775,var_off=(0x0; 0xffffffffff)) R10=fp0 fp-48=???mmmmm fp-56=mmmmmmmm fp-64=mmmmmmmm fp-72=mmmmmmmm fp-80=mmmmmmmm fp-96=mmmmmmmm fp-104=mmmmmmmm fp- (truncated, 1249 line(s) omitted)","stacktrace":"main.main\n\t/app/cli/main.go:88\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:250"}
```

### Expected behavior

The app runs and is instrumented.

MrAlias:
This was not reproducible with the keyval version of this code, so the cause is likely in the diff between this and that project.

[...]

PR #128, opened 2023-05-01
Add OTEL_GO_AUTO_SHOW_VERIFIER_LOG env var

edeNFed:
Recent versions of cilium/ebpf library stopped printing detailed verifier logs by default.
This makes it hard to debug verifier errors (like #78).
This PR adds an additional env var called `OTEL_GO_AUTO_SHOW_VERIFIER_LOG` which will print the full verifier log if set to true.
Collecting verifier logs is an expensive operation, and therefore is turned off by default.


pkg/instrumentors/utils/ebpf.go
```diff
@@ -0,0 +1,62 @@
...
+	return err
+}
+
+// Getting full verifier log is expensive, so we only do it if the user explicitly asks for it.
+func shouldShowVerifierLogs() bool {
+	val, exists := os.LookupEnv(showVerifierLogEnvVar)
+	if exists {
+		boolVal, err := strconv.ParseBool(val)
+		if err != nil {
+			return true
```
MikeGoldsmith, line 56: Maybe we should default to `false` (disabled) if the value in the env var is invalid?

pkg/instrumentors/utils/ebpf.go
```diff
@@ -0,0 +1,62 @@
...
+const (
+	showVerifierLogEnvVar = "OTEL_GO_AUTO_SHOW_VERIFIER_LOG"
+)
+
+// LoadEbpfObjects loads eBPF objects from the given spec into the given interface.
+// If the environment variable OTEL_GO_AUTO_SHOW_VERIFIER_LOG is set to true, the verifier log will be printed.
+func LoadEbpfObjects(spec *ebpf.CollectionSpec, to interface{}, opts *ebpf.CollectionOptions) error {
+	showVerifierLogs := shouldShowVerifierLogs()
+	if showVerifierLogs {
+		opts.Programs.LogSize = ebpf.DefaultVerifierLogSize * 100
```
MikeGoldsmith, line 35: Why do we need to extend the default size? Should we consider allowing the log size to be adjusted too?

MikeGoldsmith (changes requested):
Look good to me. I've left a couple of queries regarding default values.

pkg/instrumentors/utils/ebpf.go
MikeGoldsmith: Could we add a comment to share what you've put above, maybe as a declare var?

MikeGoldsmith (approved):
Looks good - I think a comment / issue to review LogSize configuration would be good 👍🏻

pkg/instrumentors/utils/ebpf.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/cilium/ebpf"
+)
+
+const (
+	showVerifierLogEnvVar = "OTEL_GO_AUTO_SHOW_VERIFIER_LOG"
+)
+
+// LoadEbpfObjects loads eBPF objects from the given spec into the given interface.
+// If the environment variable OTEL_GO_AUTO_SHOW_VERIFIER_LOG is set to true, the verifier log will be printed.
+func LoadEbpfObjects(spec *ebpf.CollectionSpec, to interface{}, opts *ebpf.CollectionOptions) error {
```
MrAlias, line 32: Hmm, I thought we had a lint check to catch this initialism. I'll have to follow up on that.

```suggestion
// LoadEBPFObjects loads eBPF objects from the given spec into the given interface.
// If the environment variable OTEL_GO_AUTO_SHOW_VERIFIER_LOG is set to true, the verifier log will be printed.
func LoadEBPFObjects(spec *ebpf.CollectionSpec, to interface{}, opts *ebpf.CollectionOptions) error {
```

pkg/instrumentors/utils/ebpf.go
```diff
@@ -0,0 +1,61 @@
...
+	showVerifierLogs := shouldShowVerifierLogs()
```
MrAlias, line 33:
```suggestion
	// Getting full verifier log is expensive, so we only do it if the user explicitly asks for it.
	showVerifierLogs := shouldShowVerifierLogs()
```

pkg/instrumentors/utils/ebpf.go
```diff
@@ -0,0 +1,61 @@
...
+		var ve *ebpf.VerifierError
+		if errors.As(err, &ve) {
+			fmt.Printf("Verifier log: %+v\n", ve)
+		}
+	}
+
+	return err
+}
+
+// Getting full verifier log is expensive, so we only do it if the user explicitly asks for it.
```
MrAlias, line 50:
```suggestion
// shouldShowVerifierLogs returns if the user has configured verifier logs to be emitted.
```

Issue #66, opened 2023-04-04
Release v0.1.0-alpha


MrAlias:
Blocking

- [x] https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/40

damemi:
With #40 merged, are we set to tag the release?

MrAlias:
Do we want to setup a [vanity URL](https://github.com/open-telemetry/opentelemetry-go-vanityurls) for this package before we release it?

It currently has a package of `github.com/open-telemetry/opentelemetry-go-instrumentation`, but all other OTel Go projects use `go.opentelemetry.io/`.

MrAlias:
Maybe go.opentelemetry.io/auto?

damemi:
Vanity url looks really cool, I don't have a big preference on the name but definitely +1 to setting it up

TylerHelmuth:
@MrAlias need to open this back up as the [release workflow didn't pass](https://github.com/open-telemetry/opentelemetry-go-instrumentation/actions/runs/4724518902/jobs/8381827448).  Looks like our dockerhub token isn't properly setup

TylerHelmuth:
Followed up with the original ticket to create the secretes here: https://github.com/open-telemetry/community/issues/1377

MrAlias:
> If we don't want to wait I believe we could update the release workflow to only push to `ghcr.io`, but we'd have to do a new tag to rerun the workflow.

I'm okay waiting for the TC to try and create the repo.

TylerHelmuth:
Think I should open a second ticket?  Technically the closed ticket I linked didn't asked for the repository to be made, only for the secrets.

MrAlias:
> Think I should open a second ticket? Technically the closed ticket I linked didn't asked for the repository to be made, only for the secrets.

Yeah, that makes sense to me.

TylerHelmuth:
New issue created: https://github.com/open-telemetry/community/issues/1442

MrAlias:
@TylerHelmuth: https://github.com/open-telemetry/community/issues/1442 looks resolved, but the build is still failing:

https://github.com/open-telemetry/opentelemetry-go-instrumentation/actions/runs/4724518902/jobs/8406320441

TylerHelmuth:
Ok I will keep investigating

TylerHelmuth:
[...]

PR #129, opened 2023-05-01
docs: update emojivoto example

JamieDanielson:
This updates the emojivoto example in the getting-started docs:

- Remove launcher, which is no longer needed after #40 
- Update instrumentation to use `otel-go-instrumentation` image instead of `keyval/otel-go-agent` 
- Simplify paths to kubernetes manifests (use relative instead of absolute)

Note: This instruments gRPC but has the issue referenced in #78 . Also, the docker image has not yet been published so this currently requires a local build. Since those are already known issues, I think this is worth merging in anyway - I'm not sure how helpful it is to show an example of the instrumentation if it's not actually instrumentation from this library.

{"level":"error","ts":1682967887.022351,"caller":"instrumentors/runner.go:88","msg":"error while loading instrumentors, cleaning up","name":"google.golang.org/grpc","error":"field UprobeHttp2ClientCreateHeaderFields: program uprobe_Http2Client_CreateHeaderFields: load program: permission denied: 912: (73) *(u8 *)(r10 -196) = r7: 913: (79) r2 = *(u64 *)( (truncated, 1185 line(s) omitted)","stacktrace":"go.opentelemetry.io/auto/pkg/instrumentors.(*Manager).load\n\t/app/pkg/instrumentors/runner.go:88\ngo.opentelemetry.io/auto/pkg/instrumentors.(*Manager).Run\n\t/app/pkg/instrumentors/runner.go:36\nmain.main\n\t/app/cli/main.go:86\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:250"}

### DIFF
--- a/docs/getting-started/emojivoto-instrumented.yaml
+++ b/docs/getting-started/emojivoto-instrumented.yaml
@@ -22,17 +22,6 @@ spec:
       serviceAccountName: emoji
       shareProcessNamespace: true
       terminationGracePeriodSeconds: 0
-      initContainers:
-        - name: copy-launcher
-          image: keyval/launcher:v0.1
-          command:
-            - cp
-            - -a
-            - /kv-launcher/.
-            - /odigos-launcher/
-          volumeMounts:
-            - name: launcherdir
-              mountPath: /odigos-launcher
       containers:
         - env:
             - name: GRPC_PORT
@@ -41,12 +30,6 @@ spec:
               value: "8801"
           image: docker.l5d.io/buoyantio/emojivoto-emoji-svc:v11
           name: emoji-svc
-          command:
-            - /odigos-launcher/launch
-            - /usr/local/bin/emojivoto-emoji-svc
-          volumeMounts:
-            - mountPath: /odigos-launcher
-              name: launcherdir
           ports:
             - containerPort: 8080
               name: grpc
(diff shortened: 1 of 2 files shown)


docs/getting-started/emojivoto-instrumented.yaml
```diff
@@ -56,7 +39,7 @@ spec:
             requests:
               cpu: 100m
         - name: emojivoto-emoji-instrumentation
-          image: keyval/otel-go-agent:v0.6.0
+          image: otel-go-instrumentation:v0.1
```
MikeGoldsmith, line 42: How do you feel about setting the image as `otel-go-instrumentation` (without version suffix) and updating the demo prerequisites to require the docker image be created first with`make docker-build`?

Then we can update the image to an published one once #66 is done.

MikeGoldsmith (changes requested):
Looks good for me. Made a suggestion to require the docker image be created locally as part of the example until an official image is available.

docs/getting-started/emojivoto-instrumented.yaml
JamieDanielson: Updated based on your feedback 👍

edeNFed (approved):

MikeGoldsmith (approved):
Thanks @JamieDanielson  👍🏻

JamieDanielson:
> Thanks @JamieDanielson - just needs a changelog entry 👍🏻

Happy to add, but wasn't sure it needed one since it's just an example update.

PR #155, opened 2023-05-12
[chore] Consolidate Docker image names

### LINKED ISSUE #153: Standardize docker image names
We have two formats for our docker images;
- built locally using Makefile: `otel-instrumentation-go`
- built in CI workflow and published to github & docker hub: `autoinstrumentation-go`

We should have one format and consolidate them to be consistent. It's useful to have the `otel` prefix for locally built images but not required for published images because they have the repository prefix.

              I didn’t see this before, but we use a different default name in the makefile than here. 
https://github.com/open-up/opentelemetry-go-instrumentation/blob/main/Makefile#L33

It would be nice to use the same name. Locally it’s good to have the otel prefix so it’s easy to find but not as important for published images as it gets prefix from the image store. 

I don’t want to block this PR to bike shed on names so will create a follow-up issue to track our image naming.

_Originally posted by @MikeGoldsmith in https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/152#discussion_r1189469925_

MasonLegere:
Consolidates the image name used locally and for testing with the same name published during releases resolving https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/153. This should be merged following https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/152 where it can be rebased to pull in the new registry changes.

Based on the information in the issue I set the name to `otel-autoinstrumentation-go` which at 27 characters seems a bit verbose. Happy to change it to something else.

- Closes #153
- Blocked by #152

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -9,7 +9,8 @@ jobs:
     permissions:
       contents: read
       packages: write
-
+    env:
+      IMAGE_NAME: otel-autoinstrumentation-go
     steps:
       - uses: actions/checkout@v3.5.2
 
@@ -30,7 +31,7 @@ jobs:
         id: meta
         uses: docker/metadata-action@v4.4.0
         with:
-          images: ghcr.io/${{ github.repository }}/autoinstrumentation-go
+          images: ghcr.io/${{ github.repository }}/${{ env.IMAGE_NAME }}
 
       - name: Build and push
         uses: docker/build-push-action@v4.0.0
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -14,6 +14,12 @@ jobs:
     steps:
       - uses: actions/checkout@v3.5.2
 
+      - name: Log in to Docker Hub
+        uses: docker/login-action@v2.1.0
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+
       - name: Log in to the Container registry
         uses: docker/login-action@v2.1.0
         with:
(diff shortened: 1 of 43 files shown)


edeNFed (approved):
Lgtm

MikeGoldsmith (approved):
Looks great, thanks @MasonLegere 👍🏻 

PS I updated your PR description to automatically close #153 when this is merged.

pellared (changes requested):
`e2e-tests / kubernetes-test` are still failing

MikeGoldsmith:
@MasonLegere This is failing because the sample job isn't terminating correctly, you can see it [here](https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/main/.github/workflows/e2e/k8s/sample-job.yml#L25).

My guess is that the kill command isn't finding the go instrumentation container to kill properly. Maybe there's more references to `otel-go-instrumentation` that need updating.

.github/workflows/release.yaml
```diff
@@ -10,7 +10,7 @@ jobs:
       contents: read
       packages: write
     env:
-      IMAGE_NAME: autoinstrumentation-go
+      IMAGE_NAME: otel-autoinstrumentation-go
```
MrAlias, line 13: This will make the docker image `otel/otel-autoinstrumentation-go`. Do we want the stutter in the name?

MikeGoldsmith: Hmm, no I don't think the additional `otel` prefix.

MikeGoldsmith:
I'm going to update this PR to latest & fix build issues 👍🏻

MrAlias:
This PR looks abandoned, closing. Please re-open if you are able to update this and address all the feedback.

PR #156, opened 2023-05-12
[chore] Fix linter error - use copy instead of loop

rogercoll:
On arm:

```sh
$ make
golangci-lint --fix .
pkg/process/ptrace/ptrace_linux_arm64.go:74:2: S1001: should use copy(to[:], from) instead of a loop (gosimple)
	for index, arg := range args {
	^
make: *** [Makefile:68: golangci-lint/.] Error 1
```

pkg/process/ptrace/ptrace_linux_arm64.go:74:2: S1001: should use copy(to[:], from) instead of a loop (gosimple)

### DIFF
--- a/pkg/process/ptrace/ptrace_linux_arm64.go
+++ b/pkg/process/ptrace/ptrace_linux_arm64.go
@@ -26,7 +26,7 @@ var endian = binary.LittleEndian
 
 const syscallInstrSize = 4
 
-// see kernel source /include/uapi/linux/elf.h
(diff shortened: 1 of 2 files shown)


MikeGoldsmith (approved):
Confirmed this works by executing the nethttp, gorillamux and gin fixtures on my M1 (arm64) laptop.

Thanks @rogercoll 👍🏻

MrAlias (approved):

PR #150, opened 2023-05-08
Fix gRPC instrumentation on recent kernels

edeNFed:
In recent kernel versions, the eBPF verifier checks that a valid number of bytes (positive and not overflow) are being written/read from memory.
This PR adds additional bound checks so gRPC instrumentation now passes verification.

Tested on Ubuntu 20.02, x86, running in DigitalOcean.

- Closes https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/78
- Closes https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/141

PR #143, opened 2023-05-04
update http span names to include method and route

JamieDanielson:
Current [semantic conventions](https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/http/#name) state "HTTP server span names SHOULD be `{http.method} {http.route}`".

This PR adjusts http span names from `{http.route}` to `{http.method} {http.route}`. Also updated are the `traces.json` files to show the new name.

PR #158, opened 2023-05-15
Release v0.2.1-alpha

MikeGoldsmith:
### Fixed

- Fix gRPC instrumentation memory access issue on newer kernels. ([#150](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/150))
- Fix missing spans in gorillamux instrumentation. ([#86](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/86))

### Changed

- Update HTTP span names to include method and route to match semantic conventions. ([#143](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/143))

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ OpenTelemetry Go Automatic Instrumentation adheres to [Semantic Versioning](http
 
 ## [Unreleased]
 
-## [v0.3.0-alpha] - 2023-05-15
+## [v0.2.1-alpha] - 2023-05-15
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,6 @@ OpenTelemetry Go Automatic Instrumentation adheres to [Semantic Versioning](http
 
 ### Changed
 
-- Only pull docker image if not present for the emojivoto example. ([#149](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/149))
 - Update HTTP span names to include method and route to match semantic conventions. ([#143](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/143))
 
 ## [v0.2.0-alpha] - 2023-05-03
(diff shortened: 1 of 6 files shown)


pellared:
I suggest removing the following line from changelog https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/a12d31ccb6ed15e7ce073a6b2d9622d6832dd225/CHANGELOG.md?plain=1#L18

pellared:
nit: Do we have any rules saying what should be the new version number? I thought that this would rather be a `0.2.1` as there are no breaking changes. My assumption was taken from a rule that we have in .NET auto-instrumentation: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/main/docs/versioning.md#version-numbers-before-100 Do we want to follow a similar pattern?

MikeGoldsmith:
I did consider removing it from the changelog because it is not a functional change but as it didn't have skip changelog it was kept. I'm happy to remove it.

I'm also okay to set version as 0.2.1-alpha. I don't think versions mean all that much at the moment. Regarding no breaking changes, there wasn't any breaking changes when we went from 0.1 to 0.2, I think we did it to help signify the bump in compatibility (arm support) & features (gin support).

MrAlias (commented):
Can we add a changelog entry for this: https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/152?

MrAlias:
This looks ready to merge: https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/154 

Should we merge it before this to include it in the release?

MrAlias (approved):

pellared:
> Regarding no breaking changes, there wasn't any breaking changes when we went from 0.1 to 0.2, I think we did it to help signify the bump in compatibility (arm support) & features (gin support).

This is not true. We did one user-facing breaking change: https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/b29eef681a0b3865c4f1948512f07c033dabe8c7/CHANGELOG.md?plain=1#L35-L36

PR #182, opened 2023-05-28
Reduce docker image size

edeNFed:
This PR changes the base image to be `gcr.io/distroless/base`
Image size before the change: **172MB**
Image size after the change: **23.8MB**


Dockerfile
```diff
@@ -7,6 +7,6 @@ WORKDIR /app
 COPY . .
 RUN make build
 
-FROM registry.fedoraproject.org/fedora-minimal:38
+FROM cgr.dev/chainguard/glibc-dynamic:latest
```
pellared, line 10: Can we use something else than `latest`? Mainly in order to get reproducible builds, but also to have more control on what we publish.

Unfortunately, I think that `cgr.dev/chainguard/glibc-dynamic` does not have SemVer tags, then we would need to use the digest. We would need to manually update the images e.g. each week (side note: AFAIK this approach is actually seen as a better practice).

References: 
- https://vsupalov.com/docker-latest-tag/
- https://medium.com/@michael.vittrup.larsen/why-we-should-use-latest-tag-on-container-images-fc0266877ab5
- https://hackernoon.com/docker-images-name-vs-tag-vs-digest

pellared, line 10: Are we confident that we can trust/rely on Chainguard's artifacts?

MrAlias: From SIG meeting:

- the project has its source published
- @edeNFed is going to link the upstream Dockerfile so it can be reviewed
- @pellared can you post issues you have found with chainguard artifacts?

pellared: > @edeNFed is going to link the upstream Dockerfile so it can be reviewed

The images are not created using Dockerfiles. Therefore the review is not easy. The images are creating using https://github.com/chainguard-dev/apko. The OS used by the images is https://github.com/wolfi-dev/ (also maintained by Chainguard)

> @pellared can you post issues you have found with chainguard artifacts?

Personally, I have not found any issues but I do not feel also competent enough 😢 


```
$ docker scout recommendations cgr.dev/chainguard/glibc-dynamic:latest
    ✓ Image stored for indexing
    ✓ Indexed 10 packages

Could not display recommendations for image  cgr.dev/chainguard/glibc-dynamic:latest 
image has no base image

$ docker scout cves cgr.dev/chainguard/glibc-dynamic:latest
    ✓ SBOM of image already cached, 10 packages indexed
    ✓ No vulnerable package detected


$ docker scout quickview cgr.dev/chainguard/glibc-dynamic:latest
    ✓ SBOM of image already cached, 10 packages indexed
ERROR   Status: image has no base image, Code: 1

$ trivy image cgr.dev/chainguard/glibc-dynamic:latest
2023-05-31T09:38:51.084+0200    INFO    Need to update DB
2023-05-31T09:38:51.084+0200    INFO    DB Repository: ghcr.io/aquasecurity/trivy-db
2023-05-31T09:38:51.084+0200    INFO    Downloading DB...
37.36 MiB / 37.36 MiB [------------------------------------------------------------------------------------------------------------------] 100.00% 8.19 MiB p/s 4.8s
2023-05-31T09:38:59.716+0200    INFO    Vulnerability scanning is enabled
2023-05-31T09:38:59.716+0200    INFO    Secret scanning is enabled
2023-05-31T09:38:59.716+0200    INFO    If your scanning is slow, please try '--scanners vuln' to disable secret scanning
2023-05-31T09:38:59.716+0200    INFO    Please see also https://aquasecurity.github.io/trivy/dev/docs/secret/scanning/#recommendation for faster secret detection
2023-05-31T09:38:59.977+0200    INFO    Detected OS: wolfi
2023-05-31T09:38:59.977+0200    INFO    Detecting Wolfi vulnerabilities...
2023-05-31T09:38:59.977+0200    INFO    Number of language-specific files: 0

cgr.dev/chainguard/glibc-dynamic:latest (wolfi 20230201)

Total: 0 (UNKNOWN: 0, LOW: 0, MEDIUM: 0, HIGH: 0, CRITICAL: 0)
```

I think that maybe also because of the way the images are created they are hard to be diagnosed by container image scanning tools.

The OS used by the images is https://github.com/wolfi-dev/ (also maintained by Chainguard).

Personally, I do not see any good reason to have confidence on using Chainguard's products. Especially in the light of:

image

_Source: https://www.chainguard.dev/chainguard-images_

How about trying to use Alpine and add `apk add gcompat`?

MikeGoldsmith:
We have set dependency requirements for the runtime image, maybe looking to build our own using the ChainGuard image as a guide would be better? Then we'd at least be in control of dependency updates and publish schedules & versioning.

edeNFed:
@pellared @MrAlias 
This is ready for another review.
Changed from the Chainguard image to Google distroless and used the sha256 digest as version

dineshg13 (approved):
LTGM

MrAlias:
For reference, I think this is the repository the images are generated from: https://github.com/GoogleContainerTools/distroless

MrAlias:
It looks like there are support timelines: https://github.com/GoogleContainerTools/distroless/blob/main/SUPPORT_POLICY.md

MrAlias:
[Docs](https://github.com/GoogleContainerTools/distroless/blob/main/base/README.md):

> ## Image Contents
> 
> This image contains a minimal Linux, glibc-based system. It is intended for use directly by "mostly-statically compiled" languages like Go, Rust or D.
> 
> Statically compiled applications (Go) that do not require libc can use the `gcr.io/distroless/static` image, which contains:
> 
> * ca-certificates
> * A /etc/passwd entry for a root user
> * A /tmp directory
> * tzdata
> 
> Most other applications (and Go apps that require libc/cgo) should start with `gcr.io/distroless/base`, which contains all
> of the packages in `gcr.io/distroless/static`, and 
> 
> * glibc
> * libssl
> * openssl

Dockerfile
```diff
@@ -1,12 +1,12 @@
...
-RUN dnf install clang llvm make libbpf-devel -y
+RUN apt-get update && apt-get install -y curl clang gcc llvm make libbpf-dev -y
 RUN curl -LO https://go.dev/dl/go1.20.linux-${TARGETARCH}.tar.gz && tar -C /usr/local -xzf go*.linux-${TARGETARCH}.tar.gz
 ENV PATH="/usr/local/go/bin:${PATH}"
 WORKDIR /app
 COPY . .
 RUN make build
 
-FROM registry.fedoraproject.org/fedora-minimal:38
+FROM gcr.io/distroless/base@sha256:73deaaf6a207c1a33850257ba74e0f196bc418636cada9943a03d7abea980d6d
```
MrAlias, line 10: [> ### Base Operating System
> 
> Distroless images are based on Debian 11 (bullseye). Images are explicitly tagged with Debian version suffixes (e.g. `-debian11`). Specifying an image without the distribution will currently select `-debian11` images, but that will change in the future to a newer version of Debian. It can be useful to reference the distribution explicitly, to prevent breaking builds when the next Debian version is released.](https://github.com/GoogleContainerTools/distroless/tree/main#base-operating-system)

To avoid breaking changes, let's pin the underlying distro.

```suggestion
FROM gcr.io/distroless/base-debian11@sha256:73deaaf6a207c1a33850257ba74e0f196bc418636cada9943a03d7abea980d6d
```

MrAlias (approved):

damemi (approved):

Issue #197, opened 2023-06-14
Auto Instrumentation - Monitor multiple processes 


dineshg13:
**Is your feature request related to a problem? Please describe.**
Our auto instrumentation system is currently limited to monitoring a single process. This approach poses an issue when dealing with a large cluster containing 1000 pods, for example, as it would necessitate deploying an instrumentation service as a sidecar for each pod. We can reduce our resource utilization by allowing a single instance of the auto instrumentation service to supervise numerous services.

**Describe the solution you'd like**
We want the auto instrumentation service to be able to monitor multiple processes. To achieve this 

- Make OTEL_GO_AUTO_TARGET_EXE optional 
- Use https://github.com/google/gops to monitor for new processes on the machine. 
- Instrument the new process 
- Have an exclude list to ignore processes for auto instrumentation 

This approach involves significant code refactoring. We need to approach this in multiple PRs. Some concerns from slack 
* Ensure we have  a single OTel exporter for all the processes vs. running one instance per process
* Watching for terminated processes and clean up the eBPF programs/maps for the terminated process
* How can we make env vars like OTEL_SERVICE_NAME work when dealing with multiple processes?

dineshg13:
Existing code:
<img width="1314" alt="image" src="https://github.com/open-telemetry/opentelemetry-go-instrumentation/assets/2471669/f1b5ff6d-279b-48ff-8ac5-7eed3c47a931">
Proposed Changes:
* Add orchestrator that splits the responsibility 
<img width="1206" alt="image" src="https://github.com/open-telemetry/opentelemetry-go-instrumentation/assets/2471669/5c8d8087-6dfc-429d-94f8-32836eb50de4">

Plan for getting this across

dianashevchenko:
@edeNFed @RonFed I hope you're doing well. I wanted to check in and ask if there are any plans to prioritize this issue in the near future. Your work is greatly appreciated, and any updates would be helpful 🙇

PR #208, opened 2023-06-25
Add orchestrator for supporting multiple processes

dineshg13:
changes necessary to resolve https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/197 . 
This PR  moves the logic from `main`, and creates a new orchestrator pkg for handling multiple processes.


dineshg13:
@edeNFed This PR is working on baremetal and as sidecar for a pod. But when i run same as a deployment in k8s, am getting error below error. Can you please help with same. 

```
{"level":"info","ts":1687914965.2464294,"caller":"cli/main.go:60","msg":"starting Go OpenTelemetry Agent ..."}
{"level":"info","ts":1687914965.2465029,"caller":"cli/main.go:62","msg":"Establishing connection to OTLP receiver ..."}
{"level":"info","ts":1687914967.2483935,"caller":"orchestrator/orchestrator.go:151","msg":"No go process not found yet, trying again in 2 seconds"}
{"level":"info","ts":1687914969.2487526,"caller":"orchestrator/orchestrator.go:151","msg":"No go process not found yet, trying again in 2 seconds"}
{"level":"info","ts":1687914971.24835,"caller":"orchestrator/orchestrator.go:151","msg":"No go process not found yet, trying again in 2 seconds"}
{"level":"info","ts":1687914973.247508,"caller":"orchestrator/orchestrator.go:151","msg":"No go process not found yet, trying again in 2 seconds"}
{"level":"info","ts":1687914975.2487097,"caller":"orchestrator/orchestrator.go:97","msg":"Add auto instrumetors","pid":2247,"serviceName":"sample-app"}
{"level":"info","ts":1687914975.250355,"caller":"ptrace/ptrace_linux.go:102","msg":"retry attaching thread","tid":2247,"retryCount":1,"limit":10,"error":"no such process"}
{"level":"info","ts":1687914975.2504175,"caller":"ptrace/ptrace_linux.go:102","msg":"retry attaching thread","tid":2248,"retryCount":1,"limit":10,"error":"no such process"}
{"level":"info","ts":1687914975.2504244,"caller":"ptrace/ptrace_linux.go:102","msg":"retry attaching thread","tid":2249,"retryCount":1,"limit":10,"error":"no such process"}
{"level":"info","ts":1687914975.2504296,"caller":"ptrace/ptrace_linux.go:102","msg":"retry attaching thread","tid":2250,"retryCount":1,"limit":10,"error":"no such process"}
{"level":"info","ts":1687914975.2504346,"caller":"ptrace/ptrace_linux.go:102","msg":"retry attaching thread","tid":2251,"retryCount":1,"limit":10,"error":"no such process"}
{"level":"info","ts":1687914975.25044,"caller":"ptrace/ptrace_linux.go:102","msg":"retry attaching thread","tid":2252,"retryCount":1,"limit":10,"error":"no such process"}
{"level":"info","ts":1687914975.250455,"caller":"ptrace/ptrace_linux.go:102","msg":"retry attaching thread","tid":2247,"retryCount":2,"limit":10,"error":"no such process"
```

cli/main.go
```diff
@@ -15,18 +15,41 @@
...
+	"go.opentelemetry.io/auto"
 	"go.opentelemetry.io/auto/pkg/log"
-	"go.opentelemetry.io/auto/pkg/opentelemetry"
+	"go.opentelemetry.io/auto/pkg/orchestrator"
 	"go.opentelemetry.io/auto/pkg/process"
 )
 
+var (
+	// Controller-local reference to the auto-instrumentation release version.
+	releaseVersion = auto.Version()
```
damemi, line 39: `releaseVersion` is only used in the line below, do we need this variable?

pkg/process/args.go
```diff
@@ -40,12 +46,14 @@ func (t *TargetArgs) Validate() error {
...
 		result.ExePath = val
+		serviceName, exists := os.LookupEnv(otelServiceNameEnvVar)
+		if exists {
+			result.ServiceName = serviceName
+		}
+		return result
 	}
-
-	return result
+	return nil
```
damemi, line 58: you return `nil` here if the exePath isn't set, which is invalid config, but then you can't call `Validate()` to catch it

maybe change this like:

```go
    result := &TargetArgs{}
	if val, exists := os.LookupEnv(ExePathEnvVar); exists {
		result.ExePath = val
	}
    if serviceName, exists := os.LookupEnv(otelServiceNameEnvVar); exists {
		result.ServiceName = serviceName
	}
	return result
```

breaking the `serviceName` check out of the nested exePath check is a little easier to read too. you could also fail earlier if exePath isn't set, but we have Validate() functions for that anyway

pkg/instrumentors/manager.go
```diff
@@ -30,26 +30,24 @@ import (
...
+	otelController *opentelemetry.Controller
 }
 
 // NewManager returns a new [Manager].
 func NewManager(otelController *opentelemetry.Controller) (*Manager, error) {
 	m := &Manager{
 		instrumentors:  make(map[string]Instrumentor),
 		done:           make(chan bool, 1),
-		incomingEvents: make(chan *events.Event),
+		incomingEvents: make(chan *events.Event, 10),
```
damemi, line 50: why is this buffered to 10?

pkg/opentelemetry/controller.go
```diff
@@ -98,35 +77,23 @@ func (c *Controller) convertTime(t int64) time.Time {
...
-	serviceName, exists := os.LookupEnv(otelServiceNameEnvVar)
-	if !exists {
-		return nil, fmt.Errorf("%s env var must be set", otelServiceNameEnvVar)
-	}
-
-	ctx := context.Background()
+func NewController(
+	ctx context.Context,
+	serviceName string,
+	exporter sdktrace.SpanExporter,
```
damemi, line 83: not a blocker yet since this isn't stable but we might want to consider a different approach to passing params in if this list could grow

pkg/process/discover.go
```diff
@@ -91,30 +93,55 @@ func (a *Analyzer) findProcessID(target *TargetArgs) (int, error) {
...
-					return pid, nil
-				}
-			} else if exeName == target.ExePath {
-				return pid, nil
+			exe := filepath.Base(exeFullPath)
+
+			if _, ok := a.ignoreProcesses[exe]; ok {
+				continue
+			}
+			if target != nil && !strings.Contains(string(exeFullPath), target.ExePath) {
```
damemi, line 116: If I'm running in single-target mode, will this keep looping through processes after the target exe has been found?

damemi:
> @edeNFed This PR is working on baremetal and as sidecar for a pod. But when i run same as a deployment in k8s, am getting error below error. Can you please help with same.

@dineshg13 could you share the deployment yaml you're trying?

pkg/process/discover.go
dineshg13: Yes. Currently, once we have the pid and start monitoring the process. If the process dies, auto-instrumentation doesn't try to find the new pid. It doesn't allow us to run on bare metal or as a deployment.

dineshg13:
@damemi I have converted our test case from side car to a deployment. 

`auto-instrumentation - pod `
```
apiVersion: v1
kind: Pod
metadata:
  name: "auto-instrumentation"
  namespace: default
  labels:
    app: "auto-instrumentation"
spec:
  containers:
    - name: auto-instrumentation
      image: otel-go-instrumentation
      imagePullPolicy: IfNotPresent
      env:
        - name: OTEL_GO_AUTO_TARGET_EXE
          value: /sample-app/main
        - name: OTEL_EXPORTER_OTLP_ENDPOINT
          value: "http://test-opentelemetry-collector:4317"
        - name: OTEL_SERVICE_NAME
          value: "sample-app"
        - name: OTEL_PROPAGATORS
          value: "tracecontext,baggage"
      resources: {}
      securityContext:
        runAsUser: 0
        runAsGroup: 0
        capabilities:
          add:
            - SYS_PTRACE
        privileged: true
      volumeMounts:
        - mountPath: /sys/kernel/debug
          name: kernel-debug
        - mountPath: /proc
          name: hostproc
  volumes:
    - name: kernel-debug
      hostPath:
        path: /sys/kernel/debug
    - name: hostproc
      hostPath:
        path: /proc
  restartPolicy: Always
---
```
our app
```
apiVersion: v1
kind: Pod
metadata:
  name: "sample-app"
  namespace: default
  labels:
    app: "sample-app"
spec:
  containers:
    - name: sample-app
      image: sample-app
      imagePullPolicy: IfNotPresent
      command: ["/bin/sh", "-c"]
      args: ["/sample-app/main"]
      ports:
        - containerPort: 8080
          name: http
  restartPolicy: Always
---
```

edeNFed:
@dineshg13 please try to use `hostPID: true` instead of mounting the `/proc` directory

pkg/instrumentors/manager.go
dineshg13: am surprised that we didn't had buffer before. Without buffer, the process would wait for GRPC request to complete.  This would impact performance of the process. I choose 10 at random, but ideally we should make this configurable.

pkg/opentelemetry/controller.go
dineshg13: Updated . Added controller settings.

edeNFed:
Looks good @dineshg13 
I think we should merge #211 before this one, so we will avoid overwriting different BPF structures when instrumenting multiple processes.

cli/main.go
```diff
@@ -35,56 +58,49 @@ func main() {
...
-
-	instManager, err := instrumentors.NewManager(otelController)
+	targetArgs := process.ParseTargetArgs()
+	if targetArgs != nil {
+		if err := targetArgs.Validate(); err != nil {
+			log.Logger.Error(err, "invalid target args")
+			return
+		}
+	}
+	r, err := orchestrator.New(ctx, targetArgs, traceExporter)
```
damemi, line 78: what if `targetArgs` is nil here?

damemi:
> @damemi i prefer to add an e2e test in a follow-up PR. This PR is already big and backward compatible with existing functionality. The current test framework has only k8s sidecar, I want to refactor the tests to support all 3 paths viz sidecar, deployment & bare-metal. Let me know your thoughts.

I think it would be a lot better to have some testing for the new functionality before merging. You should be able to use the workflow matrix to deploy a different set of manifests for the sample app + instrumentation. I'm happy to help if you have any questions

pkg/instrumentors/manager.go
damemi: cool, yeah maybe some benchmarking tests would be good to have at some point

pkg/process/discover.go
damemi: I mean, if I set a targetExe will this loop continue searching all pids even after the target pid is found (unnecessarily)? Would you maybe want to update L123 to something like:

```go
pids[pid] = exe
if strings.Contains(string(exeFullPath), target.ExePath) {
  break
}
```

pkg/process/args.go
damemi: what about an explicit field in `targetArgs` that says you want to monitor all services (eg `targetArgs.allTargets = true`) instead of relying on `nil` to imply that's what you're doing?

dineshg13: That is good idea. Will add it .

dineshg13:
@damemi i added bare-metal tests to the workflow. Currently the tests are failing on service-name. This might require some discussion, I have added this agenda to next SIG meeting.

damemi:
@dineshg13 thanks, the bare metal test is a good idea and probably wouldn't fit with the kind tests since the setup is so different.

If it helps, I put together a commit https://github.com/damemi/opentelemetry-go-instrumentation/pull/10/commits/9ef0f4d6e5dd24454e21b9390cde1e0834300fd8 based on your branch that shows what I mean for the deployment tests. Feel free to pick this commit to this PR or I can push it to your branch if you'd like (not sure I have permission). The tests pass in a test PR https://github.com/damemi/opentelemetry-go-instrumentation/pull/10

Right now, I think this would be good enough to show everything working in the new feature. The fixtures can be shared between deployment+sidecar modes because we expect the output to be identical. But if we add more configurability (like service name), we should look at restructuring the fixtures. This can be part of a broader refactor for the e2e framework.

dineshg13:
@damemi Thanks. I was able to cherry-pick the commit and push the same. Ideally, for both deployment & bare-metal we should run one instance of collector for multiple apps. We can do that refactor later as well. We should make the service configurable via yaml instead of env vars. We can discuss this tomorrow.

MrAlias:
Specification reference for service name: https://github.com/open-telemetry/opentelemetry-specification/tree/bb1aa87ed221c16fc915a9fbfbfa262b989e1f79/specification/resource/semantic_conventions#service

dineshg13:
@damemi @MrAlias @edeNFed  The PR is ready for review. I changed the code to read `OTEL_SERVICE_NAME` from the process envars. I have made it mandatory, and auto-instrumenation would only monitor processes with `OTEL_SERVICE_NAME` defined. It also removes the need to have ignore-processes list. Let me know your thoughts.

pkg/orchestrator/orchestrator.go
```diff
@@ -0,0 +1,169 @@
...
+			for _, m := range i.managers {
+				m.Close()
+			}
+
+			close(i.deadProcess)
+			close(i.processch)
+
+			return nil
+		case d := <-i.deadProcess:
+			log.Logger.Info("process died cleaning up", "pid", d)
```
edeNFed, line 85: Need to clean BPFFS for dead processes. This can be done by calling `bpffs.Cleanup` with the dead PID

edeNFed:
@dineshg13 is this ready for another review?

dineshg13:
@edeNFed yes.

edeNFed (approved):

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ OpenTelemetry Go Automatic Instrumentation adheres to [Semantic Versioning](http
 
 ## [Unreleased]
 
+- Add ability to monitor multiple processes. ([#197](https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/197))
```
MrAlias, line 11:
```suggestion
### Added

- Add ability to monitor multiple processes. ([#197](https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/197))
```

pkg/process/ptrace/ptrace_linux.go
```diff
@@ -99,7 +99,17 @@ func NewTracedProgram(pid int, logger logr.Logger) (*TracedProgram, error) {
...
+					logger.Info(
+						"retry attaching thread",
+						"tid",
+						tid,
+						"retryCount",
+						retryCount[tid],
+						"limit",
+						threadRetryLimit,
+						"error",
+						err,
```
MrAlias, line 111: If this error is not `nil` shouldn't this be logged as an error?

pkg/process/ptrace/ptrace_linux.go
```diff
@@ -208,5 +217,13 @@ func (p *TracedProgram) Step() error {
...
-	return p.Syscall(syscall.SYS_MMAP, 0, length, syscall.PROT_READ|syscall.PROT_WRITE|syscall.PROT_EXEC, syscall.MAP_ANON|syscall.MAP_PRIVATE|syscall.MAP_POPULATE, fd, 0)
+	return p.Syscall(
+		syscall.SYS_MMAP,
+		0,
+		length,
+		syscall.PROT_READ|syscall.PROT_WRITE|syscall.PROT_EXEC,
+		syscall.MAP_ANON|syscall.MAP_PRIVATE|syscall.MAP_POPULATE,
+		fd,
+		0,
+	)
```
MrAlias, line 228: This change does not look related to the PR:

```suggestion
	return p.Syscall(syscall.SYS_MMAP, 0, length, syscall.PROT_READ|syscall.PROT_WRITE|syscall.PROT_EXEC, syscall.MAP_ANON|syscall.MAP_PRIVATE|syscall.MAP_POPULATE, fd, 0)
```

pkg/process/discover.go
```diff
@@ -91,30 +77,79 @@ func (a *Analyzer) findProcessID(target *TargetArgs) (int, error) {
...
+		return false
+	}
+
+	_, _, err = a.getModuleDetails(elfF)
+
+	return err == nil
+}
+
+func getEnvVars(pid int) (map[string]string, error) {
+	bytes, err := ioutil.ReadFile(fmt.Sprintf("/proc/%d/environ", pid))
```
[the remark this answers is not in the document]
dineshg13: Our ebpf auto instrumentation only works on linux systems and it needs `/proc` to be mounted. For Linux systems we support, this should be good.

pkg/process/discover.go
```diff
@@ -91,30 +77,79 @@ func (a *Analyzer) findProcessID(target *TargetArgs) (int, error) {
...
 
-	return 0, errors.ErrProcessNotFound
+	return pids
 }
 
-// Close closes the analyzer.
-func (a *Analyzer) Close() {
-	a.done <- true
+func (a *Analyzer) isGo(pid int) bool {
+	// TODO
```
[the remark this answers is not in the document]
dineshg13: removed.

pkg/instrumentors/manager.go
```diff
@@ -84,9 +82,9 @@ func (m *Manager) GetRelevantFuncs() map[string]interface{} {
 	return funcsMap
 }
 
-// FilterUnusedInstrumentors filterers Instrumentors whose functions are
+// filterUnusedInstrumentors filterers Instrumentors whose functions are
 // already instrumented out of the Manager.
-func (m *Manager) FilterUnusedInstrumentors(target *process.TargetDetails) {
+func (m *Manager) filterUnusedInstrumentors(target *process.TargetDetails) {
```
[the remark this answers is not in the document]
dineshg13: Added

pkg/instrumentors/runner.go
```diff
@@ -91,12 +93,12 @@ func (m *Manager) load(target *process.TargetDetails) error {
...
 	}
 
-	log.Logger.V(0).Info("loaded instrumentors to memory", "total_instrumentors", len(m.instrumentors))
+	log.Logger.V(0).
+		Info("loaded instrumentors to memory", "total_instrumentors", len(m.instrumentors))
 	return nil
 }
 
 func (m *Manager) cleanup() {
-	close(m.incomingEvents)
```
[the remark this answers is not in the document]
dineshg13: good catch. Added it back

Makefile
```diff
@@ -132,9 +135,10 @@ fixtures/%:
 	if [ ! -d "opentelemetry-helm-charts" ]; then \
 		git clone https://github.com/open-telemetry/opentelemetry-helm-charts.git; \
 	fi
-	helm install test -f .github/workflows/e2e/k8s/collector-helm-values.yml opentelemetry-helm-charts/charts/opentelemetry-collector
+	helm install test -f .github/workflows/e2e/k8s/$(FIXTURE_MODE)/collector-helm-values.yml opentelemetry-helm-charts/charts/opentelemetry-collector
+	sleep 5
```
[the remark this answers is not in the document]
dineshg13: Fixture tests were failing locally without sleep. The next waits for the container to be ready, but sometimes the container isn't created and fails.

pkg/process/analyze.go
```diff
@@ -144,9 +144,6 @@ func (a *Analyzer) Analyze(pid int, relevantFuncs map[string]interface{}) (*Targ
 		EndAddr:   addr + mapSize,
 	}
 
-	if err != nil {
-		return nil, err
-	}
```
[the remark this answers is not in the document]
dineshg13: we already have `err` check above. after the check `err` isn't assigned

pkg/orchestrator/orchestrator.go
```diff
@@ -0,0 +1,176 @@
...
+	exporter    sdktrace.SpanExporter
+	pidTicker   <-chan time.Time
+}
+
+// New creates a new Implementation of orchestrator Interface.
+func New(
+	ctx context.Context,
+	targetArgs *process.TargetArgs,
+	exporter sdktrace.SpanExporter,
+) (Interface, error) {
```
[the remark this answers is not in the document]
dineshg13: Updated. Added Service struct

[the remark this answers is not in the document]
dineshg13: I am using https://github.com/segmentio/golines for formatting. is it not recommended ?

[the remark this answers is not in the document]
MrAlias: It does not match the rest of the project or the other OTel Go projects. I would not use that formater if it produces this syntax.

CHANGELOG.md
```diff
@@ -8,20 +8,27 @@ OpenTelemetry Go Automatic Instrumentation adheres to [Semantic Versioning](http
...
+
 ### Changed
 
+- `pkg/instrumentors` Manager struct method  FilterUnusedInstrumentors is made private.
+- `pkg/opentelemtry/` NewController method now accepts ControllerSettings param.
+- `pkg/process/` Analyzer struct `Close` method is removed.
 - The function signature of `"go.opentelemetry.io/auto/offsets-tracker/downloader".DownloadBinary` has changed.
-  It now has an additional flag indicating whether it'll build a dummy app for Go stdlib packages or not. ([#256]https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/256)
+  It now has an additional flag indicating whether it'll build a dummy app for Go stdlib packages or not. ([#256]<https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/256>)
```
MrAlias, line 21:
```suggestion
  It now has an additional flag indicating whether it'll build a dummy app for Go stdlib packages or not. ([#256](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/256))
```

CHANGELOG.md
```diff
@@ -8,20 +8,27 @@ OpenTelemetry Go Automatic Instrumentation adheres to [Semantic Versioning](http
...
+- Use UPROBE_RETURN to declare the common uprobe return logic (finding the corresponding context, setting up end time, and sending the event via perf buffer) ([#257]<https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/257>)
+- BASE_SPAN_PROPERTIES as common fields (start time, end time, SpanContext and ParentSpanContext) for all instrumentations events (consistent between C and Go structs). ([#257]<https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/257>)
+- Header guards in eBPF code. ([#257]<https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/257>)
 
 ### Fixed
 
 - Fix context propagation across different goroutines. ([#118](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/118))
-- The offset tracker can once again build binaries for the Go stdlib. ([#256]https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/256)
+- The offset tracker can once again build binaries for the Go stdlib. ([#256]<https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/256>)
```
MrAlias, line 31:
```suggestion
- The offset tracker can once again build binaries for the Go stdlib. ([#256](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/256))
```

CHANGELOG.md
```diff
@@ -8,20 +8,27 @@ OpenTelemetry Go Automatic Instrumentation adheres to [Semantic Versioning](http
...
 
+### Added
+
+- Add ability to monitor multiple processes. ([#197](https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/197))
+
 ### Changed
 
+- `pkg/instrumentors` Manager struct method  FilterUnusedInstrumentors is made private.
+- `pkg/opentelemtry/` NewController method now accepts ControllerSettings param.
+- `pkg/process/` Analyzer struct `Close` method is removed.
```
MrAlias, line 19:
```suggestion
- The `FilterUnusedInstrumentors` method of  `"go.opentelemetry.io/auto/pkg/instrumentors".Manager` is removed. ([#208](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/208))
- The function signature of `"go.opentelemetry.io/auto/pkg/opentelemtry".NewController` has changed.
  It now accepts `context.Context` and `ControllerSettings` parameters. ([#208](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/208))
- The `Close` method of `"go.opentelemetry.io/auto/pkg/process".Analyzer` is removed. ([#208](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/208))
```

Makefile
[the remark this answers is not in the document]
dineshg13: We already have `kubectl` wait for condition below. Wait for condition is failing when i run this locally.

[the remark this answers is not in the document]
MrAlias: The issue with adding sleep is that is going to be brittle (it may work on you machine, but fail on others) and they accumulate.

I'm surprised the `wait` command would fail, but I'm also curious if the `rollout` command works better for you.

pkg/orchestrator/orchestrator.go
```diff
@@ -0,0 +1,176 @@
...
+	exporter    sdktrace.SpanExporter
+	pidTicker   <-chan time.Time
+}
+
+// New creates a new Implementation of orchestrator Service.
+func New(
+	ctx context.Context,
+	targetArgs *process.TargetArgs,
+	exporter sdktrace.SpanExporter,
+) (*Service, error) {
```
MrAlias, line 53:
```suggestion
func New(ctx context.Context, targetArgs *process.TargetArgs, exporter sdktrace.SpanExporter) (*Service, error) {
```

pkg/opentelemetry/controller.go
```diff
@@ -98,35 +83,22 @@ func (c *Controller) convertTime(t int64) time.Time {
...
-	serviceName, exists := os.LookupEnv(otelServiceNameEnvVar)
-	if !exists {
-		return nil, fmt.Errorf("%s env var must be set", otelServiceNameEnvVar)
-	}
-
-	ctx := context.Background()
+func NewController(
+	ctx context.Context,
+	settings ControllerSetting,
+) (*Controller, error) {
```
MrAlias, line 89:
```suggestion
func NewController(ctx context.Context, settings ControllerSetting) (*Controller, error) {
```

MrAlias:

pkg/instrumentors/manager.go
dineshg13: Added https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/334

Makefile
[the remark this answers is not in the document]
dineshg13: Converted to `until`

cli/main.go
```diff
@@ -34,30 +30,15 @@ func main() {
...
-		case <-ch:
-			cancel()
-		case <-ctx.Done():
-		}
-	}()
-
-	log.Logger.V(0).Info("starting instrumentors...")
-	if err = inst.Run(ctx); err != nil && err != errors.ErrInterrupted {
-		log.Logger.Error(err, "instrumentation crashed")
+	log.Logger.V(0).Info("starting Go OpenTelemetry Agent ...")
```
MrAlias, line 40:
```suggestion
	log.Logger.V(0).Info("starting OpenTelemetry Go Agent ...")
```

instrumentation.go
```diff
@@ -65,64 +62,62 @@ func NewInstrumentation(opts ...InstrumentationOption) (*Instrumentation, error)
...
-		return nil, err
-	}
-
-	ctrl, err := opentelemetry.NewController(Version(), c.serviceName)
-	if err != nil {
-		return nil, err
-	}
 
-	mngr, err := instrumentors.NewManager(ctrl)
+	ctx := contextWithSigterm(context.Background())
```
MrAlias, line 66: The change in context handling seems incorrect to me. We want users to still be able to cancel a `Run` meaning they need to be able to pass a context to that method.

internal/pkg/orchestrator/orchestrator.go
```diff
@@ -0,0 +1,221 @@
...
+)
+
+type pidServiceName struct {
+	serviceName string
+	pid         int
+}
+
+// New creates a new Implementation of orchestrator Service.
+func New(
+	ctx context.Context,
```
MrAlias, line 37: This should be passed directly to `Run`.

internal/pkg/orchestrator/service.go
```diff
@@ -0,0 +1,209 @@
...
+
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	semconv "go.opentelemetry.io/otel/semconv/v1.4.0"
+
+	"go.opentelemetry.io/auto/internal/pkg/instrumentors"
+	"go.opentelemetry.io/auto/internal/pkg/process"
+)
+
+// Service is responsible for managing all instrumentation.
+type Service struct {
```
MrAlias, line 34: Why is service defined across both this file and `orchestrator.go`? Can you merge the type definition and its methods into a single file?

internal/pkg/orchestrator/orchestrator.go
```diff
@@ -0,0 +1,221 @@
...
+	"go.opentelemetry.io/auto/internal/pkg/process"
+)
+
+type pidServiceName struct {
+	serviceName string
+	pid         int
+}
+
+// New creates a new Implementation of orchestrator Service.
+func New(
```
MrAlias, line 36:
```suggestion
func NewService(
```

The created type is not named `Orchestrator`.

internal/pkg/orchestrator/orchestrator.go
```diff
@@ -0,0 +1,221 @@
...
+	ctx context.Context,
+	opts ...ServiceOpt,
+) (*Service, error) {
+	s := Service{
+		ctx:         ctx,
+		analyzer:    process.NewAnalyzer(),
+		processch:   make(chan *pidServiceName, 10),
+		deadProcess: make(chan int, 10),
+		managers:    make(map[int]*instrumentors.Manager),
+		pidTicker:   time.NewTicker(2 * time.Second).C,
```
MrAlias, line 46: This ticker looks to never have its `Stop` method called and its resources are never appropriately released.

internal/pkg/orchestrator/service.go
```diff
@@ -0,0 +1,209 @@
...
+}
+
+func WithPID(pid int) ServiceOpt {
+	return fnOpt(func(c Service) Service {
+		c.pid = pid
+		return c
+	})
+}
+
+func WithVersion(version string) ServiceOpt {
```
MrAlias, line 204: These need comments documenting them.

internal/pkg/orchestrator/service.go
```diff
@@ -0,0 +1,209 @@
...
+// If OTEL_SERVICE_NAME is defined it will take precedence over any value
+// passed here.
+func WithServiceName(serviceName string) ServiceOpt {
+	return fnOpt(func(c Service) Service {
+		c.serviceName = serviceName
+		return c
+	})
+}
+
+func WithMonitorAll(monitorAll bool) ServiceOpt {
```
MrAlias, line 183: This seems to always be `true`. Is there a reason we need it as an option?

internal/pkg/orchestrator/service.go
```diff
@@ -0,0 +1,209 @@
...
+	monitorAll  bool
+	processch   chan *pidServiceName
+	deadProcess chan int
+	managers    map[int]*instrumentors.Manager
+	exporter    sdktrace.SpanExporter
+	pid         int
+	pidTicker   <-chan time.Time
+}
+
+type ServiceOpt interface {
```
MrAlias, line 49: Missing documentation.

internal/pkg/orchestrator/service.go
```diff
@@ -0,0 +1,209 @@
...
+		}
+		key := strings.TrimSpace(k)
+		if key == string(semconv.ServiceNameKey) {
+			serviceName = strings.TrimSpace(v)
+		}
+	}
+	return serviceName
+}
+
+func (s Service) Validate() error {
```
MrAlias, line 119: Missing documentation.

internal/pkg/orchestrator/service.go
```diff
@@ -0,0 +1,209 @@
...
+	if s.monitorAll {
+		return nil
+	}
+	if s.serviceName == "" {
+		return fmt.Errorf("serviceName is nil")
+	}
+
+	if s.exePath == "" {
+		return fmt.Errorf("execPath is nil")
+	}
```
MrAlias, line 133: This will only evaluate at most one of these conditionals. Is this intentional?

instrumentation.go
```diff
@@ -65,64 +62,62 @@ func NewInstrumentation(opts ...InstrumentationOption) (*Instrumentation, error)
...
+	log.Logger.V(0).Info("Establishing connection to OTLP receiver ...")
+	otlpTraceClient := otlptracegrpc.NewClient(
+		otlptracegrpc.WithDialOption(grpc.WithUserAgent(autoinstUserAgent)),
+	)
+	traceExporter, err := otlptrace.New(ctx, otlpTraceClient)
```
MrAlias, line 71: Should this be passed as an option instead?

It seems like this could continue to be setup in the controller. It could be changed in a targeted follow up PR instead.

I think changing the exporter setup in this PR is going to bloat this PR. Especially since this doesn't seem like the appropriate way to expose the option to the user.

internal/pkg/instrumentors/runner.go
```diff
@@ -28,27 +27,24 @@ import (
 )
 
 // Run runs the event processing loop for all managed Instrumentors.
-func (m *Manager) Run(ctx context.Context, target *process.TargetDetails) error {
+func (m *Manager) Run(target *process.TargetDetails) error {
```
MrAlias, line 30: Removing the `ctx` arg means this cannot be canceled and shut down appropriately. I don't think that is an appropriate change, right?

instrumentation.go
dineshg13: 👍  Added context to `Run` method

internal/pkg/orchestrator/service.go
```diff
@@ -0,0 +1,209 @@
...
+
+// WithTarget returns an [InstrumentationOption] defining the target binary for
+// [Instrumentation] that is being executed at the provided path.
+//
+// If multiple of these options are provided to an [Instrumentation], the last
+// one will be used.
+//
+// If OTEL_GO_AUTO_TARGET_EXE is defined it will take precedence over any value
+// passed here.
+func WithTarget(path string) ServiceOpt {
```
RonFed, line 162: Is there a way to avoid this duplication with the functions in `instrumentation.go` ?

internal/pkg/orchestrator/orchestrator.go
```diff
@@ -0,0 +1,221 @@
...
+			}()
+		}
+	}
+}
+
+func (s *Service) findProcess() {
+	if s.pid != 0 {
+		s.processch <- &pidServiceName{
+			pid:         s.pid,
+			serviceName: s.serviceName,
```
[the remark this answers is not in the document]
dineshg13: Yes. That is intent.

internal/pkg/orchestrator/service.go
dineshg13: I would like remove `instrumentation.go` . I will do so in future PRs. After orchestrator implementation, there is no need for `instrumentation.go` . I didn't want to introduce breaking changes in this PR.

internal/pkg/orchestrator/service.go
dineshg13: removed it for now. We can add in future, if we would like to have a version of auto version without ability to monitor all services.

internal/pkg/orchestrator/service.go
dineshg13: Yes. It is the current behaviour as well.

internal/pkg/orchestrator/orchestrator.go
dineshg13: Calling `Stop`  when context is done.

instrumentation.go
```diff
@@ -65,64 +63,62 @@ func NewInstrumentation(opts ...InstrumentationOption) (*Instrumentation, error)
...
+		orchestrator: r,
+		ctx:          ctx,
 	}, nil
 }
 
 // Run starts the instrumentation.
-func (i *Instrumentation) Run(ctx context.Context) error {
-	return i.manager.Run(ctx, i.target)
+func (i *Instrumentation) Run() error {
+	return i.orchestrator.Run(i.ctx)
```
MrAlias, line 100:
```suggestion
func (i *Instrumentation) Run(ctx context.Context) error {
	return i.orchestrator.Run(ctx)
```

instrumentation.go
MrAlias: The context added to `Run` is still this context. It should not be. The creator of an `Instrumentor` needs to be the one creating and passing this context.

instrumentation.go
dineshg13: We want to use same exporter for monitoring all the processes. If we do inside the controller, we will be creating on grpc connection for each process.

instrumentation.go
dineshg13: Added it as arg.

dineshg13:
There are too many breaking changes. It has become harder to update the PR. Closing the effort.

PR #211, opened 2023-07-05
BPF FS: add support for multiple processes

edeNFed:
Needed as part of #197 
This PR adds support for multiple processes in the bpf filesystem.
Storing bpf data structures in `/sys/fs/bpf` will cause a conflict and race conditions when instrumenting multiple processes.
Instead, this change will store BPF data structures in `/sys/fs/bpf/<processID>`


pkg/instrumentors/bpffs/bpfsfs.go
```diff
@@ -14,5 +14,54 @@
...
+	if err != nil {
+		log.Logger.Error(err, "failed to statfs bpf filesystem")
+		return false
+	}
+
+	return stat.Type == unix.BPF_FS_MAGIC
+}
+
+// Cleanup removes the BPF file-system for the given target.
+func Cleanup(target *process.TargetDetails) error {
```
dineshg13, line 65: Do we need to call this explicit ?

edeNFed: Yes, when detecting that a process is dead and unloading the eBPF programs, this should also be invoked to cleanup the relevant data structures in BPFFS

dineshg13 (approved):

pkg/instrumentors/bpffs/bpfsfs.go
```diff
@@ -14,5 +14,54 @@
...
+			return err
+		}
+
+		err := unix.Mount(bpfFsPath, bpfFsPath, "bpf", 0, "")
+		if err != nil {
+			return err
+		}
+	}
+
+	return os.Mkdir(PathForTargetApplication(target), 0755)
```
MikeGoldsmith, line 50:
```suggestion
	// create directory with read, write and execute permissions
	return os.Mkdir(PathForTargetApplication(target), 0755)
```

MikeGoldsmith (approved):
One small suggestion to add a comment to describe file permissions when creating directory, otherwise looks good to me 👍🏻

Dockerfile
```diff
@@ -1,12 +1,12 @@
-FROM debian:11 as builder
```
MrAlias, line 1: Why is this switching the image?

Issue #220, opened 2023-07-11
net/http context propagation appears flaky


damemi:
The [test app](https://github.com/open-telemetry/opentelemetry-go-instrumentation/tree/main/test/e2e) fixtures are showing a weird flakiness where [`parentSpanId`](https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/7fcfd9126694c339773731b73164f9f40822bad7/test/e2e/gorillamux/traces.json#L78) is sometimes set (redacted to `xxxxx`), and sometimes empty.

First noticed in https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/213#issuecomment-1623937567. In this case, the PR did *not* have `parentSpanId` set, but the Github workflow output *did*. 

After this, in that same PR, a [commit](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/213/commits/8d1d928caff4aa4da6c5d077e957940bd2a6d2b6) was pushed to add `parentSpanId`. The Github workflow then failed *again*, [trying to remove the `parentSpanId` value](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/213#issuecomment-1624159751).

The `parentSpanId` field was first set in https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/91 for [`gorillamux`](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/91/files#diff-b6e87a1406e21be98ec01fad4a91d41a329b1e2cae8c090854e3475e8e480f65R78) and [`nethttp`](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/91/files#diff-67b5343bc1e04e7ef0b7f863cfd45699043441c3934616b5bd541ff85f5425e7R49) fixtures (https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/91/commits/56cbc7d9e41df12cfadc7f418267fbf5e0cc8b18).

Possible causes:

* Github workflow timing -- the sample apps may be taking longer than expected to complete a full request, not allowing full context propagation.
* Context propagation -- the affected spans seem to be the server (receiving) span for net/http (and libraries that use net/http, like gorilla).

It could also be a combination of both, or something else entirely.

[...]

PR #228, opened 2023-07-20
Fix net/http client instrumentation

oxeye-gal:
Hi,

This PR fixes the flakiness issue described in #220 as it aligns the net/http client instrumentation so it will not update the `spans_in_progress` map similarly to how the gRPC client instrumentation is implemented.

It is worth mentioning that it will not fix the underlying issue described in my comment on #220 which means sometimes injections to header map might still occur twice due to the eBPF program being ran twice.


MikeGoldsmith:
Thanks for sharing @oxeye-gal - sorry for the delay in getting back to you.

I think this is now superseded by #118 where we utilise `start_tracking_span` and `stop_tracking_span` utility funcs to manages the active spans and associated contexts.

PR #240, opened 2023-07-27
Database/sql instrumentation

RonFed:
This PR adds instrumentation for `database/sql` package.
Added additional example under `examples/httpPlusDb`
As a first step, this instrumentation only includes the query string.
Additional fields from the opentelemetry spec will be added in future PRs.
http_Db_traces


pkg/instrumentors/bpf/database/sql/probe.go
```diff
@@ -0,0 +1,215 @@
...
+	"github.com/cilium/ebpf"
+	"github.com/cilium/ebpf/link"
+	"github.com/cilium/ebpf/perf"
+	"golang.org/x/sys/unix"
+
+	"go.opentelemetry.io/auto/pkg/instrumentors/context"
+	"go.opentelemetry.io/auto/pkg/instrumentors/events"
+	"go.opentelemetry.io/auto/pkg/log"
+	"go.opentelemetry.io/otel/attribute"
+	semconv "go.opentelemetry.io/otel/semconv/v1.7.0"
```
pellared, line 34:
```suggestion
	semconv "go.opentelemetry.io/otel/semconv/v1.18.0"
```

pkg/instrumentors/bpf/database/sql/probe.go
```diff
@@ -0,0 +1,215 @@
...
+
+	return &events.Event{
+		Library:     h.LibraryName(),
+		Name:        "DB",
+		Kind:        trace.SpanKindClient,
+		StartTime:   int64(e.StartTime),
+		EndTime:     int64(e.EndTime),
+		SpanContext: &sc,
+		Attributes: []attribute.KeyValue{
+			semconv.DBStatementKey.String(query),
```
pellared, line 191: Collection of `db.statement` should be opt-in as it may contain sensitive data.
See https://opentelemetry.io/docs/specs/semconv/database/database-spans/#call-level-attributes

pellared (commented):
E2E tests are missing.

_I made only a quick look._

pkg/instrumentors/bpf/database/sql/probe.go
RonFed: Added an environment variable to opt-in for query collection

edeNFed (approved):
@RonFed looks great!

pkg/instrumentors/bpf/database/sql/probe.go
```diff
@@ -0,0 +1,240 @@
...
+// Instrumentor is the database/sql instrumentor.
+type Instrumentor struct {
+	bpfObjects   *bpfObjects
+	uprobes      []link.Link
+	returnProbs  []link.Link
+	eventsReader *perf.Reader
+}
+
+// ExePathEnvVar is the environment variable key whose value points to the
+// instrumented executable.
```
pellared, line 62: The comment is bad.

pkg/instrumentors/bpf/database/sql/probe.go
```diff
@@ -0,0 +1,240 @@
...
+const IncludeSQLQUeryEnvVar = "OTEL_GO_AUTO_INCLUDE_SQL_QUERY"
```
pellared, line 63: How about.
```suggestion
const IncludeDBStatementEnvVar = "OTEL_GO_AUTO_INCLUDE_DB_STATEMENT"
```

pkg/instrumentors/bpf/database/sql/probe.go
```diff
@@ -0,0 +1,239 @@
...
+
+	if h.bpfObjects != nil {
+		h.bpfObjects.Close()
+	}
+}
+
+// shouldIncludeDBStatement returns if the user has configured SQL queries to be included.
+func shouldIncludeDBStatement() bool {
+	val, exists := os.LookupEnv(IncludeDBStatementEnvVar)
+	if exists {
```
pellared, line 231:
```suggestion
	val := os.Getenv(IncludeDBStatementEnvVar)
	if val != "" {
```

Reason:

> The SDK MUST interpret an empty value of an environment variable the same way as when the variable is unset.

From https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md#parsing-empty-value

examples/httpPlusdb/test.db
pellared, line 1: I am not a fan of adding binaries to `git`.  I would prefer to create the DB in code. It also makes the example more readable.

test/e2e/databasesql/test.db
pellared, line 1: I am not a fan of adding binaries to `git`.  I would prefer to create the DB in code. Also this would enable us to test not only SQL DQL, but also SQL DDL and DML.

RonFed:
@pellared  I resolved your comments, please let me know if there is anything else to fix

pkg/instrumentors/bpf/database/sql/bpf/probe.bpf.c
```diff
@@ -0,0 +1,90 @@
...
+
+// Injected in init
+volatile const bool should_include_db_statement;
+
+// This instrumentation attaches uprobe to the following function:
+// func (db *DB) queryDC(ctx, txctx context.Context, dc *driverConn, releaseConn func(error), query string, args []any)
+SEC("uprobe/queryDC")
+int uprobe_queryDC(struct pt_regs *ctx) {
+    // argument positions
+    u64 context_ptr_pos = 3;
```
MikeGoldsmith, line 38: I think these should be checked by the offset process and injected like the HTTP param offsets. Offsets can change between version.

pkg/instrumentors/bpf/database/sql/bpf/probe.bpf.c
```diff
@@ -0,0 +1,90 @@
...
+
+    if (should_include_db_statement) {
+        // Read Query string
+        void *query_str_ptr = get_argument(ctx, query_str_ptr_pos);
+        u64 query_str_len = (u64)get_argument(ctx, query_str_len_pos);
+        u64 query_size = MAX_QUERY_SIZE < query_str_len ? MAX_QUERY_SIZE : query_str_len;
+        bpf_probe_read(sql_request.query, query_size, query_str_ptr);
+    }
+
+    // Get goroutine as the key fro the SQL request context
```
MikeGoldsmith, line 53:
```suggestion
    // Get goroutine as the key for the SQL request context
```

pkg/instrumentors/bpf/database/sql/bpf/probe.bpf.c
```diff
@@ -0,0 +1,90 @@
...
+    if (span_ctx != NULL) {
+        // Set the parent context
+        bpf_probe_read(&sql_request.psc, sizeof(sql_request.psc), span_ctx);
+        copy_byte_arrays(sql_request.psc.TraceID, sql_request.sc.TraceID, TRACE_ID_SIZE);
+        generate_random_bytes(sql_request.sc.SpanID, SPAN_ID_SIZE);
+    } else {
+        sql_request.sc = generate_span_context();
+    }
+
+    bpf_map_update_elem(&context_to_sql_events, &goroutine, &sql_request, 0);
```
MikeGoldsmith, line 67: These should be replaced with `start_tracking_span` added in #118.

pkg/instrumentors/bpf/database/sql/bpf/probe.bpf.c
```diff
@@ -0,0 +1,90 @@
...
+    void *goroutine = get_goroutine_address(ctx, context_ptr_pos);
+    void *sqlReq_ptr = bpf_map_lookup_elem(&context_to_sql_events, &goroutine);
+
+    struct sql_request_t sqlReq = {0};
+    bpf_probe_read(&sqlReq, sizeof(sqlReq), sqlReq_ptr);
+    sqlReq.end_time = bpf_ktime_get_ns();
+
+    bpf_perf_event_output(ctx, &events, BPF_F_CURRENT_CPU, &sqlReq, sizeof(sqlReq));
+
+    bpf_map_delete_elem(&context_to_sql_events, &goroutine);
```
MikeGoldsmith, line 87: Should be replaced with `stop_tracking_span` added in #118.

MikeGoldsmith (changes requested):
This looks really good - thanks @RonFed 👍🏻 

I've left some suggestions around parameter offsets. Also, it looks like the example app is all boiler plate - is there an official docker image published that could be used instead?

pkg/instrumentors/bpf/database/sql/bpf/probe.bpf.c
RonFed: Comment removed after using the updated context propagation

pkg/instrumentors/bpf/database/sql/bpf/probe.bpf.c
RonFed: This context argument is part of the instrumented function signature - which I think we assume is constant. The cases where we use the offset tracker are for struct fields. In the http case the context is part of the Request struct and not passed "directly" as an argument

RonFed:
@MikeGoldsmith  Can you please have another look ?

MikeGoldsmith (approved):
Looks good, thanks @RonFed

PR #249, opened 2023-08-07
Consolidate HTTP server instrumentations to use single bpf probe implementation

MikeGoldsmith:
Currently we have separate bpf probes for each http server instrumentation (net/http, gorilla mux & gin-gonic). The bpf code is identical across each implementation with only the go usage attaching the probe to different user space functions.

This PR does the following:
- Create new root bpf directory and move all bpf headers into it
- Move the net/http server bpf probe code to new bpf directory
- Update generate Makefile target to also set the BPF_PROBES directory which instrumentations can referece
- Update gorilla mux and gin-gonic instrumentations to use generic http service bpf probe, including:
  - Update event struct to add ParentSpanContext field
  - Add missing offset fields
  - Setting the parent span context when converting events
  - Updating expected test json structure to validate a parent span context is being set correctly
- Remove gorilla mux and gin-gonic custom bpf probes


MikeGoldsmith:
Gin e2e test is still flakey - it works locally. Could someone else try and run to see if they something different?

MrAlias:

MikeGoldsmith:
Closing as there's been lot of changes in the bpf code and project structurally which has meant this change isn't as needed as before.

PR #257, opened 2023-08-18
Use common macro for Uprobe return flow

RonFed:
* Use UPROBE_RETURN to declare the common uprobe return logic (finding the corresponding context, setting up end time, and sending the event via perf buffer)
* BASE_SPAN_PROPERTIES as common fields for all instrumentations
* Header guards


edeNFed (approved):
Great idea @RonFed

include/uprobe.h
MrAlias, line 1: Should this have a header guard as well?

MrAlias (approved):

include/uprobe.h
RonFed: Right. I added header guards for this and the rest of our headers

PR #259, opened 2023-08-21
Update remote mmap, avoid page faults using madvise syscall. Remove l…

RonFed:
WIP
* Update remote mmap, avoid page faults using madvise syscall
* Remove large local variable in uprobe_HttpClient_Do to avoid reaching the 512 bytes limit of eBPF

### DIFF
--- a/cli/main.go
+++ b/cli/main.go
@@ -76,6 +76,14 @@ func main() {
 		log.Logger.Error(err, "error while analyzing target process")
 		return
 	}
+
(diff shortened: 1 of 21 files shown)


edeNFed:
Adding some more context on this:
We observed at some Keyval customers that when the machine is lacking resources (especially memory), Linux would free the memory area used for context propagation which will cause `bpf_write_user` to fail. This should fix this scenario

MrAlias:

PR #264, opened 2023-08-22
Bump support for Go 1.21

MrAlias:


edeNFed (approved):

.github/workflows/checks.yml
```diff
@@ -7,7 +7,7 @@ on:
   workflow_dispatch:
 
 env:
-  GO_VERSION: "1.20"
+  GO_VERSION: "~1.21.1"
```
pellared, line 10:
```suggestion
  GO_VERSION: "1.21"
```

.github/workflows/checks.yml
```diff
@@ -28,7 +28,7 @@ jobs:
       - name: Install Go
         uses: actions/setup-go@v4.0.0
         with:
-          go-version: ${{ env.DEFAULT_GO_VERSION }}
+          go-version: ${{ env.GO_VERSION }}
```
pellared, line 31:
```suggestion
          go-version: ${{ env.GO_VERSION }}
          check-latest: true
```

.github/workflows/checks.yml
```diff
@@ -52,6 +52,31 @@ jobs:
         run: make license-header-check go-mod-tidy golangci-lint
       - name: Check clean repository
         run: make check-clean-work-tree
+  compatibility-test:
+    strategy:
+      matrix:
+        go-version: ["~1.21.1", "~1.20.8"]
```
pellared, line 58:
```suggestion
        go-version: ["1.21", "1.20"]
```

.github/workflows/build.yaml
```diff
@@ -14,7 +14,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v4
         with:
-          go-version: "1.20"
+          go-version: "~1.21.1"
```
pellared, line 17:
```suggestion
          go-version: "1.21"
          check-latest: true
```

.github/workflows/checks.yml
```diff
@@ -52,6 +52,31 @@ jobs:
...
+        # https://docs.github.com/en/actions/hosting-your-own-runners/using-self-hosted-runners-in-a-workflow
+        arch: [amd64]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v3
+      - name: Setup Go
+        uses: actions/setup-go@v4
+        with:
+          go-version: ${{ matrix.go-version }}
```
pellared, line 72:
```suggestion
          go-version: ${{ matrix.go-version }}
          check-latest: true
```

.github/workflows/checks.yml
```diff
@@ -60,7 +85,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v4.0.0
         with:
-          go-version: ${{ env.DEFAULT_GO_VERSION }}
+          go-version: ${{ env.GO_VERSION }}
```
pellared, line 88:
```suggestion
          go-version: ${{ env.GO_VERSION }}
          check-latest: true
```

.github/workflows/kind.yml
```diff
@@ -21,7 +21,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v4
         with:
-          go-version: "1.20"
+          go-version: "~1.21.1"
```
pellared, line 24:
```suggestion
          go-version: "1.21"
          check-latest: true
```

pellared (commented):
Proposing changes so that we do not update each patch release of Go 😉

.github/workflows/checks.yml
MrAlias: Why not set `1.21.1` as the minimum here, `1.21.0` includes security issues resolved by `1.21.1`?

pellared: `1.21` is not `1.21.0`. If you set `check-latest: true` it should grab the latest `1.21.x` version.
See https://github.com/actions/setup-go#check-latest-version

MrAlias: [It should "grab the latest `1.21.x` version" with `"~1.21.1"` as well](https://github.com/npm/node-semver): 



However, this would ensure `1.21.0` is never used.

pellared (approved):

.github/workflows/checks.yml
MrAlias: Yeah, I can see adding `check-latest: true`. Just wondering if this needed.

bryan-aguilar: > However, this would ensure 1.21.0 is never used.

Agreed, and this was my intent through the waves of PRs I submitted yesterday. I think the combination of both `go-version` and `check-latest` is the safest way to ensure that a minimum versions is used.

pellared (approved):

PR #266, opened 2023-08-23
Enhance the inject_header function, fix bug in net/http server instru…

RonFed:
* Fix bug in net/http server instrumentation in which the span context was always generated as a random string rather than taking the parent one if exists
* Enhance the inject header function to be more readable


edeNFed (approved):

RonFed:
@oxeye-gal Can you please have a look on the changes made to the inject_headers function ?

MikeGoldsmith (approved):

MrAlias:

RonFed:
> Please rebase these changes into `internal/pkg` (changed in #282). The `pkg` directory is going to be removed in #302 given that package is no longer used.

Rebased

internal/include/go_types.h
```diff
@@ -54,6 +54,17 @@ struct map_bucket {
     void *overflow;
 };
 
+// A header for a Go map. Defined in go/src/runtime/map.go
+struct go_map_header {
```
MrAlias, line 58: This seems brittle. What happens when the Go `hmap` field order or structure changes? Is there a better way to pass this from the Go memory space?

RonFed: It looks like the last time a field was added was 6 years ago - so I think it is quite stable. Theoretically, we could do something similar to the offset tracker with this but I don't think it is worth the effort considering that it didn't change much

MrAlias: What can be done with the offset tracker? I'd be more in favor of exploring less brittle options then assuming an unexported type won't change.

internal/include/go_types.h
```diff
@@ -54,6 +54,17 @@ struct map_bucket {
...
+    u64 map_keyvalue_count;
```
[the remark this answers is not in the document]
MrAlias: What's the downside off using `s64` here instead?

internal/include/go_types.h
RonFed: I guess we could analyze the binary by looking for the map header struct and extracting the offsets of the fields that we care about.  In that case, we won't be able to read the entire header in one read operation as I did in this PR but would need to read one field at a time based on its offset.
As documented in go/src/runtime/map.go the count field that we read must be first, so I think we could be fine with just getting the offset of the `buckets` field which is not that bad. 
Again, I'm not sure if it is worth it. @edeNFed what do you think ?

internal/include/go_types.h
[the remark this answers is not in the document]
RonFed: Done

internal/include/go_types.h
MrAlias: > so I think we could be fine with just getting the offset of the `buckets` field which is not that bad.

+1

I think generating these offsets should be similar to the other `runtime` offsets we generate in that they shouldn't be too hard to figure out (once we fix the offset generator).

RonFed: @MrAlias Ok. Can I update the offsets json in this PR? or do you prefer for the offset-tracker to be fixed first?

RonFed: I added the buckets field to the offset tracker and used the injected value to read the pointer

MrAlias (approved):

Issue #270, opened 2023-08-24
Move all exported code that does not need to be exported into `internal`


MrAlias:
Currently this project exports many packages that an end-user is not expected to interact with. This public API surface area will be locked into long-term stability requirements when this project becomes stable. The existing exported packages should be deprecated and copied into an equivalent package in `internal`.

Packages to move:

- [x] `include/*`: https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/282
- [x] `offsets-tracker/*`: https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/281
- [x] `pkg/*`: https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/282
- [x] `test/*`: https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/282

Replacements:

- [ ] High-level instrumentation API: https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/284

MrAlias:
_Originally posted by @edeNFed in https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/282#issuecomment-1708804490_

> @MrAlias can we leave a few APIs in the pkg directory? We are using the Go auto instrumentation as a library and moving everything to internal will break our code: https://github.com/keyval-dev/odigos/blob/main/odiglet/pkg/ebpf/objects.go

A replacement API needs to be built to support this use case.

MrAlias:
Going to remove the new API as a tracked issue of this issue as the work completed is good enough for a release. Going to have #284 track the new API task in the next milestone.

PR #281, opened 2023-09-06
Replace offsets-tracker with a copy in internal/tools

MrAlias:
Part of #270; this moves code that does not need to be exported into the `internal` package.

- Deprecate offsets-tracker
- Copy existing offsets-tracker into a new `internal/tools/offsets` package
- Update tooling to use new package


.github/workflows/build.yaml
```diff
@@ -14,7 +17,8 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v4
         with:
-          go-version: "1.20"
+          go-version: ${{ env.GO_VERSION }}
```
pellared, line 20: This should be updated if https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/264 is merged before this PR.

pellared (approved):

PR #282, opened 2023-09-06
Copy `pkg` and `test` to `internal` and deprecate original

MrAlias:
Part of #270; this moves code that does not need to be exported into the `internal` package.

- Copy existing `pkg` and `test` packages into `internal` package
- Deprecate existing `pkg` and `test` packages and modules.
- Update tooling to use new package


MrAlias:
> @MrAlias can we leave a few APIs in the pkg directory? We are using the Go auto instrumentation as a library and moving everything to internal will break our code: https://github.com/keyval-dev/odigos/blob/main/odiglet/pkg/ebpf/objects.go

Sure, I'll take a look and update.

MrAlias:
@edeNFed it seems like you are already working with a fork of this project: https://github.com/keyval-dev/odigos/blob/acf1ec5daa52fb56a551ee47ef22bb0515e1ddad/odiglet/go.mod#L98. I'm not able to find things like `NewControllerWithServiceName` in the existing packages.

This makes me think that we need to work on our API so these kinds of workarounds are not needed. I have thoughts on a revised and simpiler API, but don't want to make that proposal in this PR. What do you think about keeping this PR as is, tracking the revised API design in a new issue, and adding the API to the top-level `go.opentelemetry.io/auto` package?

edeNFed:
Sure sounds great

edeNFed (approved):

pellared (approved):

PR #284, opened 2023-09-06
proposal: revised instrumentation API

MrAlias:
Part of #270.

Much of the existing API is being un-exported, but [there is still a need to expose programmatic setup of instrumentation](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/282#issuecomment-1708804490). This is a proposal replacement for the current instrumentation APIs.

Design:

- Top-level
- A single `Instrumentation` type manages the instrumentation setup and running
  - The type is a `struct` which can have method and fields added in the future in a backwards compatible manner.
- The `NewInstrumentation` creation function is extensible by using the [option pattern](https://github.com/open-telemetry/opentelemetry-go/blob/main/CONTRIBUTING.md#option)


MrAlias:
@edeNFed: I think this should work as a replacement for https://github.com/keyval-dev/odigos/blob/main/odiglet/pkg/ebpf/objects.go. The only thing missing is passing the service name as an argument, but that can be added as an option with the proposed design.

internal/pkg/instrumentors/runner.go
```diff
@@ -15,19 +15,20 @@
...
 import (
+	"context"
 	"fmt"
 
 	"github.com/cilium/ebpf/link"
 	"github.com/cilium/ebpf/rlimit"
 
 	"go.opentelemetry.io/auto/internal/pkg/inject"
-	"go.opentelemetry.io/auto/internal/pkg/instrumentors/context"
+	iCtx "go.opentelemetry.io/auto/internal/pkg/instrumentors/context"
```
pellared, line 25:
```suggestion
	ictx "go.opentelemetry.io/auto/internal/pkg/instrumentors/context"
```

fix casing

MrAlias: The casing is intentional, it is short for two words "instrumentors context"

MrAlias: I'm okay with leaving it as it given the Go language supports this and it has semantic significance.

pellared (approved):

pdelewski (approved):

PR #304, opened 2023-09-13
Fix module structure

MrAlias:
- Remove `internal/test` modules
  - These modules have not been released yet
  - They do not need to be distinct from the top-level module given this project is not meant to be a library and the number of dependencies is not a concern
- Unify all examples into a single module
  - The `httpPlusdb` example has not been released yet, remove the `go.mod` in that directory
  - Deprecate `rolldice` to have it moved to the new examples module
- Fix the versions file
  - Fix the soon-to-be-removed test/e2e/databasesql module name and ignore it in versions.yaml
  - Add the examples module


MrAlias:
Blocked by https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/305

dineshg13 (approved):
LGTM

pellared (approved):
👍

PR #337, opened 2023-09-19
Do not publish examples module

### LINKED ISSUE #336: Do not publish examples
These modules:

https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/a4ce6b9837ee98194f40d06d7860eb0cfef80a08/versions.yaml#L20-L21

could be moved under `excluded-modules:` as I see no reason to publish them (no reusable code or "test examples" that would be available in pkg.go.dev.

I think that `go.opentelemetry.io/auto` is (and probably would be) the only module that we need to publish.

MrAlias:
Resolve #336

### DIFF
--- a/versions.yaml
+++ b/versions.yaml
@@ -17,8 +17,8 @@ module-sets:
     version: v0.3.0-alpha
     modules:
       - go.opentelemetry.io/auto
-      - go.opentelemetry.io/auto/examples
       - go.opentelemetry.io/auto/examples/rolldice
 excluded-modules:
   - github.com/hashicorp/go-version
+  - go.opentelemetry.io/auto/examples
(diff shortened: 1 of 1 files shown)


pellared:
Do we need a separate issue for moving `go.opentelemetry.io/auto/examples/rolldice` after the next release?

MrAlias:
> Do we need a separate issue for moving `go.opentelemetry.io/auto/examples/rolldice` under `excluded:` after the next release?

The plan is to delete `go.mod` in `go.opentelemetry.io/auto/examples/rolldice` after the next release.

pdelewski (approved):

pellared (approved):

PR #338, opened 2023-09-21
enable CGO for local builds

rogercoll:
Issue found while trying to compile the binary in a fresh Ubuntu 22.04 box. The main point of enabling CGO (required) during the compilation is to provide a more clear error to the user. 

In environments where CGO is disabled by default and gcc is not installed, the Go compiler cannot find imported "C" functions:

```
GOOS=linux go build -o otel-go-instrumentation cli/main.go
internal/pkg/process/ptrace/ptrace_linux.go:49:9: undefined: waitpid
```

By explicitly setting the CGO_ENABLED=1 Go environment variable, the compilation error becomes more helpful:
```
CGO_ENABLED='1' GOOS=linux go build -o otel-go-instrumentation cli/main.go
cgo: C compiler "gcc" not found: exec: "gcc": executable file not found in $PATH
```


rogercoll:
Skip changelog?

MrAlias (approved):

PR #342, opened 2023-09-21
Allow multiple offsets version ranges

### LINKED ISSUE #309: Allow multiple version ranges for traced field offsets
Currently, our offset data model expects a continuous range of supported field offsets and therefore only allows one version range per:

https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/82c7468a72ac7c6f09aef23a4abf595f128f0ffb/internal/pkg/inject/data.go#L24-L25

As has been shown in https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/273 where there is a break in the support of gRPC (https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/949e8a39f95bb9f9fbed4b78eec75e55a9a85468/internal/pkg/inject/offset_results.json#L5-L28), it is not always the case that we will be able to support continuous ranges of package versions.

We should instead define the `TracedStruct` as:

```go
type TrackedStruct map[string][]TrackedField
```

And track each version range.

Related to #271 
Blocked by https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/307 (once merged we can fix this in a single spot)

MrAlias:
Fixes #309

### DIFF
--- a/internal/pkg/inject/data.go
+++ b/internal/pkg/inject/data.go
@@ -57,7 +57,7 @@ func (o *TrackedOffsets) GetOffset(strct, field, ver string) (uint64, bool) {
 		}
 
 		if v.LessThan(oldest) || v.GreaterThan(newest) {
-			return 0, false
+			continue
 		}
 
 		// Search from the newest version (last in the slice).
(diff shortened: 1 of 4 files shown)


pellared (commented):
Can you add new test cases in `data_test.go` that verifies the new functionality?

pellared (approved):

PR #346, opened 2023-09-24
Add WithServiceName and WithPid as config options for instrumentation…

RonFed:
This PR add two InstrumentationOption:
* `WithServiceName` similar to `WithTarget` (with the same behavior for which environment variable is possible and will take precedence).
* `WithPID` which calls `WithTarget`.
These will be useful for https://github.com/keyval-dev/odigos/blob/main/odiglet/pkg/ebpf/director.go

In addition the `Cleanup` in bpffs was not called which resulted in garbage folders under /sys/fs/bpf not get deleted after the instrumentation is finished

### DIFF
--- a/instrumentation.go
+++ b/instrumentation.go
@@ -25,9 +25,13 @@ import (
 	"go.opentelemetry.io/auto/internal/pkg/process"
 )
 
-// envTargetExeKey is the key for the environment variable value pointing to the
-// target binary to instrument.
-const envTargetExeKey = "OTEL_GO_AUTO_TARGET_EXE"
+const (
+	// envTargetExeKey is the key for the environment variable value pointing to the
+	// target binary to instrument.
+	envTargetExeKey = "OTEL_GO_AUTO_TARGET_EXE"
+	// envServiceName is the key for the envoriment variable value containing the service name.
+	envServiceNameKey = "OTEL_SERVICE_NAME"
+)
 
 // Instrumentation manages and controls all OpenTelemetry Go
 // auto-instrumentation.
@@ -41,6 +45,8 @@ var (
 	// Error message returned when instrumentation is launched without a target
 	// binary.
 	errUndefinedTarget = fmt.Errorf("undefined target Go binary, consider setting the %s environment variable pointing to the target binary to instrument", envTargetExeKey)
+	// Error message returned when no service name is specified.
+	errUndefinedServiceName = fmt.Errorf("undefined service name, consider setting %s", envServiceNameKey)
 )
 
 // NewInstrumentation returns a new [Instrumentation] configured with the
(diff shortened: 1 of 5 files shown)


internal/pkg/instrumentors/allocator/allocator_linux.go
```diff
@@ -45,3 +46,7 @@ func (a *Allocator) Load(ctx *context.InstrumentorContext) error {
 
 	return nil
 }
+
+func (a *Allocator) Clean(target *process.TargetDetails) error {
```
MrAlias, line 50: Missing comment.

instrumentation.go
```diff
@@ -150,3 +163,32 @@ func WithTarget(path string) InstrumentationOption {
...
 }
+
+// WithServiceName returns an [InstrumentationOption] defining the service name
+//
+// If multiple of these options are provided to an [Instrumentation], the last
+// one will be used.
+//
+// If OTEL_SERVICE_NAME is defined it will take precedence over any value
+// passed here.
+func WithServiceName(serviceName string) InstrumentationOption {
```
MrAlias, line 174: This needs a changelog entry in an `Added` section.

instrumentation.go
```diff
@@ -150,3 +163,32 @@ func WithTarget(path string) InstrumentationOption {
...
+		c.serviceName = serviceName
+		return c
+	})
+}
+
+// WithPID returns an [InstrumentationOption] corresponding to the executable
+// used by the provided pid.
+//
+// If WithTarget is used, the one which is used last will take precedence.
+func WithPID(pid int) InstrumentationOption {
```
MrAlias, line 185: This needs a changelog entry in an `Added` section.

MrAlias (commented):
Can you split this PR into 4 PRs:

- Add the service name option
- Add the PID option
- Fix the close
- Fix the use of `ioutil`

Currently it is trying solve multiple issues in one PR.

instrumentation.go
```diff
@@ -122,13 +129,19 @@ func (c instConfig) applyEnv() instConfig {
...
+	}
 	return c
 }
 
 func (c instConfig) validate() error {
 	if c.target == nil {
 		return errUndefinedTarget
 	}
+	if c.serviceName == "" {
+		return errUndefinedServiceName
```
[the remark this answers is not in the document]
RonFed: Currently, we are not allowing for not setting the service name env variable. So this code preserves that.
We can change it to a default service name if not set.

pellared:
Missing tests 😉

RonFed:
> The tests are missing.

@pellared Where do you think would be a good place to add tests for these options? I couldn't find tests for the existing one

pellared:
> > The tests are missing.
> 
> @pellared Where do you think would be a good place to add tests for these options? I couldn't find tests for the existing one

I am open to suggestions. Personally, I think both options deserve integration/e2e tests. 

For service name handling we could also add unit tests for `newInstConfig` function.

RonFed:
WithServiceName https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/353
WithPID: https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/355
bpffs cleanup https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/347
stop using ioutil https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/351

PR #363, opened 2023-10-02
Call runtime.LockOSThread() before ptrace

RonFed:
From my testing, ptrace-ing the instrumented process failed every once in a while with an error "no such process"
https://github.com/golang/go/issues/43685 seems to describe the issue.
This PR should make sure the OS thread of our code is fixed durring the ptrace attachment and mmap call.


edeNFed (approved):

edeNFed:
lgtm

PR #368, opened 2023-10-10
Improve eBPF context propagation stability

RonFed:
The main goal of this PR is to improve the context propagation in the eBPF code by making it more stable across kernel versions and across instrumented libraries. Those fixes are the result of investigation of Keyval's customers running go auto instrumentation in production.
The changes can be divided into a few categories:
1. Memory allocation: we use `bpd_probe_write_user` helper function in order to modify user memory.  The main challenge is to write to memory which we "allocate" since page faults won't get handled if they occur while the eBPF code attempts to write to such addresses.  In order to reduce the page faults I added the following changes:
* madvise and mlock syscalls after we perform the mmap.
* allocate 8 pages per CPU. This number is a result of some experiments I did but the tradeoff as I understand it is: on one side we want to have enough memory available for handling a large number of instrumneted functions without overriding exiting trace memory. On the other side, allocating too much memory increases the chance of page faults - which will cause the write operations to fail, and damage the context propagation.
* Since the added code, I added `allocate.go`
2. Improving the resiliancy of context propagation relevant functions in the eBPF code. This mainly include error handling and a unified API for all the uprobes.
3. Extracting the Go context key via a new util function.
4. Adding utils function in the eBPF code for reading user memory strings.
5. Change the http server uprobe name to better fit the instrumented function


internal/include/go_context.h
```diff
@@ -72,20 +79,82 @@ static __always_inline struct span_context *get_parent_span_context(void *ctx) {
...
+    if (err != 0)
+    {
+        bpf_printk("Failed to update tracked_spans_by_sc map: %ld", err);
+        return;
+    }
 }
 
-static __always_inline void stop_tracking_span(struct span_context *sc) {
+static __always_inline void stop_tracking_span(struct span_context *sc, struct span_context *psc) {
+    if (psc == NULL || sc == NULL) {
```
edeNFed, line 100: Isn't NULL psc a valid state? If we are now creating the root span psc will be NULL

edeNFed:
Great work! The code looks much better now. I left a small question regarding `stop_tracking_span`

internal/include/go_context.h
RonFed: Currently this function is passed with pointers to the stack. A root span will have a valid pointer but psc will be zero, or not existing in the map. I changed the logic here to fit other scenarios as well.

edeNFed (approved):

PR #371, opened 2023-10-10
Update PID option precedence

MrAlias:
Match the WithTarget option precedence in relation to the environment variable to provide users with consistent behavior.

Includes clarifications to the options documentation.


pellared (approved):
👍

PR #388, opened 2023-10-16
Remove SYS_PTRACE

pellared:
Testing if SYS_PTRACE capability is needed as it seems to be missing in https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/main/examples/rolldice/docker-compose.yaml.


pellared:
@edeNFed @RonFed Do you know from where the `SYS_PTRACE` requirement comes from?

CHANGELOG.md
```diff
@@ -17,9 +21,9 @@ OpenTelemetry Go Automatic Instrumentation adheres to [Semantic Versioning](http
 
 ### Changed
 
-- Fix bug in the `net/http` server instrumentation which always created a new span context. ([#266]https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/266)
+- Fix bug in the `net/http` server instrumentation which always created a new span context. ([#266](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/266))
 - Fix runtime panic if OTEL_GO_AUTO_TARGET_EXE is not set. ([#339](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/339))
-- Improve eBPF context propagation stability ([#368]https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/368)
+- Improve eBPF context propagation stability ([#368](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/368))
```
pellared, line 26: I also fixed the hyperlinks in https://github.com/open-telemetry/opentelemetry-go-instrumentation/releases/tag/v0.7.0-alpha 😉

RonFed:
> @edeNFed @RonFed Do you know from where the `SYS_PTRACE` requirement comes from?

It is there since we are using ptrace to allocate memory for the traced process.
Assuming we are running as root (which we have to assume I think), this is probably not needed

pellared:
> > @edeNFed @RonFed Do you know from where the `SYS_PTRACE` requirement comes from?
> 
> It is there since we are using ptrace to allocate memory for the traced process. Assuming we are running as root (which we have to assume I think), this is probably not needed

You are right.

From https://docs.docker.com/engine/reference/run/#runtime-privilege-and-linux-capabilities

The --privileged flag gives all capabilities to the container. 

I was thinking if we could run without root (unprivileged). But I do not know enough about eBPF security to know if there is a possibility to securely configure it.

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ OpenTelemetry Go Automatic Instrumentation adheres to [Semantic Versioning](http
 
 ## [Unreleased]
 
+### Changed
+
+- Documentation does no longer say that `SYS_PTRACE` capabilty is needed. ([#388](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/388))
```
```suggestion
- Documentation no longer says that `SYS_PTRACE` capabilty is needed. ([#388](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/388))
```

MrAlias (approved):

PR #390, opened 2023-10-16
internal/pkg/instrumentors/utils: fix GetLinuxKernelVersion logic

zchee:
Fix GetLinuxKernelVersion logic.

In the case of using the `gcr.io/distroless/static:nonroot` base image, occur below error:
```
Malformed version: 6.1.42+\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000
```

I didn't check  the `Utsname` result yet because my laptop is macOS, but using the `go.opentelemetry.io/otel/sdk/resource.WithOS` attribute result is here:

```
Debian GNU/Linux Debian GNU/Linux 11 (bullseye) (Linux store-service-6c577fb4ff-gtm28 6.1.42+ #1 SMP PREEMPT_DYNAMIC Sat Aug 19 13:30:09 UTC 2023 x86_64)
```


MrAlias:
Do you have a link to documentation about what values are expected in this field? It would be nice to know this is addressing all use-cases we are missing.

zchee:
@MrAlias I found another approach to get kernel version.

zchee:
@MrAlias PTAL

internal/pkg/instrumentors/utils/kernel.go
```diff
@@ -15,28 +15,39 @@
...
+			// If we see anything else, we are likely to mis-parse it.
+			values[vi] = value
+			vi++
+			if vi >= len(values) {
+				break
+			}
+			value = 0
+		}
 	}
+	ver := fmt.Sprintf("%s.%s", strconv.Itoa(values[0]), strconv.Itoa(values[1]))
```
MrAlias, line 50: This doesn't seem right, shouldn't this be of the form "X.X.X" to match the existing behavior?

internal/pkg/instrumentors/utils/kernel.go
MrAlias, line 1: Can you add tests to ensure compatibility and prevent regressions of the bug your fixing?

internal/pkg/instrumentors/utils/kernel.go
```diff
@@ -15,28 +15,39 @@
...
 
-	var buf [65]byte
-	for i, v := range utsname.Release {
-		buf[i] = byte(v)
-	}
-
-	ver := string(buf[:])
-	if strings.Contains(ver, "-") {
-		ver = strings.Split(ver, "-")[0]
+	var (
```
MrAlias, line 32: Was this algorithm taken from somewhere? Can you please cite where it's from if so?

zchee: already commented. Should I write more details?

- https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/390/files#diff-a79ffed82bd81b27d495d5ab6eee4090890e9d9e8f0df170b97b841703bf2ceeR25

internal/pkg/instrumentors/utils/kernel.go
zchee: Fix later.

RonFed: This is the use we are doing in this function:
```
ver, err := utils.GetLinuxKernelVersion()
...

minVersion := version.Must(version.NewVersion("5.14"))
log.Logger.V(0).Info("Detected linux kernel version", "version", ver)
if ver.GreaterThanOrEqual(minVersion) {
advice = syscall.MADV_WILLNEED | MadvisePopulateRead | MadvisePopulateWrite
}
```

Since we are comparing to 5.14 I think it should be enough to take two segments of the version.
In addition, I don't see a reason why not to call https://pkg.go.dev/internal/syscall/unix#KernelVersion directly.
Also this might be worth considering: https://pkg.go.dev/github.com/moby/moby/pkg/parsers/kernel

zchee: @RonFed 
> Since we are comparing to 5.14 I think it should be enough to take two segments of the version.

I thought too.

> In addition, I don't see a reason why not to call https://pkg.go.dev/internal/syscall/unix#KernelVersion directly

That package path is `internal/syscall/unix`, we can't call cuz have `internal`, therefore I copied.

> Also this might be worth considering: https://pkg.go.dev/github.com/moby/moby/pkg/parsers/kernel

Interesting, I'll check it tomorrow.

RonFed: @zchee Any updates ?

zchee: @RonFed Sorry, what do you want any change? 

> Since we are comparing to 5.14 I think it should be enough to take two segments of the version.

Current change will be get `6.16` or etc. Should I change to get `X.X.X` as @MrAlias said?

zchee:
@MrAlias @RonFed PTAL

edeNFed (approved):

RonFed (approved):

edeNFed:
lgtm
@MrAlias please take another look

internal/pkg/instrumentors/utils/kernel.go
RonFed: I think it's Ok.

MrAlias (approved):

PR #373, opened 2023-10-13
Add a Get Started doc with three scenarios

theletterf:
Fixes #230
Fixes #231
Fixes #232
Fixes https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/204 
Fixes https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/27

I've followed @pellared 's guidance here, but I'm pretty sure the instructions need more work and validation.

PR #392, opened 2023-10-18
docs: Update compiling the project

pellared:
Tested together with @theletterf when working on https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/373


CONTRIBUTING.md
```diff
@@ -15,11 +15,21 @@ join the meeting or get in touch on
...
+Additionaly it requires (the latest versions are prefered):
+
+- `clang` - Clang C Language Family Frontend for LLVM
+- `gcc` - GNU Compiler Collection
+- `go` - Go
+- `libbpf-dev` - eBPF helper library
+- `llvm` - LLVM
+- `make` - Make
+
+You can build the Go Automatic Instrumnentation binary by running:
```
theletterf, line 28:
```suggestion
Building the project requires Linux with kernel 4.4 or higher.
In addition, use the latest versions of the following components:

- `clang`
- `gcc`
- `go`
- `libbpf-dev`
- `llvm`
- `make`

Build the Go Automatic Instrumnentation binary by running:
```

CONTRIBUTING.md
```diff
@@ -15,11 +15,21 @@ join the meeting or get in touch on
...
 `make build`
 
-Windows/Mac users will need to compile this project inside a Linux docker container by running:
-`make docker-build`
+Alternatively, you build inside a Linux Docker container by running:
```
theletterf, line 31:
```suggestion
Alternatively, you can create a Linux Docker container:
```

theletterf (approved):
Approved with a couple suggestions for rewording

CONTRIBUTING.md
```diff
@@ -15,11 +15,21 @@ join the meeting or get in touch on
...
-Linux users can build this repository by running:
+Building requires Linux with kernel 4.4 or higher.
+In addition, use the latest versions of the following components:
+
+- `clang`
+- `gcc`
+- `go`
+- `libbpf-dev`
+- `llvm`
+- `make`
```
MrAlias, line 26: Can we specify minimum versions of support here? Likely just state what versions you have tested building with and state something to the effect "other versions may work but we have tested with these".

pellared: 1. I do not have time to dig into this.
2. In .NET Automatic instrumentation we did it initially but it occurred that we did not want to waste time on updating it especially that we did not have many contributors because of the complexity of the repository. Listing just the packages was good enough so far

Side note: 

- https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/664b3638025bdb511aa22df975a0d77d9fb20ea2/.github/workflows/build.yaml#L25
- https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/664b3638025bdb511aa22df975a0d77d9fb20ea2/Dockerfile#L3

You can create an issue if you think that it will be needed.

pellared: I do not see a lot value in this because:

1. Our CI currently installs the latest available packages from the repository
2. Probably older versions would work as well
3. I do not believe we would remember to bump these unless we pin the tool versions used in CI. If we would pin the tool versions in the CI then this would make more sense in my opinion.

I think this information is still valuable even without specifying the minimum or tested version.
However, I would not block if you want to provide such information as a separate PR.

MrAlias: I was thinking this information would be more informative for users on their local systems. If they are running on an old OS would there be any issue with them compiling?

pellared: I am not a C expert 😉 
I guess as long as it is possible to install the tools and update libc then it should even work on old OSes. Maybe it requires kernel 4.4, but I am not sure 🙈

MrAlias (approved):

PR #393, opened 2023-10-18
Add struct field offset tracking package: `structfield`

### LINKED ISSUE #372: The offset file format does not evaluate back-ported patches
When a package releases a patch version to a prior minor version the offsets generator does not evaluate that new patch version if a cached version range beyond the minor version exists. This means the offset for that patch version may be incorrect.

For example, [`google.golang.org/grpc`](https://pkg.go.dev/google.golang.org/grpc) just released [`v1.58.3`](https://github.com/grpc/grpc-go/releases/tag/v1.58.3). Our offsets for struct fields from that package include a range with a max greater than `v1.58.X`. For example,

https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/31048acf70316b2221d2909dae6204ea72472357/internal/pkg/inject/offset_results.json#L51-L72

When the offsets are "updated" the cached values are used and given `1.58.3` falls within that range the new offset is not detected, the cached value is used instead.

### Possible solution

This may motivate moving back to the prior offsets format that listed each version explicitly (without version ranges).

### Related PR

https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/45

MrAlias:
Resolves #372 

## Problem

Currently, the offset data model is defined over a range over versions. The start and end of the range are the only two versions defining the range. This means when a package introduces a version within that range (i.e. in the range of `1.0.0` to `1.60.0-rc1` the version `1.58.2` is added) that package is implicitly assumed to contain the same offset as the existing range. This is not a valid assumption though.

Additionally, the current structure the offsets are tracked within is designed for data-storage not developer use. Instead of having to manage complex looping and merge operations, it would be ideal if a simpler API can be provided to interact with this data.

## Solution

This change redefines the offset data model. Instead of defining offsets over a range it states explicitly each version an offset applies to. If an offset is not found in this list it is evaluated before being added by the `offsetgen` utility.

Additionally, it wraps this new data-model in a developer-friendly API.

### `internal/pkg/structfield` package

A new `structfield` package is added to scope all functionality and types related to offsets (including the data model).

This splits from the existing design where this used to live in `internal/pkg/inject`.

#### `Index`

Within the `structfield` package a new type named `Index` is defined to hold all offset data for struct fields.

```go
// Index holds all struct field offsets.
type Index struct { data map[ID]*Offsets }

// ID is a struct field identifier for an offset.
type ID struct {
	// PkgPath package import path containing the struct field.
	PkgPath string
	// Struct is the name of the struct containing the field.
	Struct string
	// Field is the field name.
	Field string
}
```

This new type uses the Go `map` as a lookup mechanism instead of parsing through multiple slices as was done for the `TrackedOffsets`.

The values of the map are the new `Offsets` type. This type wraps an underlying `map` of version to offset values for concurrent safe access.

#### Access comparison

This new type reduces puts of offsets down to:

```go
index.PutOffset(id, offset)
```

The prior version of inserting into a `TrackeOffset` is to complex to show here, but it would include iterating through four levels of structs and maps to find the correct place to insert a version/offset pair.

The new type reduces gets of offsets down to:

```go
off, ok := index.GetOffset(id, ver)
```

Compared to the prior:

```go
off, ok := data.GetOffset(structName, field, ver)
 ```


MrAlias:
The `offsets` package name is a useful var name, alternates should be explored.

MrAlias:
> The `offsets` package name is a useful var name, alternates should be explored.

Ideas:
- `fieldpos`
- `structfield`

pellared (approved):

PR #394, opened 2023-10-18
Replace `internal/pkg/inject.Injector`

MrAlias:
The `Injector` type in `internal/pkg/inject` is used to build a map of values that will be injected into the C runtime as `const` using the `ebpf.CollectionSpec`. This type is complex, untested, and has design flaws (evident in the use of control flags passed to methods).

This introduces a new function `inject.Constants` as a replacement for the `Injector` type.

```go
// Constants injects key-values defined by opts into spec as constant. The keys
// are used as volatile const names and the values are the const values.
//
// If duplicate or colliding Options are passed, the last one passed is used.
func Constants(spec *ebpf.CollectionSpec, opts ...Option) error
```

The `Options` passed to this function are created with the following functions:

```go
// WithRegistersABI returns an option that will set the "is_registers_abi" to
// value. This information can be determined from the IsRegistersABI method of
// the TargetDetails in "go.opentelemetry.io/auto/internal/pkg/process".
//
// Commonly this is called like the following:
//
//	WithRegistersABI(target.IsRegistersABI())
func WithRegistersABI(value bool) Option 

// WithAllocationDetails returns an option that will set "total_cpus",
// "start_addr", and "end_addr".
func WithAllocationDetails(details process.AllocationDetails) Option 

// WithKeyValue returns an option that will set key to value.
func WithKeyValue(key string, value interface{}) Option

// WithOffset returns an option that sets key to the offset value of the struct
// field defined by strct and field at the specified version ver.
//
// If the offset value is not known, an error is logged and a no-op Option is
// returned.
func WithOffset(key, strct, field string, ver *version.Version) Option
```

All of these `Options` are equivalent to prior operations the `Injector` used to perform. Additionally, all of these `Options` are tested for their functionality.


pellared (approved):
👍

PR #395, opened 2023-10-18
Update Dockerfile

MrAlias:
- Use golang:1.21.3-bullseye as base build image
  - This addresses CVEs from Go 1.20 and allows dependabot to manage updates
- Cache `go.mod` and `go.sum` to speed up builds

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,15 @@
-FROM debian:12 as builder
+FROM golang:1.21.3-bullseye as builder
+
 ARG TARGETARCH
-RUN apt-get update && apt-get install -y curl clang gcc llvm make libbpf-dev -y
-RUN curl -LO https://go.dev/dl/go1.20.linux-${TARGETARCH}.tar.gz && tar -C /usr/local -xzf go*.linux-${TARGETARCH}.tar.gz
-ENV PATH="/usr/local/go/bin:${PATH}"
 WORKDIR /app
+
+RUN apt-get update && apt-get install -y curl clang gcc llvm make libbpf-dev
+
+# pre-copy/cache go.mod for pre-downloading dependencies and only redownloading
+# them in subsequent builds if they change
+COPY go.mod go.sum ./
+RUN go mod download && go mod verify
+
 COPY . .
 RUN make build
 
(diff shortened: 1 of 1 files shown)


edeNFed (approved):

Dockerfile
```diff
@@ -1,9 +1,15 @@
-FROM debian:12 as builder
+FROM golang:1.21.3-bullseye as builder
+
 ARG TARGETARCH
```
pellared, line 3: This line can be removed.

pellared: Done https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/395/commits/2b924878f9fe89f58f46c63f2ee87b7863a0d14f

pellared (approved):

Dockerfile
```diff
@@ -1,9 +1,14 @@
...
-ENV PATH="/usr/local/go/bin:${PATH}"
+FROM golang:1.21.3-bullseye as builder
+
 WORKDIR /app
+
+RUN apt-get update && apt-get install -y curl clang gcc llvm make libbpf-dev
+
+# pre-copy/cache go.mod for pre-downloading dependencies and only redownloading
+# them in subsequent builds if they change
+COPY go.mod go.sum ./
```
pellared, line 9: I think we should also do the same for `internal/tools/go.mod` and `internal/tools/go.sum`.

pellared (approved):

Dockerfile
```diff
@@ -1,9 +1,14 @@
...
+RUN go mod download && go mod verify
```
pellared, line 10: I think we should also do the same for `internal/tools/go.mod` and `internal/tools/go.sum`.

pellared (approved):

Dockerfile
pellared:
```suggestion
# pre-copy/cache go.mod for pre-downloading dependencies and only redownloading
# them in subsequent builds if they change
COPY go.mod go.sum internal/tools/go.mod internal/tools/go.sum ./
RUN go mod download && go mod verify && cd internal/tools && go mod download && go mod verify
```

MrAlias: None of the tools are used in the build though, right?

pellared: Right 😉

PR #399, opened 2023-10-19
Remove the errors pkg

MrAlias:
All the errors contained in the "errors" package are used in the process package. Just define and export the errors from there.


cli/main.go
```diff
@@ -57,7 +57,7 @@ func main() {
 	}()
 
 	log.Logger.V(0).Info("starting instrumentors...")
-	if err = inst.Run(ctx); err != nil && err != errors.ErrInterrupted {
+	if err = inst.Run(ctx); err != nil && err != process.ErrInterrupted {
```
pellared, line 60: I would prefer using  https://pkg.go.dev/errors#Is (instead of `!=`) especially that this code is not covered by tests.

pellared (approved):

cli/main.go
MrAlias: Sure. I was just preserving existing behavior, but I can add this change to the scope of this PR if there is no objections.

pellared (approved):

pdelewski (approved):

PR #401, opened 2023-10-20
Replace log pkg use with explicit loggers

MrAlias:
Instead of using a global logger from the `internal/pkg/log` package, pass loggers through the instrumentation. This distribution of a logger allows for a few things the global approach does not provide:

- Deterministic. There is not global inconsistency (or panics).
- Easily switch out the loggers. When you know all logging of a type/function are done with a passed logger you have a central point to replace the logger they use.
- Scope a logger to the type/function
  - Verbosity, common attributes, or namespace can be set for the passed logger
  - Names are built from the parent logger. Meaning a call stack can be determined by the logger names. When type A passes a logger to type C those log messages will have different name than when type B passes a logger to type C.
- Loggers can be replaced singularly for testing.

Additional cleanup:

- If an error is returned, don't log it. Have the caller decided how to handle it, including if it should be logged.
- Make helper functions for types that log methods of that type. Therefore they will have a scoped logger without an additional argument.
- Don't set verbosity to `V(0)`. This is the default, so just log at the default of the logger.
  - If we want to increase the verbosity of a component in the future we can do so at the logger level by not using `V(0)`.


pellared (approved):

cli/main.go
```diff
@@ -17,27 +17,41 @@ package main
...
+	} else {
+		logger = zapr.NewLogger(zapLog)
 	}
 
-	log.Logger.V(0).Info("building OpenTelemetry Go instrumentation ...")
+	return logger
+}
+
+func main() {
+	logger := newLogger().WithName("OpenTelemetry-Go-Auto")
```
pellared, line 49: loose proposal 
```suggestion
	logger := newLogger().WithName("go.opentelemetry.io/auto")
```

PR #402, opened 2023-10-20
chore: Use errors.Is instead of comparing equality

MrAlias:


internal/pkg/process/discover.go
```diff
@@ -84,7 +84,7 @@ func (a *Analyzer) findProcessID(target *TargetArgs) (int, error) {
 
 	for {
 		dirs, err := proc.Readdir(15)
-		if err == io.EOF {
+		if errors.Is(err, io.EOF) {
```
pellared, line 87: The current code will always work as otherwise it would be a breaking change of https://pkg.go.dev/os#File.Readdir.

I think we can keep as it is.

internal/tools/inspect/manifest.go
```diff
@@ -105,7 +105,7 @@ func gotoEntry(r *dwarf.Reader, tag dwarf.Tag, name string) bool {
 func findEntry(r *dwarf.Reader, tag dwarf.Tag, name string) (*dwarf.Entry, error) {
 	for {
 		entry, err := r.Next()
-		if err == io.EOF || entry == nil {
+		if errors.Is(err, io.EOF) || entry == nil {
```
pellared, line 108: I do not understand why we are even checking for `io.EOF`. See https://pkg.go.dev/debug/dwarf#Reader

```suggestion
		if entry == nil {
```

MrAlias: Yeah, good point. Now that I see it again, I made the same conclusion when I looked at this a few months ago.

internal/pkg/process/discover.go
MrAlias: Returning a wrapped `io.EOF` wouldn't be a breaking change in the Go ecosystem. What's the downside of making this change?

PR #412, opened 2023-10-22
Install dependencies for dependabot-pr action

MrAlias:
To run the needed make commands, these dependencies are needed.


.github/workflows/create-dependabot-pr.yml
```diff
@@ -18,6 +18,10 @@ jobs:
           go-version: "~1.21.2"
           check-latest: true
 
+      - name: Install dependencies
+        run: |
+          sudo apt-get update && sudo apt-get install -y clang llvm libbpf-dev
```
pellared, line 23: It might be handy to move this to some script as I think this command is already present in Dockerfile and other GH workflows.

MrAlias: Sure. I'm not opposed to doing that in a follow-up. This is just copied from the other workflows, which does indeed show repetition.

I'm going to merge this as is though now given it fixes the workflow to be usable.

MrAlias: Tracking with https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/413

pellared: Sure 👍

PR #417, opened 2023-10-25
Add the WithEnv option

### LINKED ISSUE #415: Add a WithEnv option
**Is your feature request related to a problem? Please describe.**

Environment variable precedence is a subjective matter. Some users would like the passed options to take precedence to them and other would like the opposite.

It would be ideal if this was configurable

Add a `WithEnv` `InstrumentationOption` that will parse the relevant environment variables and set config values.

The precedence of this option will be based on the order it is passed. If first, the following options will have precedence. And if last, it will have precedence.

MrAlias:
Resolve #415 

Add the `WithEnv` `InstrumentationOption` to configure `Instrumentation` to parse the environment. The `Instrumentation` will no longer by default parse the environment. This option needs to be used to enable environment parsing, and the order it is passed influences the environment precedence. All options passed before this one will be overridden if there are conflicts, and those passed after will override the environment. 

cc @RonFed


pellared (approved):

MrAlias:
The `offsets` package name is a useful var name, alternates should be explored.

RonFed:
lgtm

edeNFed (approved):

Issue #187, opened 2023-06-08
OTEL_GO_AUTO_TARGET_EXE process not found


gauravluckme:
The path OTEL_GO_AUTO_TARGET_EXE has been provided and the executable is also there at the container path.

However the Go auto instrumentation is not able to find it.

Error: "process not found yet, trying again soon"

gauravluckme:
N/A

zengyuxing007:
I encountered the same problem.

```shell
{"level":"info","ts":1698304109.2718306,"caller":"process/discover.go:61","msg":"process not found yet, trying again soon","exe_path":"/app/main"}
{"level":"info","ts":1698304111.2719026,"caller":"process/discover.go:61","msg":"process not found yet, trying again soon","exe_path":"/app/main"}
{"level":"info","ts":1698304113.2719774,"caller":"process/discover.go:61","msg":"process not found yet, trying again soon","exe_path":"/app/main"}
```

zengyuxing007:
> I encountered the same problem.
> 
> ```shell
> {"level":"info","ts":1698304109.2718306,"caller":"process/discover.go:61","msg":"process not found yet, trying again soon","exe_path":"/app/main"}
> {"level":"info","ts":1698304111.2719026,"caller":"process/discover.go:61","msg":"process not found yet, trying again soon","exe_path":"/app/main"}
> {"level":"info","ts":1698304113.2719774,"caller":"process/discover.go:61","msg":"process not found yet, trying again soon","exe_path":"/app/main"}
> ```

should added this to the pod spec:

```yaml
    spec:
      shareProcessNamespace: true
```

PR #422, opened 2023-10-26
Document the need for `shareProcessNamespace` in k8s pod spec

zengyuxing007:
Fixed issue #187


README.md
```diff
@@ -126,8 +126,9 @@ To instrument an application running in Kubernetes, follow these steps:
        runAsUser: 0
        privileged: true
    ```
+2. Check if the configuration [shareProcessNamespace](https://kubernetes.io/docs/tasks/configure-pod-container/share-process-namespace/) is present in the pod spec, if not, please add it.
```
MrAlias, line 129: Does this need to be added to the autoinstrumentation-go pod spec (above), just the application being run's pod spec, or both?

MrAlias: Ah, pod level would mean it is for all running images:

https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/b1ee64b52e3a65c4535642f88e7575661c9e0134/.github/workflows/e2e/k8s/sample-job.yml#L14

We don't support a daemon set yet. Nevermind.

MrAlias (approved):

pellared (approved):
Thanks 👍

PR #428, opened 2023-10-28
Update HTTP server ebpf - add status code

RonFed:
The main goal of this PR is to prepare an easier way to collect HTTP server related fields by eBPF.
* Instead of collecting the data at the entry uprobe, collect it at the exit one. This makes sense since all the relevant fields are set at the end of the handling and we can read them.
* In order to read the desired fields in the return probe, we must save the pointer to the relevant struct in the entry probe. This is because registers are not guaranteed to keep their value through the function.
* We save a pointer to a `net/http.(*response)` struct at the entry probe.
*  I added the `net/http.response` struct to the offsets being tracked (currently the `req` and `status` fields of that struct are being tracked, and we'll add more in the future to support more attributes).
* Adding the above field required changing the template file for generating the dummy HTTP application for the offset calculations, since the current one didn't include the symbol for the field.
* Moving the attributes collection to the exit probe is probably relevant to all the instrumentations. This will make the UPROBE_RETURN macro not so usable, but for now,  we should keep it until those changes are made.


edeNFed (approved):
looks great

Issue #420, opened 2023-10-25
Provide a default implementation of the Probe


MrAlias:
Currently all of the instrumentation provides their own implementations of the `Probe` interface. There is a lot of repeated, mostly untested, code shared across all of these implementations.

Instead, provide a base implementation for instrumentation to use. This will use the instrumentation manifest (#419) to construct a tail made `Probe`.

The `Manager` will continue to work with `Probe`s. This means it will still be possible for new instrumentation that is really different to provide their own. Or they can just wrap the default and override a method or two.

### TODO

- [x] PoC draft showing full change (likely too large to be submitted for a single PR): https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/451
- [x] PR base implementation: https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/452
- [x] PR refactor gRPC server use of base implementation: #477 
- [x] PR refactor gRPC client use of base implementation: #478
- [x] PR refactor net/http server use of base implementation: #479
- [x] PR refactor net/http client use of base implementation: #480
- [x] PR refactor the gin use of base implementation: #481
- [x] PR refactor database/sql use of base implementation: #482

Issue #419, opened 2023-10-25
Centralize the source-of-truth for instrumentation objects


MrAlias:
Currently, the constants injected into an eBPF runtime and the function uprobes are defined in multiple locations for instrumentation:

- In the `FuncNames` method
- In the `Load` method
- In the offsetgen utility

Instead, have this information held in a new instrumentation manifest type. This manifest can be provided by the instrumentation and act as a source of truth for it and others as to what it needs.

### High level proposal

```go
// Manifest contains information about a package being instrumented.
type Manifest struct {
	// Name is the name of the instrumentation.
	Name string

	// Modules are the instrumented Go modules.
	Modules []Module
}

// Module defines an instrumented Go module.
type Module struct {
	// ModulePath is the module path of the module.
	ModulePath string

	// StructFields are the struct fields in an instrumented package that are
	// used for instrumentation.
	StructFields []StructField
	// StructMethods are the struct methods in an instrumented package that are
	// used for instrumentation.
	StructMethods []StructMethod
}

// StructField defines a struct field that is used in instrumentation.
type StructField struct {
	// PkgPath is the unique import path of the package containing the struct.
	PkgPath string
	// Struct is the name of the struct.
	Struct string
	// Field is the name of the struct field.
	Field string
}


// StructMethod defines a struct method that is used in instrumentation.
type StructMethod struct {
	// PkgPath is the unique import path of the package containing the struct.
	PkgPath string
	// Struct is the name of the struct.
	Struct string
	// Method is the name of the struct Method within Struct.
	Method string
	// Indirect defines if the method's receiver is a pointer to the struct.
	Indirect bool
}
```

The `FuncNames` and `LibraryName` of the `Instrumentor` can be replace with a new `Manifest` method that returns the above type.

### Alternative considered

A more distributed structure for struct fields could be used. I.e.

```
module
└── package
    └── struct
        ├── field
        └── method
```

However, by themselves a field or method type are useless. They need the context of what struct and package they belong to. Given the optimization of space is not critical here, but the re-usability is, this option was abandoned.

PR #439, opened 2023-10-30
Add default instrumentation probe implementation

MrAlias:
- Return a manifest from an instrumentation probe to centralize the source of truth for all of its struct and func needs (#419).
- Provide a default probe implementation (#420)


MrAlias:
Splitting into smaller PRs.

PR #440, opened 2023-10-31
Fix kill cmd in sample-job

MrAlias:
Currently, the syntax of the kill command is invalid on the linux systems the job is being run on. Update the command to use correct syntax, be verbose for debugging purposes, include 1000ms timeouts, and progress from TERM to KILL.

### DIFF
--- a/.github/workflows/e2e/k8s/sample-job.yml
+++ b/.github/workflows/e2e/k8s/sample-job.yml
@@ -22,7 +22,7 @@ spec:
         imagePullPolicy: IfNotPresent
         command: ["/bin/sh", "-c"]
         # send SIGTERM to otel-go-instrumentation once the sample app has generated data so the job completes.
-        args: ["/sample-app/main && kill -TERM $(pidof otel-go-instrumentation)"]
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e/k8s/sample-job.yml
```diff
@@ -22,7 +22,7 @@ spec:
         imagePullPolicy: IfNotPresent
         command: ["/bin/sh", "-c"]
         # send SIGTERM to otel-go-instrumentation once the sample app has generated data so the job completes.
-        args: ["/sample-app/main && kill -TERM $(pidof otel-go-instrumentation)"]
+        args: ["/sample-app/main && kill --verbose --timeout 1000 TERM --timeout 1000 KILL $(pidof otel-go-instrumentation)"]
```
pellared, line 25: should it not be something like

```suggestion
        args: ["/sample-app/main && kill --verbose --signal TERM --timeout 1000 KILL $(pidof otel-go-instrumentation)"]
```

pellared: oh I see that we are giving 1sec for the app, right?

pellared (approved):

MrAlias:
The underlying container uses a version of `kill` that does not support these flags. Closing.

PR #442, opened 2023-11-01
Add scope version

MrAlias:


internal/pkg/opentelemetry/controller.go
```diff
@@ -40,7 +41,10 @@ func (c *Controller) getTracer(libName string) trace.Tracer {
 		return t
 	}
 
-	newTracer := c.tracerProvider.Tracer(libName)
+	newTracer := c.tracerProvider.Tracer(
+		libName,
```
pellared, line 45: How about following the recommendation from https://opentelemetry.io/docs/concepts/instrumentation-scope/

> it is recommended to use the name and version of the Instrumentation Library as the instrumentation scope

and adding a prefix to clearly define the source of the instrumentation

```suggestion
		"go.opentelemetry.io/auto" + libName,
```

pellared (commented):
I propose to also update the instrumentation scope name to make it clear that the instrumentation comes from OTel Go Auto and that the version is related to the instrumentation library and NOT instrumented library.

I think that changing both together would be more consistent.

internal/pkg/opentelemetry/controller.go
MrAlias: Please see https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/444.

I agree with the proposal, but there is more work that needs to be done first before we can update the instrumentation names.

pellared (approved):

PR #447, opened 2023-11-01
Replace `inspect.StructField` with `structfield.ID`

MrAlias:
Unify the way we define these fields in the instrumentation and in the generation.


pellared (approved):
👍

PR #451, opened 2023-11-02
[PoC] Add and use base implementation of the Probe

MrAlias:
Part of https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/420

The plan is to split this into smaller PRs. See the linked issue for more information.

PR #452, opened 2023-11-02
Add the base implementation of the Probe

MrAlias:
Part of https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/420

All of the instrumentation provided by the project themselves provide a `Probe`. Each instrumentation redefines the same common set of functionality. Instead, unify this functionality into a single implementation.

This new `Base` implementation can be returned directly by each instrumentation package (which is the plan). If they need added functionality they can always embed the type in their own implementation, or create a completely independent implementation. This is because the `Manager` still accepts the `Probe` interface not the added `Base` type.

For a complete picture of how this will be used, refer to https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/451. Follow-up PRs are going to be created to use this implementation in each instrumentation package.


internal/pkg/instrumentation/probe/probe.go
```diff
@@ -41,4 +55,254 @@ type Probe interface {
...
+// UprobeFunc is a function that will attach a eBPF program to a perf event
+// that fires when the given symbol starts executing in exec.
+//
+// It is expected the symbol belongs to are shared library and its offset can
+// be determined using target.
+//
+// Losing the reference to the resulting Link (up) will close the Uprobe and
+// prevent further execution of prog. The Link must be Closed during program
+// shutdown to avoid leaking system resources.
+type UprobeFunc[BPFObj any] func(symbol string, exec *link.Executable, target *process.TargetDetails, obj *BPFObj) ([]link.Link, error)
```
RonFed, line 228: The Uprobe attaching functions are very similar between all the probes (some may have return probes while others don't), but maybe we can have a generic function for loading the uprobes defined in here?

RonFed (approved):
Looks great

internal/pkg/instrumentation/probe/probe.go
MrAlias: The hard part is going to be that each function needs to access a particular field of the `BPFObj`. We could also add an additional interface that each probe could provide that would access the correct field, but that seems to also be about equivalent to having each probe provide an implementation for this.

I think we can explore options though in a future PR after this is merged.

PR #470, opened 2023-11-07
Add semantic convention required attributes to the `net/http` instrumentation

alonkeyval:


alonkeyval:
This pull request adds a new HTTP span attribute that includes the host, remote address, and protocol information.
Screenshot 2023-11-07 at 17 09 59

RonFed (approved):

MrAlias:
Please run `make`. It looks like there are a few linting issues that need to be resolved.

MrAlias:
Closing/reopening to re-trigger actions.

internal/test/e2e/nethttp/traces.json
```diff
@@ -68,6 +68,30 @@
...
+                  }
+                },
+                {
+                  "key": "net.peer.name",
+                  "value": {
+                    "stringValue": "::1"
+                  }
+                },
+                {
+                  "key": "net.peer.port",
```
MrAlias, line 91: These need to be updated to follow the [`v1.23` semantic conventions](https://github.com/open-telemetry/semantic-conventions/blob/v1.23.0/docs/http/http-spans.md#common-attributes).

MrAlias (changes requested):

internal/test/e2e/nethttp/traces.json
RonFed: `go.opentelemetry.io/otel/semconv/v1.21.0` is the latest one we can import now, right?
I couldn't find `v1.23` in that repo.

MrAlias: Ah, that looks correct. We need to update there first. Resolving.

internal/pkg/instrumentation/bpf/net/http/server/bpf/probe.bpf.c
```diff
@@ -26,13 +26,19 @@ char __license[] SEC("license") = "Dual MIT/GPL";
 #define MAX_CONCURRENT 50
 #define W3C_KEY_LENGTH 11
 #define W3C_VAL_LENGTH 55
+#define REMOTE_ADDR_MAX_LEN 32
+#define HOST_MAX_LEN 32
```
MrAlias, line 30: I think this needs to be 253 to support https://www.rfc-editor.org/rfc/rfc7230.html#section-5.4 right?

Or maybe 255 given this will include the port, right?

internal/pkg/instrumentation/bpf/net/http/server/probe_test.go
```diff
@@ -48,6 +48,12 @@ func TestProbeConvertEvent(t *testing.T) {
 		Method: [8]byte{0x47, 0x45, 0x54},
 		// "/foo/bar"
 		Path: [128]byte{0x2f, 0x66, 0x6f, 0x6f, 0x2f, 0x62, 0x61, 0x72},
+		// "localhost:8080"
+		RemoteAddr: [32]byte{0x6c, 0x6f, 0x63, 0x61, 0x6c, 0x68, 0x6f, 0x73, 0x74, 0x3a, 0x38, 0x30, 0x38, 0x30, 0x0},
+		// "localhost:8080"
+		Host: [32]byte{0x6c, 0x6f, 0x63, 0x61, 0x6c, 0x68, 0x6f, 0x73, 0x74, 0x3a, 0x38, 0x30, 0x38, 0x30, 0x0},
```
MrAlias, line 54: Can you make these different to show that the net host name comes from one and the peer attributes come from another?

internal/pkg/instrumentation/bpf/net/http/server/probe_test.go
```diff
@@ -48,6 +48,12 @@ func TestProbeConvertEvent(t *testing.T) {
 		Method: [8]byte{0x47, 0x45, 0x54},
 		// "/foo/bar"
 		Path: [128]byte{0x2f, 0x66, 0x6f, 0x6f, 0x2f, 0x62, 0x61, 0x72},
+		// "localhost:8080"
+		RemoteAddr: [32]byte{0x6c, 0x6f, 0x63, 0x61, 0x6c, 0x68, 0x6f, 0x73, 0x74, 0x3a, 0x38, 0x30, 0x38, 0x30, 0x0},
```
MrAlias, line 52: I think the server will always set this to "IP:port": https://pkg.go.dev/net/http#Request

If not, we need to revise the storage length to accommodate the address space this could contain.

internal/test/test_helpers/utilities.bash
```diff
@@ -42,7 +42,8 @@ redact_json() {
 		jq --sort-keys '
 			del(
 				.resourceSpans[].scopeSpans[].spans[].startTimeUnixNano,
-				.resourceSpans[].scopeSpans[].spans[].endTimeUnixNano
+				.resourceSpans[].scopeSpans[].spans[].endTimeUnixNano,
+				(.resourceSpans[].scopeSpans[].spans[].attributes[] | select(.key == "net.peer.port"))
```
MrAlias, line 46: If the peer port is changed into something like the trace/span IDs (i.e. "******") we will still be able to evaluate they are included. By removing them, as is done here, it is removing the ability to test the attribute is added.

internal/pkg/instrumentation/bpf/net/http/server/bpf/probe.bpf.c
```diff
@@ -243,6 +252,14 @@ int uprobe_HandlerFunc_ServeHTTP(struct pt_regs *ctx)
     return 0;
 }
 
+void read_go_string(void *base, int offset, char *output, int maxLen, const char *errorMsg) {
+    void *ptr = (void *)(base + offset);
+    if (!get_go_string_from_user_ptr(ptr, output, maxLen)) {
+        bpf_printk("Failed to get %s", errorMsg);
```
MikeGoldsmith, line 258: `bpfprintf` calls can be fairly expensive, we should avoid them whenever possible. 

Also, it looks like this is wrapping the `get_go_string_from_user_ptr` helper func but does some extra work by combining base and offset - is this needed or could we just use the helper direclty?

internal/pkg/instrumentation/bpf/net/http/server/bpf/probe.bpf.c
MikeGoldsmith: Yeah, host will be the full string value so needs to be 255.

MikeGoldsmith (changes requested):
Looks to be on the right track 👍🏻  

Please can you update your PR title to have a valid prefix and add a PR body?

internal/pkg/instrumentation/bpf/net/http/server/bpf/probe.bpf.c
alonkeyval: bpf_printf call is triggered only in the case of an error, reducing the overall impact on performance.

internal/pkg/instrumentation/bpf/net/http/server/bpf/probe.bpf.c
alonkeyval: change both REMOTE_ADDR_MAX_LEN and HOST to 256

CHANGELOG.md
```diff
@@ -33,6 +33,7 @@ OpenTelemetry Go Automatic Instrumentation adheres to [Semantic Versioning](http
 - Add the `WithTraceExporter` `InstrumentationOption` to configure the trace `SpanExporter` used by an `Instrumentation`. ([#426](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/426))
 - Add HTTP status code attribute to `net/http` server instrumentation. ([#428](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/428))
 - The instrumentation scope now includes the version of the auto-instrumentation project. ([#442](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/442))
+- Add http relevant attributes. ([#470]https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/470)
```
MrAlias, line 36:
```suggestion
- Add `net.host.name`, `net.protocol.name`, `net.peer.name`, and `net.peer.port` attributes to HTTP server spans. ([#470]https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/470)
```

internal/test/e2e/nethttp/traces.json
```diff
@@ -6,7 +6,7 @@
           {
             "key": "process.runtime.description",
             "value": {
-              "stringValue": "go version 1.21.4 linux/amd64"
+              "stringValue": "go version 1.21.4 linux/arm64"
```
RonFed, line 9: change back to amd64

MrAlias (approved):

PR #477, opened 2023-11-10
Refactor gRPC server probe to use base probe implementation

MrAlias:
Part of #420


internal/pkg/instrumentation/bpf/google.golang.org/grpc/server/probe.go
```diff
@@ -15,205 +15,123 @@
...
+	offset, err := target.GetFunctionOffset(name)
 	if err != nil {
-		return err
+		return nil, err
 	}
-	g.headersProbe = hProbe
 
-	rd, err := perf.NewReader(g.bpfObjects.Events, os.Getpagesize())
+	opts := &link.UprobeOptions{Address: offset}
+	l, err := exec.Uprobe("", obj.UprobeDecodeStateDecodeHeader, opts)
```
grcevski, line 121: It might be good to rename this BPF function in the BPF code, to match to the name of the function where the probe attaches to - `operateHeaders`. In older versions of gRPC the function used to be called `decodeHeader`, but that has been removed and the probe now attaches to `operateHeaders`.

grcevski (commented):
The refactor looks good to me! I just left one minor comment to match the function name in the bpf code to what the probe attaches now to.

internal/pkg/instrumentation/bpf/google.golang.org/grpc/server/probe.go
MrAlias: Sounds good to me. I'll open an issue to track this as a follow up. :+1:

MrAlias: #497

pellared (approved):

PR #495, opened 2023-11-13
Only copy Go files in example Dockerfiles

MrAlias:


RonFed (approved):

examples/httpPlusdb/Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM golang:1.21.4
 WORKDIR /app
-COPY . .
+COPY ./main.go ./main.go
```
pdelewski, line 3: Currently examples are only one file apps, however that might change, should not we copy all go files
```
COPY ./*.go *.go
```

PR #503, opened 2023-11-15
Fallback to the single offset if possible and version can't be parsed

RonFed:
The main goal of this PR is to handle cases where an offset is required for a package with an  un-official version (contains commit hash: for example `v0.0.0-20230227094218-b8c73b2037b8`
Currenly we can't handle such cases since we don't find a valid version.
I added a fallback to check if only a single offset were saved across all versions to use that offset. If more than one offset was saved and the version is not foundd, fail.
A more robust solution is probably required to handle this kind of version, but this fallback is a good solution until then.

### DIFF
--- a/internal/pkg/structfield/structfield.go
+++ b/internal/pkg/structfield/structfield.go
@@ -220,6 +220,11 @@ type Offsets struct {
 
 	// values is a map between version and offset value.
 	values map[verKey]offsetVersion
+
+	// singleVer is a pointer to the single version of the Offsets.
+	// If there is only one version, this will be set to that version.
+	// Otherwise, this will be nil.
+	singleVer *offsetVersion
 }
 
 // NewOffsets returns a new empty *Offsets.
@@ -238,13 +243,11 @@ func (o *Offsets) Get(ver *version.Version) (uint64, bool) {
 	v, ok := o.values[newVerKey(ver)]
 	o.mu.RUnlock()
 
-	if !ok && len(o.values) == 1 {
+	if !ok && o.singleVer != nil {
 		// If we don't have the exact version, but we only have one version, we
 		// fallback to use that version. This can happen when a non official version is being used
 		// which contains commit hash in the version string.
-		for _, v := range o.values {
-			return v.offset, true
-		}
+		return o.singleVer.offset, true
 	}
 	return v.offset, ok
 }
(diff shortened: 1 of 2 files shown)


internal/pkg/inject/offset_results.json
```diff
@@ -448,6 +451,9 @@
                       "1.0.2",
                       "1.0.3",
                       "1.0.4",
+                      "1.0.5",
```
MrAlias, line 454: Something seems off here, this version of the `google.golang.org/grpc/internal/transport` package does not exist as far as I can tell.

I'm also not able to find any version prior to [`v1.14.0`](https://pkg.go.dev/google.golang.org/grpc@v1.14.0/internal/transport?tab=versions)

internal/pkg/structfield/structfield.go
```diff
@@ -220,6 +224,11 @@ type Offsets struct {
 
 	// values is a map between version and offset value.
 	values map[verKey]offsetVersion
+
+	// singleOffset is the single offset in the values map.
+	// If there is only one offset, this will be that offset.
+	// Otherwise, this will be invalidOffset.
+	singleOffset uint64
```
MrAlias, line 231: A type will likely be more useful here so the zero value is unset and a "special" offset doesn't need to be used. E.g.:

```go
type single struct {
	value uint64
	valid bool
}
```

note, the naming is not well thought through but just and example.

MrAlias (commented):
What are the consequences if the offset is wrong here? I'm guessing the best case scenario is the telemetry contains garbage, but is the worse case that the user application crashes?

I worry that this guess becomes problematic at the end of the version range. If a user is asking for an offset that is released after our last know version we have less confidence that value hasn't changed. Can we address that?

MrAlias (commented):
What are the consequences if the offset is wrong here? I'm guessing the best case scenario is the telemetry contains garbage, but is the worse case that the user application crashes?

I worry that this guess becomes problematic at the end of the version range. If a user is asking for an offset that is released after our last know version we have less confidence that value hasn't changed. Can we address that?

RonFed:
> What are the consequences if the offset is wrong here? I'm guessing the best case scenario is the telemetry contains garbage, but is the worse case that the user application crashes?
> 
> I worry that this guess becomes problematic at the end of the version range. If a user is asking for an offset that is released after our last know version we have less confidence that value hasn't changed. Can we address that?

It depends if the offset is used to read from or write to the user app. Most (all except one that I can think of) of the offsets are used for attribute collection, so a wrong offset will lead to garbage value in the worst case (some probes drop the span if an attribute cna't be read, maybe we should change that). For those we use for writing (adding HTTP header for example), a wrong offset can cause a crash in the worst case. There are a lot of checks in the eBPF code in the writing to user memory parts, to minimize that chance, and the only case which I can think of this might happen is the HTTP header update (the offset there is part of the standard library)

Regarding newer versions, we can identify this case as a different one from the hash commit version and maybe act differently?

internal/pkg/inject/offset_results.json
RonFed: I can remove this for now, originally added it because of the offsets test

MrAlias (approved):

PR #506, opened 2023-11-15
Update helm install cmd for fixtures targ

MrAlias:
This command has started to fail on my system. I'm guessing it is due to and upgrade:

```console
$ helm version
version.BuildInfo{Version:"v3.12.0", GitCommit:"c9f554d75773799f72ceef38c51210f1842a1dea", GitTreeState:"clean", GoVersion:"go1.20.4"}
```

As far as I can tell this works for old versions of helm as well.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -157,7 +157,7 @@ fixtures/%:
 	if [ ! -d "opentelemetry-helm-charts" ]; then \
 		git clone https://github.com/open-telemetry/opentelemetry-helm-charts.git; \
 	fi
-	helm install test -f .github/workflows/e2e/k8s/collector-helm-values.yml opentelemetry-helm-charts/charts/opentelemetry-collector
+	helm install test -f .github/workflows/e2e/k8s/collector-helm-values.yml opentelemetry-helm-charts/opentelemetry-collector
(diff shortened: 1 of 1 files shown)


pellared (approved):

damemi:
Think I figured this out, the Makefile command uses a locally cloned copy of the helm charts github repo: https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/910f58ad679db2a6bc02e7e6b6fb69e097ea2f7c/Makefile#L168-L170

Where the charts are located under `charts`

I was hitting this error too and my local folder was empty somehow. When I deleted it and re-ran the fixture command I got the right output.

In the e2e we just use the checkout action and build from there: https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/910f58ad679db2a6bc02e7e6b6fb69e097ea2f7c/.github/workflows/kind.yml#L57-L62

That said I don't think this `helm repo add` command is necessary if we're cloning a local copy: https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/910f58ad679db2a6bc02e7e6b6fb69e097ea2f7c/Makefile#L167

PR #519, opened 2023-11-20
Add telemetry distro name & version resource attributes

### LINKED ISSUE #239: Add `telemetry.auto.name` resource attribute
See https://github.com/open-telemetry/semantic-conventions/pull/178

MikeGoldsmith:
Updates the resource attributes configured by the agent to use the new semantic conventions, `telemetry.distro.name` and `telemetry.distro.version`. The existing version attribute has not been removed yet. 

- Closes #239

### DIFF
--- a/instrumentation.go
+++ b/instrumentation.go
@@ -28,6 +28,7 @@ import (
 	"github.com/go-logr/stdr"
 	"github.com/go-logr/zapr"
 	"go.opentelemetry.io/contrib/exporters/autoexport"
+	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracehttp"
 	"go.opentelemetry.io/otel/sdk/resource"
 	"go.opentelemetry.io/otel/sdk/trace"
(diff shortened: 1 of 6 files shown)


instrumentation.go
```diff
@@ -243,7 +244,11 @@ func (c instConfig) res() *resource.Resource {
 		semconv.SchemaURL,
 		semconv.ServiceNameKey.String(c.serviceName),
 		semconv.TelemetrySDKLanguageGo,
+		// will be removed from semconv in the future
 		semconv.TelemetryAutoVersionKey.String(Version()),
+		// will be added to semconv in the future
```
pellared, line 249: Wouldn't it be better to do it after semconv bump (to v1.21)??

MikeGoldsmith: I checked and the [v1.21](https://pkg.go.dev/go.opentelemetry.io/otel@v1.21.0/semconv/v1.21.0) package didn't include the new keys. 

image

MikeGoldsmith: We could wait for new semconv package. I don't see the harm in adding early as we'll want to duplicate the fields for a period of time anyway.

pellared: I meant 1.23 :facepalm:

MrAlias:
Closing. Will replace with a new PR that uses the newest release of `semconv`.

PR #522, opened 2023-11-21
Add `WithResourceAttributes` option

edeNFed:
This PR adds `WithAdditionalResourceAttributes` which allows users to pass additional resource attributes that should be reported by the instrumentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,7 +30,7 @@ OpenTelemetry Go Automatic Instrumentation adheres to [Semantic Versioning](http
 - The instrumentation scope now includes the version of the auto-instrumentation project. ([#442](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/442))
 - Add a new `WithSampler` method allowing end-users to provide their own implementation of OpenTelemetry sampler directly through the package API. ([#468](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/468)).
 - Add uprobes to `execDC` in order to instrument SQL DML. ([#475](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/475))
-- Add `WithAdditionalResourceAttributes` `InstrumentationOption` to configure `Instrumentation` to add additional resource attributes. ([#522](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/522))
+- Add `WithResourceAttributes` `InstrumentationOption` to configure `Instrumentation` to add additional resource attributes. ([#522](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/522))
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ OpenTelemetry Go Automatic Instrumentation adheres to [Semantic Versioning](http
 - The instrumentation scope now includes the version of the auto-instrumentation project. ([#442](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/442))
 - Add a new `WithSampler` method allowing end-users to provide their own implementation of OpenTelemetry sampler directly through the package API. ([#468](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/468)).
 - Add uprobes to `execDC` in order to instrument SQL DML. ([#475](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/475))
+- Add `WithAdditionalResourceAttributes` `InstrumentationOption` to configure `Instrumentation` to add additional resource attributes.
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


instrumentation.go
```diff
@@ -396,3 +408,13 @@ func WithSampler(sampler trace.Sampler) InstrumentationOption {
...
+
+// WithAdditionalResourceAttributes returns an [InstrumentationOption] that will
+// configure an [Instrumentation] to use the provided attributes as additional
+// OpenTelemetry Resource attributes.
+func WithAdditionalResourceAttributes(attrs []attribute.KeyValue) InstrumentationOption {
+	return fnOpt(func(_ context.Context, c instConfig) (instConfig, error) {
+		c.additionalResAttrs = attrs
+		return c, nil
+	})
+}
```
pellared, line 420: How about making the API similar to https://pkg.go.dev/go.opentelemetry.io/otel/metric#WithAttributes?

Something like:

```suggestion
func WithResourceAttributes(attrs ...attribute.KeyValue) InstrumentationOption {
	return fnOpt(func(_ context.Context, c instConfig) (instConfig, error) {
		c.additionalResAttrs = append(c.additionalResAttrs, attrs...)
		return c, nil
	})
}
```

CHANGELOG.md
```diff
@@ -30,6 +30,7 @@ OpenTelemetry Go Automatic Instrumentation adheres to [Semantic Versioning](http
 - The instrumentation scope now includes the version of the auto-instrumentation project. ([#442](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/442))
 - Add a new `WithSampler` method allowing end-users to provide their own implementation of OpenTelemetry sampler directly through the package API. ([#468](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/468)).
 - Add uprobes to `execDC` in order to instrument SQL DML. ([#475](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/475))
+- Add `WithAdditionalResourceAttributes` `InstrumentationOption` to configure `Instrumentation` to add additional resource attributes.
```
pellared, line 33:
```suggestion
- Add `WithAdditionalResourceAttributes` `InstrumentationOption` to configure `Instrumentation` to add additional resource attributes. ([#522](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/522))
```

RonFed (approved):

instrumentation_test.go
```diff
@@ -19,8 +19,11 @@ import (
 	"fmt"
 	"testing"
 
+	"go.opentelemetry.io/otel/attribute"
+
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+
 	semconv "go.opentelemetry.io/otel/semconv/v1.21.0"
```
pellared, line 27: why the empty lines in between?

instrumentation.go
```diff
@@ -396,3 +408,13 @@ func WithSampler(sampler trace.Sampler) InstrumentationOption {
 		return c, nil
 	})
 }
+
+// WithResourceAttributes returns an [InstrumentationOption] that will configure
+// an [Instrumentation] to use the provided attributes as OpenTelemetry resource
+// attributes.
```
pellared, line 414: Can we document that the attributes are appended. Maybe:

```suggestion
// WithResourceAttributes returns an [InstrumentationOption] that will configure
// an [Instrumentation] to add the provided attributes to the OpenTelemetry resource.
```

instrumentation_test.go
```diff
@@ -127,6 +129,18 @@ func TestOptionPrecedence(t *testing.T) {
...
+	testAttributes := []attribute.KeyValue{
+		semconv.K8SContainerName("test_container_name"),
+		semconv.K8SPodName("test_pod_name"),
+	}
+
+	// Use WithResourceAttributes to config the additional resource attributes
+	c, err := newInstConfig(context.Background(), []InstrumentationOption{WithResourceAttributes(testAttributes...)})
+	require.NoError(t, err)
+	assert.Equal(t, testAttributes, c.additionalResAttrs)
+}
```
pellared, line 142: How about something like

```suggestion
func TestWithResourceAttributes(t *testing.T) {
	attr1 := semconv.K8SContainerName("test_container_name"),
	attr2 := semconv.K8SPodName("test_pod_name"),
	attr3 := semconv.K8SNamespaceName("test_namespace_name"),

	// Use WithResourceAttributes to config the additional resource attributes
	c, err := newInstConfig(context.Background(), []InstrumentationOption{WithResourceAttributes(attr1, attr2), WithResourceAttributes(attr3)})
	require.NoError(t, err)
	assert.Equal(t, []attribute.KeyValue{attr1, attr2, attr3}, c.additionalResAttrs)
}
```

It would avoid a regression bug if `WithResourceAttributes` would replace instead of append.

MikeGoldsmith (commented):
Outside of @pellared's suggestions, I think this looks good.

Have you done any testing of how this works alongside the SDK picking up resource attributes from the `OTEL_RESOURCE_ATTRIBUTES` environment variable? I believe it should be fine as the two sets of attributes will be merged as part of the `resource.NewWithAttributes` call.

edeNFed:
Thanks @MikeGoldsmith I added some more test cases

pellared (approved):
👍

Issue #352, opened 2023-09-27
Add manual instrumentation support


pellared:
**Is your feature request related to a problem? Please describe.**

One of the key aspects of observability is to have the ability to create instruments which are oriented around the application logic. Without this possibility the automatic instrumentation is more like "monitoring" tool.

**Describe the solution you'd like**

Add the application developer the possibility to augment the instrumentation in the application so that they can:
- create custom spans (and tracers)
- create custom metrics (and meters)
- attach additional instrumentation libraries

For example, let's consider this application which is based on https://opentelemetry.io/docs/instrumentation/go/getting-started/:

```golang
package main

import (
	"io"
	"log"
	"math/rand"
	"net/http"
	"strconv"

	"go.opentelemetry.io/otel"
	"go.opentelemetry.io/otel/attribute"
	"go.opentelemetry.io/otel/metric"
)

func main() {
	http.HandleFunc("/rolldice", rolldice)

	log.Fatal(http.ListenAndServe(":8080", nil))
}

var (
	tracer  = otel.Tracer("rolldice")
	meter   = otel.Meter("rolldice")
	rollCnt metric.Int64Counter
)

func init() {
	var err error
	rollCnt, err = meter.Int64Counter("dice.rolls",
		metric.WithDescription("The number of rolls by roll value"),
		metric.WithUnit("{roll}"))
	if err != nil {
		panic(err)
	}
}

func rolldice(w http.ResponseWriter, r *http.Request) {
	ctx, span := tracer.Start(r.Context(), "roll")
	defer span.End()

	roll := 1 + rand.Intn(6)

	rollValueAttr := attribute.Int("roll.value", roll)
	span.SetAttributes(rollValueAttr)
	rollCnt.Add(ctx, 1, metric.WithAttributes(rollValueAttr))

	resp := strconv.Itoa(roll) + "\n"
	if _, err := io.WriteString(w, resp); err != nil {
		log.Printf("Write failed: %v\n", err)
	}
}
```

Key requirements:

1. The telemetry recorded in the application code (`roll` spans and `dice.rolls` metrics) is exported recorded and exported via the tracer provider  and meter provider that is set up by Go Automatic Instrumentation.
1. The telemetry recorded in the application code is related (as child) with the spans created by Go Automatic Instrumentation.

So the manual telemetry should be integrated with the telemetry produced by  Go Automatic Instrumentation.

[...]

PR #523, opened 2023-11-22
Manual instrumentation support - Phase 1

RonFed:
This PR addresses some of the functionality described in #352 .
The following change adds the ability to integrate with manually created spans via the OpenTelemetry APIs.
Currently the eBPF collects the start and end time, span name and attributes passed by the user via `SetAttributes`.
This functionality is disabled by default and can be enabled via an env var or passing an Instrumentation option `WithGlobal()`
Added e2e tests.
This is only a partial implementation, the following are known gaps that will be added in future PRs:
* Slice attributes.
* Events
* Span errors

Next steps until full implementation are described in https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/352#issuecomment-1811166718


internal/include/otel_types.h
```diff
@@ -0,0 +1,145 @@
...
+	case FLOAT64:
+		bpf_probe_read(&attr->value, sizeof(s64), &go_attr_value->numeric);
+		return true;
+	case STRING:
+		if (go_attr_value->string.len <= 0){
+			return false;
+		}
+		if (go_attr_value->string.len >= OTEL_ATTRIBUTE_VALUE_MAX_LEN) {
+			// String value is too large
+			return false;
```
edeNFed, line 78: Worth logging a message with `bpf_printk` to let the user know

internal/include/otel_types.h
```diff
@@ -0,0 +1,146 @@
...
+		return true;
+	case STRING:
+		if (go_attr_value->string.len <= 0){
+			return false;
+		}
+		if (go_attr_value->string.len >= OTEL_ATTRIBUTE_VALUE_MAX_LEN) {
+			bpf_printk("Aattribute string value is too long\n");
+			return false;
+		}
+		if (!get_go_string_from_user_ptr(&go_attr_value->string, attr->value, OTEL_ATTRIBUTE_VALUE_MAX_LEN)) {
```
MikeGoldsmith, line 80: Could this be simplified to the following?

```suggestion
		return get_go_string_from_user_ptr(&go_attr_value->string, attr->value, OTEL_ATTRIBUTE_VALUE_MAX_LEN)
```

internal/include/otel_types.h
```diff
@@ -0,0 +1,146 @@
...
+			return false;
+		}
+		return true;
+	// TODO: handle slices
+	case BOOLSLICE:
+	case INT64SLICE:
+	case FLOAT64SLICE:
+	case STRINGSLICE:
+	case INVALID:
```
MikeGoldsmith, line 89: Maybe move the INVALID case to a separate case and log a message so we can see when something unexpected happens?

MikeGoldsmith (commented):
Is the intention here that you configure your application with the OTel SDK and out it into non-recording (eg no export) mode then the auto-instrumentation process watches for the tracer starting spans and duplicate them so they can be exported? It kinda feels wasteful as we'll be creating every span twice.

edeNFed (approved):

MrAlias (changes requested):
This doesn't seem to be inline with the proposed solutions: https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/352#issuecomment-1811166718

> Provide a TracerProvider implementation from auto-instrumentation that manual instrumentation can explicitly use to say they want their data to go to the same pipeline. This means the auto-instrumentation will control the type that manual instrumentation uses to create, edit, and end spans if they want the data to be sent to the auto-instrumentation pipeline

The proposal was to provide a `TracerProvider` implementation here that we would detect being used by other processes. The changes proposed in this PR are instrumenting the global implementation from `go.opentelemetry.io/otel`.

There are going to be serious limitations to this approach given the global from `otel` provides delegation support (i.e. the `nonRecordingSpan` or `Tracer` there will not be called when the user sets a `TracerProvider`).

I'm not sure this is a viable solution.

MrAlias:
> This doesn't seem to be inline with the proposed solutions: [#352 (comment)](https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/352#issuecomment-1811166718)
> 
> > Provide a TracerProvider implementation from auto-instrumentation that manual instrumentation can explicitly use to say they want their data to go to the same pipeline. This means the auto-instrumentation will control the type that manual instrumentation uses to create, edit, and end spans if they want the data to be sent to the auto-instrumentation pipeline
> 
> The proposal was to provide a `TracerProvider` implementation here that we would detect being used by other processes. The changes proposed in this PR are instrumenting the global implementation from `go.opentelemetry.io/otel`.
> 
> There are going to be serious limitations to this approach given the global from `otel` provides delegation support (i.e. the `nonRecordingSpan` or `Tracer` there will not be called when the user sets a `TracerProvider`).
> 
> I'm not sure this is a viable solution.

This is not addressing point (1) from the linked solutions, it is addressing point (2).

MrAlias:
Talked with @edeNFed in SIG meeting:

This is only meant to address the situation where a user has not registered an SDK with the global and they opt-in with the auto-instrumentation to record this telemetry. This is not intended to capture the telemetry from a default OTel SDK.

internal/include/otel_types.h
```diff
@@ -0,0 +1,145 @@
...
+
+#define INVALID 0
+#define BOOL 1
+#define INT64 2
+#define FLOAT64 3
+#define STRING 4
+#define BOOLSLICE 5
+#define INT64SLICE 6
+#define FLOAT64SLICE 7
+#define STRINGSLICE 8
```
MrAlias, line 32: Shouldn't we inject these into the eBPF space given they are constants we can import in the Go space? This seems highly error prone.

internal/include/otel_types.h
```diff
@@ -0,0 +1,145 @@
...
+	case STRING:
+		if (go_attr_value->string.len <= 0){
+			return false;
+		}
+		if (go_attr_value->string.len >= OTEL_ATTRIBUTE_VALUE_MAX_LEN) {
+			bpf_printk("Aattribute string value is too long\n");
+			return false;
+		}
+		return get_go_string_from_user_ptr(&go_attr_value->string, attr->value, OTEL_ATTRIBUTE_VALUE_MAX_LEN);
+	// TODO: handle slices
```
MrAlias, line 81: Please add an issue to track this.

internal/include/otel_types.h
```diff
@@ -0,0 +1,145 @@
...
+} go_otel_attr_value_t;
+
+typedef struct go_otel_key_value {
+	struct go_string     key;
+	go_otel_attr_value_t value;
+} go_otel_key_value_t;
+
+#define OTEL_ATTRIBUTE_KEY_MAX_LEN      (32)
+#define OTEL_ATTRIBUTE_VALUE_MAX_LEN    (128)
+#define OTEL_ATTRUBUTE_MAX_COUNT        (16)
```
MrAlias, line 48: These seem small (the default limits are higher than some of these https://github.com/open-telemetry/opentelemetry-go/blob/204be6157f13594eb92a91914aaa7a7561bfd08a/sdk/trace/span_limits.go#L26). Is there any way we can support more dynamic string lengths?

internal/pkg/instrumentation/bpf/go.opentelemetry.io/otel/global/bpf/probe.bpf.c
```diff
@@ -0,0 +1,173 @@
...
+struct {
+	__uint(type, BPF_MAP_TYPE_PERF_EVENT_ARRAY);
+} events SEC(".maps");
+
+// Injected in init
+volatile const u64 span_name_pos;
+volatile const u64 span_attributes_pos;
+
+// This instrumentation attaches uprobe to the following function:
+// func (t *tracer) Start(ctx context.Context, name string, opts ...trace.SpanStartOption) (context.Context, trace.Span)
```
MrAlias, line 69:
```suggestion
// func (t *tracer) Start(ctx context.Context, name string, opts ...trace.SpanStartOption) (context.Context, trace.Span)
// https://github.com/open-telemetry/opentelemetry-go/blob/98b32a6c3a87fbee5d34c063b9096f416b250897/internal/global/trace.go#L149
```

internal/pkg/instrumentation/bpf/go.opentelemetry.io/otel/global/bpf/probe.bpf.c
```diff
@@ -0,0 +1,173 @@
...
+
+    // Save the span name in map to be read once the Start function returns
+    void *context_ptr_val = get_Go_context(ctx, 3, 0, true);
+    void *key = get_consistent_key(ctx, context_ptr_val);
+    bpf_map_update_elem(&span_name_by_context, &key, &span_name, 0);
+    return 0;
+}
+
+// This instrumentation attaches uprobe to the following function:
+// func (t *tracer) Start(ctx context.Context, name string, opts ...trace.SpanStartOption) (context.Context, trace.Span)
```
MrAlias, line 88:
```suggestion
// func (t *tracer) Start(ctx context.Context, name string, opts ...trace.SpanStartOption) (context.Context, trace.Span)
// https://github.com/open-telemetry/opentelemetry-go/blob/98b32a6c3a87fbee5d34c063b9096f416b250897/internal/global/trace.go#L149
```

internal/pkg/instrumentation/bpf/go.opentelemetry.io/otel/global/bpf/probe.bpf.c
```diff
@@ -0,0 +1,173 @@
...
+// func (*nonRecordingSpan) SetAttributes(...attribute.KeyValue)
+SEC("uprobe/SetAttributes")
+int uprobe_SetAttributes(struct pt_regs *ctx) {
+    void *non_recording_span_ptr = get_argument(ctx, 1);
+    struct otel_span_t *span = bpf_map_lookup_elem(&active_spans_by_span_ptr, &non_recording_span_ptr);
+    if (span == NULL) {
+        return 0;
+    }
+
+    // In Go, "..." is equilaent to passing a slice: https://go.dev/ref/spec#Passing_arguments_to_..._parameters
```
MrAlias, line 147:
```suggestion
    // In Go, "..." is equivalent to passing a slice: https://go.dev/ref/spec#Passing_arguments_to_..._parameters
```

internal/pkg/instrumentation/bpf/go.opentelemetry.io/otel/global/bpf/probe.bpf.c
```diff
@@ -0,0 +1,173 @@
...
+
+    bpf_map_update_elem(&active_spans_by_span_ptr, &span_ptr_val, otel_span, 0);
+    start_tracking_span(ret_context_ptr_val, &otel_span->sc);
+    return 0;
+}
+
+// This instrumentation attaches uprobe to the following function:
+// func (*nonRecordingSpan) SetAttributes(...attribute.KeyValue)
+SEC("uprobe/SetAttributes")
+int uprobe_SetAttributes(struct pt_regs *ctx) {
```
MrAlias, line 140: The following methods need to be included to provide a full implementation:

- `IsRecording`
  - This is critical given many users will use this method to gate other telemetry annotation operations.
- `SetStatus`
- `SetError`
- `RecordError`
- `AddEvent`
- `SetName`

The tracking of status and events will mean that the `probe.Event` needs to be updated to support these.

internal/pkg/instrumentation/bpf/go.opentelemetry.io/otel/global/bpf/probe.bpf.c
```diff
@@ -0,0 +1,173 @@
...
+    struct otel_span_t *zero_span = bpf_map_lookup_elem(&otel_span_storage_map, &zero_span_key);
+    if (zero_span == NULL) {
+        return 0;
+    }
+
+    u32 otel_span_key = 1;
+    // Zero the span we are about to build, eBPF doesn't support memset of large structs (more than 1024 bytes)
+    bpf_map_update_elem(&otel_span_storage_map, &otel_span_key, zero_span, 0);
+    // Get a pointer to the zeroed span
+    struct otel_span_t *otel_span = bpf_map_lookup_elem(&otel_span_storage_map, &otel_span_key);
```
MrAlias, line 110: Was the [C++ OTel implementation](https://github.com/open-telemetry/opentelemetry-cpp/tree/main/sdk/src/trace) considered for use here instead?

internal/test/e2e/otelglobal/main.go
```diff
@@ -0,0 +1,55 @@
...
+	span.SetAttributes(attribute.String("inner.key", "inner.value"))
+}
+
+func createMainSpan(ctx context.Context) {
+	ctx, span := tracer.Start(ctx, "parent")
+	defer span.End()
+	intAttr := attribute.Int("int_key", 42)
+
+	innerFunction(ctx)
+
```
MrAlias, line 40:
```suggestion

	innerFunction(ctx)

	intAttr := attribute.Int("int_key", 42)
```

Makefile
```diff
@@ -157,7 +158,7 @@ fixtures/%:
 	if [ ! -d "opentelemetry-helm-charts" ]; then \
 		git clone https://github.com/open-telemetry/opentelemetry-helm-charts.git; \
 	fi
-	helm install test -f .github/workflows/e2e/k8s/collector-helm-values.yml opentelemetry-helm-charts/opentelemetry-collector
+	helm install test -f .github/workflows/e2e/k8s/collector-helm-values.yml opentelemetry-helm-charts/charts/opentelemetry-collector
```
MrAlias, line 161: This will revert https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/506. Does the current command not work for you?

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ OpenTelemetry Go Automatic Instrumentation adheres to [Semantic Versioning](http
 
 ### Added
 
+- Manual instrumentation support - Phase 1 ([#523]https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/523)
```
MrAlias, line 13: This needs to be updated to describe to the end user the functionality added. Ideally it will include the CLI flag when added and describe the details of its use.

examples/rolldice/main.go
MrAlias, line 1: The whole example will need to be updated, including the description and expected output. I would suggest maybe doing this in a separate PR and tracking with its own issue.

Makefile
RonFed: Yes, that change broke the fixture commands for me.
```
 ~ helm version
version.BuildInfo{Version:"v3.13.0", GitCommit:"825e86f6a7a38cef1112bfa606e4127a706749b1", GitTreeState:"clean", GoVersion:"go1.21.1"}
```

internal/include/otel_types.h
RonFed: Done: https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/525

internal/pkg/instrumentation/bpf/go.opentelemetry.io/otel/global/probe.go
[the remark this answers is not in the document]
RonFed: Changed to `traceglobal`

internal/include/otel_types.h
RonFed: I agree, changed

internal/pkg/instrumentation/bpf/go.opentelemetry.io/otel/global/bpf/probe.bpf.c
RonFed: I don't think it is possible since eBPF is run inside the kernel, and the allowed functions to call is quite limited https://ebpf.io/what-is-ebpf/#helper-calls

instrumentation.go
```diff
@@ -53,6 +54,9 @@ const (
 	// envTracesExportersKey is the key for the environment variable value
 	// containing what OpenTelemetry trace exporter to use.
 	envTracesExportersKey = "OTEL_TRACES_EXPORTER"
+	// envOtelAPIIntegrationKey is the key for the environment variable value
+	// containing whether to create spans from non recording spans declared manually.
+	envOtelAPIIntegrationKey = "OTEL_GO_AUTO_MANUAL_INTEGRATION"
```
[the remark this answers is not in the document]
RonFed: Added the CLI flag and removed this env var. I'm not sure what is the most suitable documentation for that flag, and whether the `usage` string I added is enough.

internal/include/otel_types.h
RonFed: There is but it was challenging to make the eBPF verifier happy with more efficient encoding. I initially tried to pack all the string keys and values into the same buffer and separate them with null bytes. This will create a limit on the combined size of the strings instead of the length of each one and will allow more flexible string lengths. 
We can make the `OTEL_ATTRUBUTE_MAX_COUNT` bigger, but it will make the span entry sent to the perf buffer larger (few KB).

instrumentation.go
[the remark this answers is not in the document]
MrAlias: > I'm not sure what is the most suitable documentation for that flag, and whether the usage string I added is enough.

Yeah, the info added to the flag looks good to me :+1:

internal/pkg/instrumentation/bpf/go.opentelemetry.io/otel/global/bpf/probe.bpf.c
MrAlias: > `IsRecording` should be probed to always return true?

That's a good question. As long as the auto-instrumentation `TracerProvider` sampler doesn't return `DROP` this should return `true`: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#sampler

Is there a way we can inject a sampling function to determine what the sampler will return?

examples/rolldice/main.go
RonFed: Done https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/526

CHANGELOG.md
RonFed: Updated

internal/pkg/instrumentation/bpf/go.opentelemetry.io/otel/global/bpf/probe.bpf.c
RonFed: We can pass any relevant information from Go to eBPF that will be relevant to the sampling logic (but not code to execute).

Makefile
MrAlias: Hmm. This should be reverted by #527. It's still not working locally for me, but we can look into this later.

RonFed:
@MrAlias Can you please take another look?

internal/include/otel_types.h
```diff
@@ -0,0 +1,143 @@
...
+	// String values
+	if (vtype == attr_type_string) {
+		if (go_attr_value->string.len >= OTEL_ATTRIBUTE_VALUE_MAX_LEN) {
+			bpf_printk("Aattribute string value is too long\n");
+			return false;
+		}
+		return get_go_string_from_user_ptr(&go_attr_value->string, attr->value, OTEL_ATTRIBUTE_VALUE_MAX_LEN);
+	}
+
+	// TODO: handle slices
```
MrAlias, line 91:
```suggestion
	// TODO (#525): handle slices
```

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ OpenTelemetry Go Automatic Instrumentation adheres to [Semantic Versioning](http
 
 ### Added
 
+- Manual instrumentation support - Phase 1. Adding cli flag named `global-impl` that allows to record telemetry from the OpenTelemetry default global implementation. ([#523]https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/523)
```
```suggestion
- The CLI flag `global-impl` is added.
  This flag, when used, enables the instrumentation of the OpenTelemetry default global implementation (https://pkg.go.dev/go.opentelemetry.io/otel).
  This means that all trace telemetry from this implementation that would normally be dropped will instead be recorded with the auto-instrumentation pipeline. ([#523]https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/523)
```

internal/pkg/instrumentation/bpf/go.opentelemetry.io/otel/global/bpf/probe.bpf.c
MrAlias: `SetStatus` tracking issue: https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/538

MrAlias: `SetError` tracking issue: https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/539

MrAlias: `RecordError` tracking issue: https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/540

MrAlias: `AddEvent` tracing issue: https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/541

MrAlias: `SetName` tracking issue: https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/542

MrAlias (approved):

Issue #472, opened 2023-11-07
Reading the HTTP status code in server requests does not work with custom http.ResponseWriter's


grcevski:
### Describe the bug

A Go developer can write a custom http.ResponseWriter which does not have to have a status field and especially may also have a status field at a different offset.

Reading the non-existent status field will cause the HTTP status code to not be read or be read as random bytes.

### Environment

 - OS: Linux
 - Go Version: Any
 - Version: Any

### To Reproduce

Here's a simplified example that shows a custom ResponseWriter:

```
type responseWriterWrapper struct {
	rw         http.ResponseWriter
	statusCode int
	...
}

func newResponseWriterWrapper(rw http.ResponseWriter) *responseWriterWrapper {
	return &responseWriterWrapper{
		rw:         rw,
		statusCode: http.StatusOK,
	}
}

func (b *responseWriterWrapper) Header() http.Header {
	return b.rw.Header()
}

func (b *responseWriterWrapper) Write(data []byte) (int, error) {
	if b.statusCode == 0 {
		b.WriteHeader(http.StatusOK)
	}
	n, err := b.rw.Write(data)
        ...
	return n, err
}

func (b *responseWriterWrapper) WriteHeader(statusCode int) {
	b.statusCode = statusCode
	b.rw.WriteHeader(statusCode)
}
```

### Expected behavior

Status code should be correctly read from this example, however there is no `status` field in the ResponseWriter.

### Additional context

N/A

MrAlias:
Plan is to look at replacing the ServeHTTP return uprobe with a uprobe for WriteHeader.

MrAlias:
Given the `ResponseWriter` is an interface and we cannot instrument its methods directly, we will need to instrument its implementations. I think we are on the correct track currently instrumenting the `response` struct:

https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/998fabad1086c9df9149a41d8c4baf659711f7bc/internal/pkg/instrumentation/bpf/net/http/server/probe.go#L72

Though, instead of using the `status` field, we should create a uprobe for the `WriteHeader` method of that type. Given any wrapping of a `ResponseWriter` will ultimately need to call that method, I think that should solve this issue.

MrAlias:
I'm not able to reproduce the failure here. Running the following as in the end-2-end fixture test still passes:

```go
package main

import (
	"fmt"
	"io"
	"log"
	"net/http"
	"time"
)

[...]

PR #529, opened 2023-12-01
Ensure net http client code also works with the RoundTripper interface

grcevski:
With this PR I'm adding support for tracking HTTP client requests which use the RoundTripper interface. The client.Do probe will not catch these events, but there's a lower level rountrip method on Transport, which handles all HTTP client calls.

I'm also adding a new integration test which covers the now fixed issue for duplicate server requests and missing response codes (for https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/472), as well as this RoundTripper client call.


internal/pkg/instrumentation/bpf/net/http/client/probe.go
```diff
@@ -79,7 +79,7 @@ func New(logger logr.Logger) probe.Probe {
 			},
 		},
 		Uprobes: map[string]probe.UprobeFunc[bpfObjects]{
-			"net/http.(*Client).do": uprobeDo,
+			"net/http.(*Transport).roundTrip": uprobeDo,
```
RonFed, line 82: Probably better to rename `uprobeDo` to something like `uprobeRoundTrip`.
In addition, I think it is worth renaming the probe name in the `bpf.probe.c` file

Makefile
```diff
@@ -142,7 +142,8 @@ license-header-check:
 	           exit 1; \
 	   fi
 
-.PHONY: fixture-nethttp fixture-gin fixture-databasesql
+.PHONY: fixture-nethttp fixture-gin fixture-databasesql fixture-nethttp-custom
```
RonFed, line 145: If adding a fixture here, it is worth adding also one in the github workflows in `kind.yml`

grcevski:
Thanks @RonFed, I made the requested changes and fixed up the lint issues.

RonFed (approved):

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ OpenTelemetry Go Automatic Instrumentation adheres to [Semantic Versioning](http
 ### Added
 
 - Add `WithResourceAttributes` `InstrumentationOption` to configure `Instrumentation` to add additional resource attributes. ([#522](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/522))
+- Add HTTP client instrumentation for `Transport.roundTrip`. ([#529](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/529))
```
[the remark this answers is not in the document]
grcevski: Done, thanks!

pellared (approved):

PR #554, opened 2023-12-09
Add the ability to define an optional uprobe.

RonFed:
* Set the `SetAttributes` uprobe as optional. This is necessary for the case which a span is defined without attributes.  In the current implementation we would not find the `SetAttributes` symbol and remove the probe and kill the entire instrumentation. With this change it makes sense to define that uprobe as Optional.
* Change the `Uprobes` map to slice, as it wasn't used as a map.

### DIFF
--- a/internal/pkg/instrumentation/probe/probe.go
+++ b/internal/pkg/instrumentation/probe/probe.go
@@ -67,9 +67,9 @@ type Base[BPFObj any, BPFEvent any] struct {
 	// Consts are the constants that need to be injected into the eBPF program
 	// that is run by this Probe.
 	Consts []Const
-	// Uprobes is a mapping from runtime symbols to a UprobeFunc. These define
-	// the eBPF program triggers that need to setup for this Probe.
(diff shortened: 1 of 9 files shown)


internal/pkg/instrumentation/probe/probe.go
```diff
@@ -218,6 +222,18 @@ func (i *Base[BPFObj, BPFEvent]) Close() {
 // shutdown to avoid leaking system resources.
 type UprobeFunc[BPFObj any] func(symbol string, exec *link.Executable, target *process.TargetDetails, obj *BPFObj) ([]link.Link, error)
 
+// Uprobe is an eBPF program that is attached in the entry point and/or the reutrn of a function.
+type Uprobe[BPFObj any] struct {
+	// Sym is the symbol name of the function to attach the eBPF program to.
+	Sym string
+	// Fn is the function that will attach the eBPF program to the function
```
MrAlias, line 229:
```suggestion
	// Fn is the function that will attach the eBPF program to the function.
```

MrAlias (approved):

PR #558, opened 2023-12-13
Fix gRPC server inst detection of incoming headers

MrAlias:
Fix #557

The frame parameter position of operateHeaders changed in v1.60.0 of google.golang.org/grpc. Inject the parameter position based on version detection.

PR #574, opened 2023-12-28
Make gRPC frame pos deterministic for eBPF verifier

RonFed:
Following #558, on some kernels the verifier rejected the program because the value being read from `ctx` could not be determined in verification. Changed the passed value from Go to be a flag instead of the position.


edeNFed (approved):
lgtm

damemi (approved):

PR #607, opened 2024-01-16
Correct ID var/field casing

MrAlias:
- Acronyms should be capitalized the same way
- Preference to lowercase starting characters should be given to local variables

https://go.dev/talks/2014/names.slide#5


pellared (approved):

RonFed (approved):

dmathieu:
Shouldn't this have been caught by golangci-lint?

MrAlias:
> Shouldn't this have been caught by golangci-lint?

Indeed. Can you please create an issue to track this?

PR #608, opened 2024-01-17
Report running errors from manager and include loading errors

RonFed:
Following #605 and #586 
This PR solves a bug that happens if the instrumentation fails to load, and Close is called. The loading error wasn't sent to the errors channel which caused Close to block.


internal/pkg/instrumentation/manager.go
```diff
@@ -207,7 +210,7 @@ func (m *Manager) cleanup(target *process.TargetDetails) error {
 func (m *Manager) Close() error {
 	m.done <- true
 	m.wg.Wait()
```
pellared, line 212: There is no guarantee that that probes are finished as this line can be reached before `m.wg.Add(len(m.probes))` is called.

Shouldn't we call `m.wg.Wait()` after `err := <-m.runningErrors`?

internal/pkg/instrumentation/manager.go
```diff
@@ -207,7 +210,7 @@ func (m *Manager) cleanup(target *process.TargetDetails) error {
...
-	err := <-m.closingErrors
+	err := <-m.runningErrors
```
pellared, line 213: nit: Do I see correctly that `Run` and `Close` are returning the same error? If so then I find it awkward. Maybe we should rename it to `Stop` and not return an error? Especially that it is not even handeled:

https://github.com/RonFed/opentelemetry-go-instrumentation_fork/blob/36a6ebe436445344912565ce87a1dcfba9b17eca/instrumentation.go#L141-L146

RonFed: Good point, I forgot to handle that error in `Instrumentation.Close`.
Currently, there are 2 types of errors in the Run function which we can return:
* "Loading" errors which can occur if we fail to load an eBPF program at the start of `Run`.
* Closing errors that can occur while cleaning up the probes.

What do you think about returning the first type of error from `Run` and the second from `Close`?

internal/pkg/instrumentation/manager.go
RonFed: fixed

internal/pkg/instrumentation/manager.go
pellared: It will be for sure better

internal/pkg/instrumentation/manager.go
```diff
@@ -118,12 +118,14 @@ func (m *Manager) FilterUnusedProbes(target *process.TargetDetails) {
 // Run runs the event processing loop for all managed probes.
 func (m *Manager) Run(ctx context.Context, target *process.TargetDetails) error {
 	if len(m.probes) == 0 {
-		m.logger.Info("there are no available instrumentations for target process")
-		return nil
+		err := errors.New("there are no available instrumentations for target process")
```
MrAlias, line 121: nit

```suggestion
		err := errors.New("no instrumentation for target process")
```

MrAlias (approved):

RonFed:
A new grpc version was released yesterday, so offsets need to be updated

Issue #290, opened 2023-09-10
bpf_probe_write_user helper function is locked down since linux kernel `5.14-rc6`


SzymonSt:
### Describe the bug

Since this [commit](https://github.com/torvalds/linux/commit/51e1bb9eeaf7868db56e58f47848e364ab4c4129) in linux kernel repository, bpf_probe_write_user is locked down and this results in `unknown func bpf_probe_write_user` error during opentelemetry-go-instrumentation startup on systems with lockdown integrity enabled.

### Environment

 - OS: Ubuntu 22.04LTS Linux  - 5.15.0.1045.52-azure, (reproducible on all linux systems with kernel above 5.14 and integrity mode enabled)
 - Go Version: 1.20
 - Version: v0.2.1-alpha , v0.2.2-alpha

### To Reproduce

1. Build docker image from v0.2.1-alpha/v0.2.2-alpha
2. Run image on host with kernel version higher that 5.14

### Expected behavior

opentelemetry-go-instrumentation starts up without any issues.

### Additional context

bpf_probe_write_user should be probably retired from being used in opentelemetry-go-instrumentation as for many cloud providers/on demand use cases editing startup kernel parameters is not possible.

SzymonSt:
Problem also  encountered and reported here: [237](https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/237)

damemi:
From 3-Mar-24 sig meeting, at @grcevski started a thread on the Linux kernel mailing list to see about unlocking this: https://www.uwsg.indiana.edu/hypermail/linux/kernel/2403.0/03026.html

It doesn't look like that thread has any replies yet, is there any way to follow up with that @grcevski ?

damemi:
From SIG call today:

* We originally tried asking about this in Cillium slack, but didn't hear back
* Could possibly be fixed by LSM policies (are those usable in cloud providers though?)
* We should ask around eBPF/CNCF for advice or help with working on this in the kernel

grcevski:
Based on this article, it seems that the lockdown LSM policies are static and cannot be modified or configured:

https://lwn.net/Articles/791863/

Proposals to make them more configurable appears to have been rejected.

grcevski:
Relevant very new patch proposal https://lore.kernel.org/bpf/20240404190146.1898103-1-elver@google.com/

grcevski:
[...]

PR #620, opened 2024-01-22
Detect if the Linux Kernel is in lockdown mode

grcevski:
This PR adds a Linux Kernel lockdown mode validation check for the two bpf programs which use bpf_probe_write_user. If the kernel is running in lockdown mode (typically auto-enabled with Secure Boot), we now put an explanatory error message instead of 'failed to load bpf program error'.

Relates to https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/290


MrAlias (approved):

edeNFed:
I am not sure we should completely not load the instrumentation in case the kernel is in lockdown mode.
For example, process-internal context propagation would still work. 

What do you think about producing spans and adding an error to them that specify something like `context propagation failed please change kernel lockdown mode`. I think this can be a good tradeoff between producing the best data that we can during kernel lockdown but also not confusing the users with a log message they will not find.

grcevski:
Sure, I think I can do that. The only way I've been able to do this kind of thing is by using defines, compiling different versions of the BPF programs and loading selectively. I know about the `bpf_core_enum_value_exists` helper, but I've never been able to use it successfully, the verifier complains regardless of the `if` statement. Does this make sense, or do you have another approach in mind?

MrAlias:
@edeNFed what do you think about doing that in a follow up PR?

grcevski:
I'm happy to work on extending this PR and following with another that does the partial trace support. It will be easier for review if we are doing it incrementally, but I don't mind to extend this one if that's preferred.

edeNFed:
@grcevski @MrAlias I am good with both options. Whatever easier for you.

edeNFed (approved):

Issue #636, opened 2024-02-01
No traces for go applications even though process has found by instrumentation container.


jyoti-borade:
I have configured otel auto-instrumetation for three of my go services. 
I am getting below logs for each of them:

{"level":"info","ts":1706790698.5967553,"logger":"go.opentelemetry.io/auto","caller":"cli/main.go:94","msg":"building OpenTelemetry Go instrumentation ...","globalImpl":false}
{"level":"info","ts":1706790700.5980866,"logger":"Instrumentation.Analyzer","caller":"process/discover.go:69","msg":"found process","pid":19}

But after this I don't see any logs, even though I hit the api calls for my go services, I don't see any traces getting generated.
I am using below image.
image: ghcr.io/open-telemetry/opentelemetry-go-instrumentation/autoinstrumentation-go:v0.10.1-alpha
also I have set below env variables:

- name: OTEL_SERVICE_NAME
        value: target-service-name
      - name: OTEL_EXPORTER_OTLP_ENDPOINT
        value: http://my-collector-service:13575
      - name: OTEL_TRACES_EXPORTER
        value: otlp

### Environment

 - OS: LINUX
 - Go Version: 1.7.5
 - Version: v0.10.0-alpha
 more system info:
/ $ uname -r
5.10.205-195.804.amzn2.x86_64
/ $ uname -m
x86_64
/ $

damemi:
is your go version really 1.7.5? If so that might be the problem. The logging output hanging at `found process` means it's stuck looking for functions to instrument

jyoti-borade:
> is your go version really 1.7.5? If so that might be the problem. The logging output hanging at `found process` means it's stuck looking for functions to instrument

Yes it is 1.7 and 1.7.5. The go apps which I am trying to instrument is developed  long back. so can't we auto-instrument it as the go version is 1.7?

damemi:
@jyoti-borade that might be it. Could you please share the full output of `go version` so I can check?

The real bug is that whatever is causing the instrumentation to fail is not reporting an error. What you're seeing is actually the code hanging trying to report the error. I proposed a fix for this here: https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/638. You could try building with that PR to see what the exact error is you're getting

jyoti-borade:
> @jyoti-borade that might be it. Could you please share the full output of `go version` so I can check?
> 
[...]

PR #638, opened 2024-02-01
Don't call manager.Close() when Analyze() fails

damemi:
There's no need to call `manager.Close()` at this point because the manager is not running yet. Doing so actually causes it to hang because of waiting for [this channel](https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/e00e32a2b582f7a30f9f8b44cbd3db837815d3b2/internal/pkg/instrumentation/manager.go#L211) added in https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/586.

This also prevents errors from propagating up as seen in https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/636

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ OpenTelemetry Go Automatic Instrumentation adheres to [Semantic Versioning](http
 
 ### Fixed
 
-- Don't call manager.Close() when Analyze() fails. ([#638](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/638))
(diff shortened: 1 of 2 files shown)


RonFed (approved):

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ OpenTelemetry Go Automatic Instrumentation adheres to [Semantic Versioning](http
 
 ## [Unreleased]
 
+### Fixed
+
+- Don't call manager.Close() when Analyze() fails. ([#638](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/638))
```
```suggestion
- Don't call `manager.Close()` when `Analyze()` fails. ([#638](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/638))
```

MrAlias (approved):

Issue #340, opened 2023-09-21
Delete examples/rolldice


pellared:
> Do we need a separate issue for moving `go.opentelemetry.io/auto/examples/rolldice` under `excluded:` after the next release?

The plan is to delete `go.mod` in `go.opentelemetry.io/auto/examples/rolldice` after the next release.

_Originally posted by @MrAlias in https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/337#issuecomment-1726319109_

MrAlias:
Closed by #423

PR #650, opened 2024-02-07
Update examples to commit go.mod and gitignore binary

damemi:


pellared (approved):

pellared (approved):

.github/dependabot.yml
```diff
@@ -118,3 +136,30 @@ updates:
     schedule:
       interval: weekly
       day: sunday
+  - package-ecosystem: gomod
+    directory: /launcher
```
MrAlias, line 140: These don't seem right. Is the `launcher` directory local?

damemi: Ah yeah it is, but it's [.gitignore'd ](https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/2ac28454bfd38c4aff508c67bf22a66cbedaf1da/.gitignore#L27). I think that's from earlier code when we used a launcher. Will fix

examples/rolldice/go.mod
MrAlias, line 1: I'm a little hesitant about this. This will revert [this](https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/340) which was done to clean up our CI and module publishing workflows.

damemi: Ah, I missed that. In this case I think deleting the go.mod entirely might not have been necessary, like @pellared mentioned in https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/337#issuecomment-1726266842 we could list those modules as `excluded`.

Not committing the go.mod files makes it more of a pain to build. They might not always work, and when someone does build them it opens up the chance that they'll just get committed again. We could add them to `.gitignore` but I think that's avoiding the issue that examples are meant to be built. wdyt?

MrAlias (approved):

pellared:
Shouldn't we update `versions.yaml`?

damemi:
> Shouldn't we update `versions.yaml`?

Yep, I'll do that. I'm also going to add the other example.

@MrAlias and I talked about this in the sig call today and whether we should push it forward or not. It's not a huge deal, but I'm curious what other people think.

My stance is that the examples are a form of documentation, so including go.mod files helps make sure they still work.
It also makes life a little easier for developers.

On the other hand, it seems a little weird to define these modules if we're just going to exclude them from publishing. Maybe that's fine.

I know that it's a pain to bind ourselves to support constraints for code samples. But in that case, is it worth it to have them if we aren't going to maintain them?

examples/go.mod
```diff
@@ -1,13 +1,3 @@
 module go.opentelemetry.io/auto/examples
 
 go 1.20
-
-require (
```
damemi, line 5: not sure why everything in examples/go.mod and examples/go.sum got deleted

RonFed (approved):

pellared (approved):

damemi:
@MrAlias this is ready to merge

PR #654, opened 2024-02-09
Skip checking extra functions in FilterUnusedProbes()

damemi:
Once we find one of the symbols for a probe, we don't need to keep checking (we know we'll use that probe).


MrAlias (commented):
Shouldn't we check that `funcsFound == len(inst.Manifest().Symbols)`? The way it currently is it will panic if there is any missing, right?

RonFed:
> Shouldn't we check that `funcsFound == len(inst.Manifest().Symbols)`? The way it currently is it will panic if there is any missing, right?

It was changed after  https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/554

RonFed (approved):

damemi:
> > Shouldn't we check that `funcsFound == len(inst.Manifest().Symbols)`? The way it currently is it will panic if there is any missing, right?
> 
> It was changed after #554

is there a test for that panic?

RonFed:
> > > Shouldn't we check that `funcsFound == len(inst.Manifest().Symbols)`? The way it currently is it will panic if there is any missing, right?
> 
> > 
> 
> > It was changed after #554
> 
> 
> 
> is there a test for that panic?

Don't think so.

damemi:
Ok, let's hold this then and I'll see if it's feasible to add a test for that

MrAlias (approved):

MrAlias:
Plan is to track the error case testing in a follow-up issue.

PR #680, opened 2024-02-19
Avoid using runtime.NumCPU to get the number of CPUs on the system

RonFed:
`runtime.NumCPU` doc states:

> NumCPU returns the number of logical CPUs usable by the current process

This is problematic when the probed process is configured to run on a subset of the available CPUs.
In addition, we perform the above call for the agent process, so even if the probed process has some configuration for this, we won't see it.
This PR replaces this call with querying the relevant files to get the number of CPUs available in the system.
Adding a boundary check in `write_target_data` to ensure that we are not writing to an un-mapped address if something goes wrong.


edeNFed (approved):
lgtm
great catch!

PR #683, opened 2024-02-21
Add base Dockerfile for apt packages, and add caching for go build

RonFed:


Dockerfile.base
```diff
@@ -0,0 +1,3 @@
+FROM golang:1.22.0-bullseye
+
+RUN apt-get update && apt-get install -y curl clang gcc llvm make libbpf-dev
```
damemi, line 3: nit: newline at eof

damemi (commented):
@RonFed could you please add some details about what this is doing? I also have a couple questions:

* Are we going to publish the new base image?
* What are the reasons for building a separate `.base` image over just multi-stage docker builds?

I'm assuming this helps with faster build times and maybe multi-arch builds? Is that just for our CI/builds or does that help users too?

damemi:
Thanks for the explanation @RonFed!

> I thought the separate base Dockerfile could be helpful for some of the Makefile targets (avoiding the apt installs). Can this be achieved by adding it as a build-stage?

No, I think what you're doing here is right for that. I just wasn't sure what the main goal was so I wanted to pose that question.

damemi (approved):

RonFed:
@damemi I figured it is possible to do it with one Dokcerfile and use the base as the first stage, so I modified it back to a single file.
```
docker-build-base:
	docker buildx build -t $(IMG_NAME_BASE) --target base .
```

MrAlias (approved):

PR #695, opened 2024-02-26
Add bpf compiled files to git

RonFed:
Currently, using go auto instrumentation as a library requires the user to generate object files themselves. (calling `make generate` or something equivalent)., this makes the library not self-contained.
Adding the compiled files to git will solve this, and by just doing `go get go.opentelemetry.io/auto` the `Instrumentation` API will be available.

Having this workflow https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/42e7a5a72ca58ae621e1e886ce7801ab47c88836/.github/workflows/build.yaml#L12-L34

Will make sure the compiled files will stay in sync with the C files.
Currently, the above workflow is failing (the compiled files differ), I assume that modifying the workflow to use the base docker image proposed in #683 instead of calling `apt-get install` should make it pass.

Blocked by #683


damemi:
We excluded these intentionally in https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/16 when we only supported x86 and were thinking about how the generated files would be different on different architectures. It looks like that isn't a problem now based on the different file names, is that right? In other words, will the user's architecture always choose the right file and there is never a need for them to build it on their own?

I'm not a huge fan of committing binary files to git in general, because git has a hard time efficiently tracking the changes between them and it causes the repo size to grow pretty quick (not just because of the files themselves, but because of the git metadata for each commit).

damemi:
(btw I think this is worth a changelog entry)

RonFed:
> We excluded these intentionally in #16 when we only supported x86 and were thinking about how the generated files would be different on different architectures. It looks like that isn't a problem now based on the different file names, is that right? In other words, will the user's architecture always choose the right file and there is never a need for them to build it on their own?
> 
> I'm not a huge fan of committing binary files to git in general, because git has a hard time efficiently tracking the changes between them and it causes the repo size to grow pretty quick (not just because of the files themselves, but because of the git metadata for each commit).

Right. The file name suffixes add build constraints for the matching architecture.
From https://pkg.go.dev/cmd/go

>  If a file's name, after stripping the extension and a possible _test suffix, matches any of the following patterns:
> 
> *_GOOS
> *_GOARCH
> *_GOOS_GOARCH
> (example: source_windows_amd64.go) where GOOS and GOARCH represent any known operating system and architecture values respectively, then the file is considered to have an implicit build constraint requiring those terms (in addition to any explicit constraints in the file).

I agree that adding these files to Git is not ideal. I tried to think about another solution but couldn't think of something that would allow clients of this library to not generate these files themselves.

damemi:
We talked about this on the sig call yesterday with @MrAlias and @grcevski, and I don't think this is something we want to do (at least right now).

The extent of the problem and how many users it affects isn't totally clear. It seems like this only affects forks/consumers of the library code itself, of which there aren't that many at this time. They also have the option of committing the generated files to their own repos and updating their gitignores, so the only conflict with upstream should be the gitignore file.

@grcevski talked about some of the issues they've seen with committing bpf binaries, where it sounds like they cause a lot of merge conflicts that basically have to be force-merged and followed up with another commit to re-generate them. We would rather avoid that kind of workflow.

That said, it sounds like there is precedent in other eBPF projects for handling their binaries like this. But the use cases seem sparse, and for example in Beyla's case it was only for their own testing.

I think we were all in agreement that having *some* way to distribute these binaries is a good idea. But without user demand for it, we don't want to commit to this approach given the headaches it can cause.

Some of the other ideas we had were using Github releases to publish the binaries as artifacts. It could be possible to update our code, or provide tooling, to check for these artifacts and pull them at startup based on release tags or SHAs. We talked about some other projects that do stuff like that too.

But, that obviously comes with a lot of extra work. So we come back to the question of how much friction does this resolve? If it's enough of a pain point that it would be worth committing these files now, then it's probably also worth taking a step back to evaluate our approach. Otherwise, we can leave these out for now.

@RonFed do you think this is enough of an issue to push forward looking into alternative approaches, or is it okay to leave it as is?

RonFed:
@damemi That makes total sense. I'm OK with continuing with the current approach.

PR #705, opened 2024-03-04
Switch HTTP client to not use external memory for header injection

grcevski:
I'm making a draft pull request for an alternative implementation of header injection, which directly writes in the outgoing request buffer at the beginning when the request headers are serialized. At that point, the outgoing buffer is 4K in size, while typically only handful of bytes are used to write the request method and URL.

This approach has a limitation that the injection will not work if the URL (and the method) are > 4027 bytes and <= 4096 bytes. URLs larger than 2048 are unsupported by various programs, but technically they can be larger. It's possible to overcome this limitation with more work if it's considered a blocker.

TODO:

- [x] Remove the debug statements
- [x] Add tests for the manifests changes


edeNFed:
Nice chnage, I liked the fallback using `-DNO_HEADER_PROPAGATION`.

A couple of questions:
- I am concerned about the performance impact of adding additional uprobe
for the `writeSubset()` function. Do you know how often` writeSubset()` is called per request?
- I wonder what are the advantages of using an additional uprobe over the existing mechanism. This is very specific to the HTTP instrumentation and in general, I am not sure we will be able to find such a function for every instrumentation in the future.
- The current implementation assumes `io.Writer` is implemented by `bufio.Writer`, I think this function can also be called by users with different `io.Writer` implementations by invoking `Header.WriteSubset()` (the exported function). This may cause the application to crash due to writing to the wrong memory address. A possible solution may be to look at the `itab` struct to verify the concrete type of the `io.Writer.

grcevski:
> Nice chnage, I liked the fallback using `-DNO_HEADER_PROPAGATION`.
> 
> A couple of questions:
> 
> * I am concerned about the performance impact of adding additional uprobe
>   for the `writeSubset()` function. Do you know how often` writeSubset()` is called per request?

So it's called once per client request, but it does get called also when headers are sent in a response of a server request. There will be one extra probe that will run, but we'll end very quickly since no header will be found, so it's mostly the probe overhead, around 100ns. We can time it with this new tool by Netflix `bpftop` vs the old instrumentation to get some actual overhead data.

> * I wonder what are the advantages of using an additional uprobe over the existing mechanism. This is very specific to the HTTP instrumentation and in general, I am not sure we will be able to find such a function for every instrumentation in the future.

It's mostly around memory safety, since we'll actually use only memory that the Go garbage collector manages. No rewriting of map spines needs to be done with non-managed memory, which theoretically can cause the GC to temporarily lose reference to the underlying map key values. I've found a way how to do the same for gRPC outgoing calls and HTTP2. My current gRPC implementation is inefficient, but I think I can do the same as for HTTP2 and it will be similar to how we handle HTTP.

If we are able to fully work without external memory segments, other things become simpler. OTel Go Auto will be able to run without `privileged: true`, I expect CAP_SYS_ADMIN to be sufficient, or maybe even CAP_BPF. Implementing multi-process instrumentation also becomes a lot simpler and leaner in terms of memory consumption.

> * The current implementation assumes `io.Writer` is implemented by `bufio.Writer`, I think this function can also be called by users with different `io.Writer` implementations by invoking `Header.WriteSubset()` (the exported function). This may cause the application to crash due to writing to the wrong memory address. A possible solution may be to look at the `itab` struct to verify the concrete type of the `io.Writer.

I believe we are safe from that scenario, because the code that assumes `bufio.Writer` is guarded by the check that verifies we have recorded the header in the RoundTrip start, and we also verify it's actually an active http client request. I think, we will only assume `bufio` when it's actually a `bufio`, but maybe I'm missing an edge case when during RoundTrip this can be another io.Writer type.

To validate `bufio` I looked at the caller of `Header.writeSubset` in the client HTTP scenario, which is 

```
func (r *Request) write(w io.Writer, usingProxy bool, extraHeaders Header, waitForContinue func() bool) (err error) {
```
In there, they explicitly make a bufio if it wasn't, before the `writeSubset` function is called:
```
	// Wrap the writer in a bufio Writer if it's not already buffered.
	// Don't always call NewWriter, as that forces a bytes.Buffer
	// and other small bufio Writers to have a minimum 4k buffer
	// size.
	var bw *bufio.Writer
	if _, ok := w.(io.ByteWriter); !ok {
		bw = bufio.NewWriter(w)
		w = bw
	}
```

I believe we are safe, but I can add extra code to confirm the type if you think it's still a problem.

edeNFed:
OK, got it on the performance overhead and `io.Writer` implementation. This makes sense I don't think we need to add an additional type check then.

Regarding memory safety, I think this is something we need to consider on a per-instrumentation basis. 
I do see the point in reducing the required permissions, however, I think that finding a pre-allocated space will not work for every library. We (Keyval) have many instrumentations that we are planning to gradually merge and I think that finding that specific buffer on every library is going to be extremely hard, inefficient, and probably impossible for some libraries.

This will tie us to mostly having gRPC/HTTP instrumentations as opposed to achieving full coverage (like [otel-go-contrib](https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/instrumentation))

grcevski:
> Regarding memory safety, I think this is something we need to consider on a per-instrumentation basis.
> I do see the point in reducing the required permissions, however, I think that finding a pre-allocated space will not work for every library. We (Keyval) have many instrumentations that we are planning to gradually merge and I think that finding that specific buffer on every library is going to be extremely hard, inefficient, and probably impossible for some libraries.

Agreed. As discussed at the SIG, let's try to implement the instrumentations we can without the external memory segment. If the application is only using instrumentations that don't require an external memory segment, we can avoid the ptrace call and the shared memory segment injection. We won't remove the functionality, we'll tag which probes require it and after filtering we can determine if injecting the foreign memory segment is required or not.

internal/pkg/instrumentation/bpf/net/http/client/bpf/probe.bpf.c
```diff
@@ -267,4 +174,54 @@ int uprobe_Transport_roundTrip_Returns(struct pt_regs *ctx) {
...
+            if (len < (size - W3C_VAL_LENGTH - W3C_KEY_LENGTH - 4)) { // 4 = strlen(":_") + strlen("\r\n")
+                char key[W3C_KEY_LENGTH + 2] = "Traceparent: ";
+                char end[2] = "\r\n";
+                bpf_probe_write_user(buf_ptr + (len & 0x0ffff), key, sizeof(key));
+                len += W3C_KEY_LENGTH + 2;
+                bpf_probe_write_user(buf_ptr + (len & 0x0ffff), tp, sizeof(tp));
+                len += W3C_VAL_LENGTH;
+                bpf_probe_write_user(buf_ptr + (len & 0x0ffff), end, sizeof(end));
+                len += 2;
+                bpf_probe_write_user((void *)(io_writer_ptr + io_writer_n_pos), &len, sizeof(len));
```
RonFed, line 218: Should these 4 write operation be "atomic"? i.e. if one of them fails, we might create an invalid header?

internal/pkg/instrumentation/bpf/net/http/client/bpf/probe.bpf.c
```diff
@@ -267,4 +174,54 @@ int uprobe_Transport_roundTrip_Returns(struct pt_regs *ctx) {
...
+                goto done;
+            }
+
+            s64 size = 0;
+            bpf_probe_read(&size, sizeof(s64), (void *)(io_writer_ptr + io_writer_buf_ptr_pos + 8)); // grab size
+
+            s64 len = 0;
+            bpf_probe_read(&len, sizeof(s64), (void *)(io_writer_ptr + io_writer_n_pos)); // grab len
+
+#ifndef NO_HEADER_PROPAGATION
```
RonFed, line 208: Maybe we should put this entire function under `#ifnded`?

internal/pkg/instrumentation/bpf/net/http/client/probe.go
```diff
@@ -91,12 +92,25 @@ func New(logger logr.Logger) probe.Probe {
...
 		},
 		Uprobes: []probe.Uprobe[bpfObjects]{
 			{
 				Sym: "net/http.(*Transport).roundTrip",
 				Fn:  uprobeRoundTrip,
 			},
+			{
+				Sym:       "net/http.Header.writeSubset",
+				Fn:        uprobeWriteSubset,
+				DependsOn: []string{"net/http.(*Transport).roundTrip"},
```
RonFed, line 112: What is the goal of `DependsOn`?
Maybe we shouldn't load this uprobe if the kernel is in lockdown?

internal/pkg/instrumentation/manager.go
```diff
@@ -103,8 +103,17 @@ func (m *Manager) FilterUnusedProbes(target *process.TargetDetails) {
...
+				for _, depends := range s.DependsOn {
+					if _, exists := existingFuncMap[depends]; exists {
+						funcsFound++
+						break
+					}
+				}
+			} else {
+				if _, exists := existingFuncMap[s.Symbol]; exists {
+					funcsFound++
+				}
```
RonFed, line 116: This PR https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/654/files is also changing this part. 
I didn't fully understand `DepandsOn` so I'm not sure what is the best approach here.

damemi: Yeah this section doesn't seem to need to integer count either. @MrAlias can you please merge #654?

internal/pkg/instrumentation/bpf/net/http/client/bpf/probe.bpf.c
grcevski: Done thanks for the suggestion!

internal/pkg/instrumentation/bpf/net/http/client/bpf/probe.bpf.c
grcevski: Done too. I had to leave an empty stub of a probe function, so that bpf2go can compile the two versions. However I added user space code that prevents the loading of the probe.

internal/pkg/instrumentation/bpf/net/http/client/probe.go
grcevski: I added a comment to clarify what this is about, but it's like @damemi  said. The writeSubset may appear in the binary of the program without roundTrip. That will force us to load the client probe and fail because roundTrip isn't optional, but also we'd be loading a bpf probe for no reason, where the dependent functionality isn't there. With DependsOn, we can specify that a uprobe should not be counted for activating the bpf program, unless the dependent symbol is found too.

grcevski: > Maybe we shouldn't load this uprobe if the kernel is in lockdown?

We discussed this at the SIG, there's some value of generating traces when we can't context propagate, like spanmetrics generation. Also customers might be confused why instrumentation doesn't work, a log message is easy to miss.

internal/pkg/instrumentation/manager.go
grcevski: Since we use bool now, the code is much simpler, we only need to count the probes that don't have dependencies.

internal/pkg/instrumentation/bpf/net/http/client/bpf/probe.bpf.c
RonFed: Actually, I meant that if it is possible to replace these 4 write operations with a single `bpf_probe_write_user` it will be better.

internal/pkg/instrumentation/manager.go
```diff
@@ -103,9 +103,11 @@ func (m *Manager) FilterUnusedProbes(target *process.TargetDetails) {
 	for name, inst := range m.probes {
 		funcsFound := false
 		for _, s := range inst.Manifest().Symbols {
-			if _, exists := existingFuncMap[s]; exists {
-				funcsFound = true
-				break
+			if len(s.DependsOn) == 0 {
```
RonFed, line 106: Shouldn't we look at the contents of `s.DependsOn` to make sure the required function is present? maybe I missed that

internal/pkg/instrumentation/bpf/net/http/client/bpf/probe.bpf.c
grcevski: Sorry, I misunderstood. I've changed the code now to use a single char buffer which combines the the 'traceparent' key and value, and I call a single bpf_probe_write_user to store the full buffer. Updating the 'N' (the new length of the buffer) must be done through separate write, however if we fail the first write it won't affect the second. And if the second fails (updating the n), we simply wouldn't get a traceparent in the buffer.

internal/pkg/instrumentation/manager.go
grcevski: I don't think we need to. If DependsOn has something, it means that the probe is not valid as a standalone and there's one or more other probes that drive its functionality. So to activate this bpf program, there must be at least one found standalone probe which doesn't depend on anything else.

I added a test for this at line 85 in the manager_test.go. 
```
t.Run("HTTP server and client dependent function only target details"
```

Technically the field can be a boolean, but after a discussion a few weeks back at the SIG it was suggested that I make this field an array.

Perhaps one thing I can do is ensure that the dependents listed in DependsOn exist, so that we don't accidentally break functionality.

grcevski: Added dependent validation and tests.

internal/pkg/instrumentation/bpf/net/http/client/bpf/probe.bpf.c
```diff
@@ -265,4 +172,73 @@ int uprobe_Transport_roundTrip_Returns(struct pt_regs *ctx) {
 
     bpf_map_delete_elem(&http_events, &key);
     return 0;
-}
\ No newline at end of file
+}
+
+#ifndef NO_HEADER_PROPAGATION
+SEC("uprobe/header_writeSubset")
```
RonFed, line 178: nit: add the function signature as a comment (similar to other probes)

RonFed (approved):
Added two small comments

MrAlias (approved):

internal/pkg/instrumentation/bpf/net/http/client/bpf/probe.bpf.c
grcevski: Done, thanks!

internal/pkg/instrumentation/bpf/net/http/client/bpf/probe.bpf.c
```diff
@@ -265,4 +172,73 @@ int uprobe_Transport_roundTrip_Returns(struct pt_regs *ctx) {
...
+
+            void *buf_ptr = 0;
+            bpf_probe_read(&buf_ptr, sizeof(buf_ptr), (void *)(io_writer_ptr + io_writer_buf_ptr_pos)); // grab buf ptr
+            if (!buf_ptr) {
+                bpf_printk("uprobe_writeSubset: Failed to get buf from io writer");
+                goto done;
+            }
+
+            s64 size = 0;
+            if (bpf_probe_read(&size, sizeof(s64), (void *)(io_writer_ptr + io_writer_buf_ptr_pos + 8))) { // grab size
```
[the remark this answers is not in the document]
grcevski: Good point! In this case the length and the capacity are the same, but it's better if we read the capacity.

grcevski:
Thanks so much @edeNFed, @RonFed, @MrAlias and @damemi!

Issue #730, opened 2024-03-21
Ctrl-C signal not handled while searching for the target PID


rogercoll:
### Describe the bug

 Ctrl-C signal is not properly handled if the process PID is not found. The Analyzer loop does not use the global context with the signal handler.

### Environment

 - OS: Linux
 - Go Version: 1.22.1
 - Version: v0.10.0

### To Reproduce

Launch the binary with an invalid target path or a not launched process. Try to stop the execution by running ctrl-c.

### Expected behavior

Finish the instrumentation process.

PR #731, opened 2024-03-21
fix: propagate global context to process analyzer

rogercoll:
- Propagate context to the DiscoverProcessID
- Remove the inner "done" Analyzer channel as it is not used. The `DiscoverProcessID` funcion is not launched as a Go routine, thus, we don't need a "Close" method.

Fixes https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/730


RonFed (approved):
lgtm

damemi (approved):

MrAlias (approved):

PR #737, opened 2024-03-26
Prerelease v0.11.0-alpha

damemi:
### Added

- Test build using [Go 1.22]. (#672)
- Base Dockerfile and build caching ([#683](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/683))
- Add `server.address`, `server.port` and `network.protocol.version` to HTTP client spans ([#696](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/696))
- Update http server attributes to latest semantic conventions ([#708](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/708))
### Fixed
- Don't call `manager.Close()` when `Analyze()` fails. ([#638](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/638))
- Close `proc` file when done discovering PID. ([#649](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/649))
- Use `debug` packages to parse Go and modules' versions. ([#653](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/653))
- Clean up warn in otelglobal `SetStatus()` when grabbing the status code. ([#675](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/675))
- Reset `proc` offset after a failed iteration. ([#681](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/681))
- Avoid using runtime.NumCPU to get the number of CPUs on the system before remote mmap ([#680](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/680))
- Cleanup eBPF maps only when we stop using the memory ([#682](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/682))

[Go 1.22]: https://go.dev/doc/go1.22

Issue #93, opened 2023-04-24
Refine e2e tests


pellared:
Based on my experience in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation (where I am a maintainer) I find testing against a "test fake collector" makes the tests more maintainable:

- spinning up a a `httptest` server or by setting up a gRPC server in tests is faster than running a real collector [image]
- you get more flexibility in assertion logic

Here you can find an example implementation and usage of a "test fake collector" in Go: https://github.com/signalfx/splunk-otel-go/blob/main/distro/otel_test.go

_Originally posted by @pellared in https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/46#pullrequestreview-1359674606_

CC @damemi

pellared:
Removing from Beta project as I do not think it is a blocker.

damemi:
From https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/738:
> > The bug in this PR can happen approximately once every 50-100 client spans (since page size is 4096 bytes and we write ~40 bytes per request in gRPC for example)
> 
> 1-2% of the time isn't insignificant, but I don't think any of our tests are set up for that kind of scenario. Maybe as part of #93 we should update the tests to send more than one span (ideally enough to cross the page threshold).

damemi:
From the sig call:

As we shift the primary focus of this repo to a "library" rather than a "component", the need for us to have full-featured e2e environments is less important (k8s, collector, etc). Instead, consumers of the library like the Beyla donation and Odigos will do their component-level testing.

So, we could take that to strip away a lot of the overhead in our tests. We also discussed the option of using a basic custom server like what you mentioned @pellared 

cc @MrAlias

PR #738, opened 2024-03-26
Fix start offset calculation in mmaped memory area

RonFed:
This PR fixes a relatively rare bug that can happen when a write operation to the mmaped area is moving between pages. We didn't take the page-aligned address into account when updating the start offset.

### DIFF
--- a/internal/include/alloc.h
+++ b/internal/include/alloc.h
@@ -118,7 +118,9 @@ static __always_inline void *write_target_data(void *data, s32 size)
     if (success == 0)
     {
         s32 start_index = 0;
-        u64 updated_start = start + size;
+        // Update the start position of this chunk, taking into account possible adjustments
+        // we made to be page aligned
+        u64 updated_start = target_u + size;
 
         // align updated_start to 8 bytes
         if (updated_start % 8 != 0) {


edeNFed (approved):

MrAlias (approved):

MrAlias:
Are there any tests we can add to prevent a regression?

RonFed:
> Are there any tests we can add to prevent a regression?

I think the e2e tests for grpc and http are covering the case of regression espicialy this part which checks the client-server propagation: https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/f18c1b2e0702d8ac31699c5e923590d714d0c1dc/internal/test/e2e/grpc/verify.bats#L48-L52

The bug in this PR can happen approximately once every 50-100 client spans (since page size is 4096 bytes and we write ~40 bytes per request in gRPC for example)

damemi:
> The bug in this PR can happen approximately once every 50-100 client spans (since page size is 4096 bytes and we write ~40 bytes per request in gRPC for example)

1-2% of the time isn't insignificant, but I don't think any of our tests are set up for that kind of scenario. Maybe as part of https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/93 we should update the tests to send more than one span (ideally enough to cross the page threshold).

@RonFed just so I understand does this bug cause spans to get dropped by being overwritten in the map when the new address isn't taken into account?

damemi (approved):

RonFed:
> > The bug in this PR can happen approximately once every 50-100 client spans (since page size is 4096 bytes and we write ~40 bytes per request in gRPC for example)
> 
> 
> 
> 1-2% of the time isn't insignificant, but I don't think any of our tests are set up for that kind of scenario. Maybe as part of https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/93 we should update the tests to send more than one span (ideally enough to cross the page threshold).
> 
> 
> 
> @RonFed just so I understand does this bug cause spans to get dropped by being overwritten in the map when the new address isn't taken into account?
> 
> 
> 
> 

Right, it can cause spans to get dropped. Additionally in some case it caused invalid characters to override the trace parent header which resulted in a RST_STREAM code in the underlying http2 (in grpc).

MrAlias:
@damemi thank you for updating #93. I'm going to merge this, and have the follow-up work tracked there.

damemi:
btw @RonFed I missed it but I think this should have had a changelog entry. I'll add one in https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/737 for this. 

Maybe we should define what needs a changelog entry somewhere?

RonFed:
> btw @RonFed I missed it but I think this should have had a changelog entry. I'll add one in https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/737 for this. 
> 
> 
> 
> Maybe we should define what needs a changelog entry somewhere?

Ah thanks 🙂, yeah I don't really know when to add a changelog entry or not.

Issue #627, opened 2024-01-25
Improve cardinality of server spans for better segmentation


SophieDeBenedetto:
**Is your feature request related to a problem? Please describe.**
GitHub has been evaluating running the ebpf sidecar container provided by opentelemetry-go-instrumentation as a solution for auto-instrumenting our Golang services. After deploying our POC, we ultimately found the server spans provided by the opentelemetry-go-instrumentation probes to be too generic for us to find useful. We would also love to see more probes developed for additional Golang dependencies beyond what's present [here](https://github.com/open-telemetry/opentelemetry-go-instrumentation/tree/main/internal/pkg/instrumentation/bpf). We're hoping to eventually engage GitHub engineers in contributing to this project, so we're starting off with this issue to describe the work we feel would benefit our use-case.

I'll provide a little more detail here on what I mean by "generic" spans. 

* DB server spans will all be named `"DB"`, [ref](https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/main/internal/pkg/instrumentation/bpf/database/sql/probe.go#L174):

```golang
// pkg/instrumentation/bpf/database/sql/probe.go#L174
return &probe.SpanEvent{
		SpanName:    "DB",
		StartTime:   int64(e.StartTime),
		EndTime:     int64(e.EndTime),
		SpanContext: &sc,
		Attributes: []attribute.KeyValue{
			semconv.DBStatementKey.String(query),
		},
		ParentSpanContext: pscPtr,
	}
```

We'd benefit from support for templatized parameters here to allow span names to be something like `DB <operation`, to match the `db.operation` [OTel semconv span name](https://opentelemetry.io/docs/specs/semconv/database/sql/). This allows us to at least segment trace data and RED metrics derived from traces by database operation at a resource level.

Similarly, HTTP server spans are named with the HTTP method only, [ref](https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/main/internal/pkg/instrumentation/bpf/net/http/server/probe.go#L204):

```golang
// pkg/instrumentation/bpf/net/http/server/probe.go#L204
	return &probe.SpanEvent{
		// Do not include the high-cardinality path here (there is no
		// templatized path manifest to reference).
		SpanName:          method,
		StartTime:         int64(e.StartTime),
		EndTime:           int64(e.EndTime),
		SpanContext:       &sc,
		ParentSpanContext: pscPtr,
		Attributes:        attributes,
	}
```

[...]

Issue #26, opened 2023-02-07
Document instrumentation versioning compatibility


MrAlias:
For each support instrumented library document what version of the library are supported.

damemi:
https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/725#issuecomment-2032609505 for other support compatibility requirements

MrAlias:
Automation of this documentation is tracked in https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/908.

PR #740, opened 2024-03-27
Support template paths as part of span names for go versions >= 1.22

RonFed:
In `go v1.22` the `net/http` package added support for pattern handlers https://go.dev/blog/routing-enhancements.
This allows us to add a low cardinality path for span names in http server.
There were a few discussions about this topic (one of them in https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/627),  and I think this solution allows for a much more informative span name while keeping it with low cardinality.

* Change the span name to `{method} {http.route}` when a route is available.
* Added the http.route attribute when the path template is available.
* Update the `net/http` e2e test to use a path template.


damemi (approved):
this LGTM, just want to double check this line from semconv:

> Instrumentation MUST NOT default to using URI path as span name, but MAY provide hooks to allow custom logic to override the default span name.

I think what this is doing would count as "providing hooks" (using the option consts) rather than defaulting to using the path. But, our instrumentation agent will be using that hook by default. @MrAlias does this seem right or does that line mean this should be behind some non-default option for the agent?

RonFed:
@damemi I think the line you quoted means that we should provide an opt-in option to include the high cardinality path in the span name, but I'm not sure. This PR changes the default span name to include the low cardinality path template if available.

internal/pkg/inject/offset_results.json
```diff
@@ -2750,6 +2750,168 @@
...
+                      "1.19.11",
+                      "1.19.12",
+                      "1.19.13",
+                      "1.20.0",
+                      "1.20.1",
+                      "1.20.2",
+                      "1.20.3",
+                      "1.20.4",
+                      "1.20.5",
+                      "1.20.6",
```
MrAlias, line 2887: This seems broken: https://cs.opensource.google/go/go/+/refs/tags/go1.20.6:src/net/http/request.go;l=106

internal/pkg/inject/offset_results.json
```diff
@@ -2915,6 +3077,168 @@
               }
             ]
           },
+          {
+            "struct": "pattern",
```
MrAlias, line 3081: Similar here. This wasn't added until 1.22.

MrAlias:
> this LGTM, just want to double check this line from semconv:
> 
> > Instrumentation MUST NOT default to using URI path as span name, but MAY provide hooks to allow custom logic to override the default span name.
> 
> I think what this is doing would count as "providing hooks" (using the option consts) rather than defaulting to using the path. But, our instrumentation agent will be using that hook by default. @MrAlias does this seem right or does that line mean this should be behind some non-default option for the agent?

This seems correct to me given it will only ever use a templatized path. The hooks are there when a templitized version of this is not available in the language, but is something the user could provide.

MrAlias:
For those also interested to see how this handles non-exact matches: https://go.dev/play/p/9ZZz4BjTQ6C

damemi:
> > this LGTM, just want to double check this line from semconv:
> > > Instrumentation MUST NOT default to using URI path as span name, but MAY provide hooks to allow custom logic to override the default span name.
> > 
> > 
> > I think what this is doing would count as "providing hooks" (using the option consts) rather than defaulting to using the path. But, our instrumentation agent will be using that hook by default. @MrAlias does this seem right or does that line mean this should be behind some non-default option for the agent?
> 
> This seems correct to me given it will only ever use a templatized path. The hooks are there when a templitized version of this is not available in the language, but is something the user could provide.

Ah so `URI Path` in that section of semconv is referring to the full non-templatized path. Makes sense thanks 👍

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ OpenTelemetry Go Automatic Instrumentation adheres to [Semantic Versioning](http
 
 ## [Unreleased]
 
+### Added
+
+- Support http server path templated from go version 1.22 (`http.route` attribute and modified span name) ([#740](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/740))
```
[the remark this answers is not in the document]
RonFed: Fixed

internal/pkg/instrumentation/bpf/net/http/http_event.go
```diff
@@ -61,3 +62,39 @@ func NetPeerAddressPortAttributes(host []byte) (addr attribute.KeyValue, port at
...
+	// ErrMissingPathOrHost is returned when the input pattern is missing path or host.
+	ErrMissingPathOrHost = errors.New("missing path or host")
+)
+
+// The string's syntax is
+//
+//	[METHOD] [HOST]/[PATH]
+//
+// https://cs.opensource.google/go/go/+/master:src/net/http/pattern.go;l=84;drc=b47f2febea5c570fef4a5c27a46473f511fbdaa3?q=PATTERN%20STRUCT&ss=go%2Fgo
+func ParsePattern(s string) (method, host, path string, err error) {
```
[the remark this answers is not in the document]
RonFed: Removed

internal/pkg/inject/offset_results.json
RonFed: I noticed that if I rebuild the `offsetgen` tool and run it for the first time it gets it right, and the second time it puts all these wrong values, so I'm guessing there is some bug in the cache mechanism there.

internal/pkg/instrumentation/bpf/net/http/http_event.go
```diff
@@ -61,3 +62,39 @@ func NetPeerAddressPortAttributes(host []byte) (addr attribute.KeyValue, port at
...
+	ErrEmptyPattern = errors.New("empty pattern")
+	// ErrMissingPathOrHost is returned when the input pattern is missing path or host.
+	ErrMissingPathOrHost = errors.New("missing path or host")
+)
+
+// The string's syntax is
+//
+//	[METHOD] [HOST]/[PATH]
+//
+// https://cs.opensource.google/go/go/+/master:src/net/http/pattern.go;l=84;drc=b47f2febea5c570fef4a5c27a46473f511fbdaa3?q=PATTERN%20STRUCT&ss=go%2Fgo
```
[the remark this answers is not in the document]
RonFed: Added

RonFed:
@MrAlias Can you please have another look?

MrAlias:
> @MrAlias Can you please have another look?

We talked about this PR in the SIG meeting today. I'm hesitant to merge this without the offsets generator being fixed.

If I understand this correctly, if a user is using a version of Go prior to 1.22 the offsets will be "found" and the instrumentation loaded, but that offset is invalid an the system would fail to handle it.

I would not want to include this in the instrumentation without first having correct offsets.

RonFed:
> > @MrAlias Can you please have another look?
> 
> We talked about this PR in the SIG meeting today. I'm hesitant to merge this without the offsets generator being fixed.
> 
> If I understand this correctly, if a user is using a version of Go prior to 1.22 the offsets will be "found" and the instrumentation loaded, but that offset is invalid an the system would fail to handle it.
> 
> I would not want to include this in the instrumentation without first having correct offsets.

The following part should make sure that if we instrument an app with go version lower than 1.22 the pattern parts are not taken into account (we won't consider the offsets in eBPF if we are not detecting a valid version), hence I think it should not necessarily block this PR. 

```
type patternPathSupportedConst struct{}

var (
	patternPathMinVersion  = version.Must(version.NewVersion("1.22.0"))
	isPatternPathSupported = false
)

func (c patternPathSupportedConst) InjectOption(td *process.TargetDetails) (inject.Option, error) {
	isPatternPathSupported = td.GoVersion.GreaterThanOrEqual(patternPathMinVersion)
	return inject.WithKeyValue("pattern_path_supported", isPatternPathSupported), nil
}
```

MrAlias:
Ah! I couldn't find this. Guess closing the other PR closed this. Thanks for reopening. I'll try to review this next.

internal/pkg/instrumentation/bpf/net/http/http_event.go
```diff
@@ -61,3 +62,38 @@ func NetPeerAddressPortAttributes(host []byte) (addr attribute.KeyValue, port at
...
+// Copyright 2023 The Go Authors. All rights reserved.
```
MrAlias, line 78: This needs to be the complete copyright notice and it needs to go at the top of this file. For example: https://github.com/open-telemetry/opentelemetry-go/blob/24d36d6931bea3391f7fccf81bee79c4e104dcd2/sdk/log/ring.go#L4-L6

MrAlias:
> Not sure why the offsets action is failing with ` "msg"="failed get offsets" "error"="Error response from daemon: No such image: golang:1.14.1"`

That is strange. The image does appear to still exist: https://hub.docker.com/layers/library/golang/1.14.1/images/sha256-23405c683ba88dffcb1493c66305533bb5b3cef179e419e2e5c2f5cb92a06891?context=explore

MrAlias:
This seems unique to this PR. The auto-offset-gen action still works fine: https://github.com/open-telemetry/opentelemetry-go-instrumentation/actions/runs/8850141574/job/24303716999

internal/pkg/instrumentation/bpf/net/http/http_event.go
RonFed: fixed

MrAlias:
I tried reproducing the failure locally, but was unsuccessful. I even deleted the local `golang:1.14.1` image but it still succeeded. This looks unique to the action.

I'm wondering if there is some caching we can clear.

MrAlias:
I'm able to reproduce in my fork: https://github.com/MrAlias/opentelemetry-go-instrumentation/actions/runs/8851008583/job/24306583275?pr=237

MrAlias:
Yikes. Looks like there is no space left on the device(?)

> 2024/04/26 16:11:56 inspector/builder: "level"=1 "msg"="pulling image" "image"="golang:1.14.1" "output"="{\"status\":\"Pulling from library/golang\",\"id\":\"1.14.1\"}\r\n{\"status\":\"Already exists\",\"progressDetail\":{},\"id\":\"f15005b0235f\"}\r\n{\"status\":\"Already exists\",\"progressDetail\":{},\"id\":\"41ebfd3d2fd0\"}\r\n{\"status\":\"Already exists\",\"progressDetail\":{},\"id\":\"b998346ba308\"}\r\n{\"status\":\"Already exists\",\"progressDetail\":{},\"id\":\"f01ec562c947\"}\r\n{\"status\":\"Already exists\",\"progressDetail\":{},\"id\":\"c5e092b8639b\"}\r\n{\"status\":\"Pulling fs layer\",\"progressDetail\":{},\"id\":\"6ee82bba3116\"}\r\n{\"status\":\"Pulling fs layer\",\"progressDetail\":{},\"id\":\"3b36b05a8864\"}\r\n{\"status\":\"Downloading\",\"progressDetail\":{\"current\":126,\"total\":126},\"progress\":\"[==================================================\\u003e]     126B/126B\",\"id\":\"3b36b05a8864\"}\r\n{\"status\":\"Verifying Checksum\",\"progressDetail\":{},\"id\":\"3b36b05a8864\"}\r\n{\"status\":\"Download complete\",\"progressDetail\":{},\"id\":\"3b36b05a8864\"}\r\n{\"status\":\"Downloading\",\"progressDetail\":{\"current\":531698,\"total\":123632532},\"progress\":\"[\\u003e                                                  ]  531.7kB/123.6MB\",\"id\":\"6ee82bba3116\"}\r\n{\"status\":\"Downloading\",\"progressDetail\":{\"current\":10180900,\"total\":123632532},\"progress\":\"[====\\u003e                                              ]  10.18MB/123.6MB\",\"id\":\"6ee82bba3116\"}\r\n{\"status\":\"Downloading\",\"progressDetail\":{\"current\":123632532,\"total\":123632532},\"progress\":\"[==================================================\\u003e]  123.6MB/123.6MB\",\"id\":\"6ee82bba3116\"}\r\n{\"errorDetail\":{\"message\":\"write /var/lib/docker/tmp/GetImageBlob3791904830: no space left on device\"},\"error\":\"write /var/lib/docker/tmp/GetImageBlob3791904830: no space left on device\"}\r\n"


https://github.com/MrAlias/opentelemetry-go-instrumentation/actions/runs/8851119331/job/24306956117?pr=237

MrAlias:
Guessing this comes down to the fact that we are having to check all the old versions of Go for the new symbols added here. Prior to this we used to just hit the cache for all the symbols and never pulled the Go images.

MrAlias:
> Guessing this comes down to the fact that we are having to check all the old versions of Go for the new symbols added here. Prior to this we used to just hit the cache for all the symbols and never pulled the Go images.

Guessing we're going to need to add a "hint" to the offset manifests about what minimum version the symbol was added in.

MrAlias:
> > Guessing this comes down to the fact that we are having to check all the old versions of Go for the new symbols added here. Prior to this we used to just hit the cache for all the symbols and never pulled the Go images.
> 
> Guessing we're going to need to add a "hint" to the offset manifests about what minimum version the symbol was added in.

#799 

Going to try and resolve this later today.

MrAlias:
@RonFed https://github.com/RonFed/opentelemetry-go-instrumentation_fork/pull/1

That looks to resolve the "out of space" issue: https://github.com/MrAlias/opentelemetry-go-instrumentation/actions/runs/8852585249/job/24311585805?pr=237

RonFed:
> Guessing this comes down to the fact that we are having to check all the old versions of Go for the new symbols added here. Prior to this we used to just hit the cache for all the symbols and never pulled the Go images.

Yeah, that makes sense.

> @RonFed [RonFed#1](https://github.com/RonFed/opentelemetry-go-instrumentation_fork/pull/1)
> 
> That looks to resolve the "out of space" issue: https://github.com/MrAlias/opentelemetry-go-instrumentation/actions/runs/8852585249/job/24311585805?pr=237

Do you think we should use this as a quick fix or add some generic logic to the cache loading?

For general solution I think that if for a given module the minimal version is `vModMin` and some field in that module has cache entries starting from version `vFieldMin` and `vFieldMin` > `vModMin` we won't consider the versions smaller than `vFieldMin`

MrAlias:
> For general solution I think that if for a given module the minimal version is `vModMin` and some field in that module has cache entries starting from version `vFieldMin` and `vFieldMin` > `vModMin` we won't consider the versions smaller than `vFieldMin`

This would mean that patch releases to prior minor/major versions would not be evaluated. I don't think that is the correct approach.

RonFed:
> > For general solution I think that if for a given module the minimal version is `vModMin` and some field in that module has cache entries starting from version `vFieldMin` and `vFieldMin` > `vModMin` we won't consider the versions smaller than `vFieldMin`
> 
> This would mean that patch releases to prior minor/major versions would not be evaluated. I don't think that is the correct approach.

Right. Another idea is to add "null" offsets to the cache for versions where there is no valid offset. It makes sense to cache the evaluation of no valid offset in a given version, but it will make the offset file much bigger.

MrAlias:
> Right. Another idea is to add "null" offsets to the cache for versions where there is no valid offset. It makes sense to cache the evaluation of no valid offset in a given version, but it will make the offset file much bigger.

That sounds reasonable to me. It would be a bit harder for a human to read, but that's not critical if we can solve #26

MrAlias (approved):

PR #742, opened 2024-04-01
Pass pid to UprobeOptions

RonFed:
This PR passes the PID to `UprobeOptions`.
The scenario where this makes a difference is when multiple processes use the same executable, we want to make sure we only probe the relevant process.


damemi:
Thanks @RonFed, do you think it would be possible to add a pid fork in the e2e samples to test this?

also sorry to nag about tests... our infra is lacking and it should be a lot easier than it currently is to test changes like this

RonFed:
> Thanks @RonFed, do you think it would be possible to add a pid fork in the e2e samples to test this?
> 
> also sorry to nag about tests... our infra is lacking and it should be a lot easier than it currently is to test changes like this

I'm in favor of improving our tests, however I can't think of a way to test this case without a big refactor to the test infra.
Currently in `sample-job.yml` we use `OTEL_GO_AUTO_TARGET_EXE` which should potentially cause the instrumentation of every process running the specified executable. So assuming 2 processes using the same binary the only way to configure the instrumentation to a specific one is passing the `InstrumentationOption` of `WithPID` (we don't have an equielent env var for this option as it does not really make sense for env var to contain pid I think).
If you have something in mind that we can do with the current infra to test this case, I'll be happy to add it.

grcevski (approved):
LGTM! This looks good to me.

MrAlias (approved):

Issue #509, opened 2023-11-16
Fix the correctness of offsets


MrAlias:
The versions of the `google.golang.org/grpc/internal/transport` package below [`v1.14.0`](https://pkg.go.dev/google.golang.org/grpc@v1.14.0/internal/transport?tab=versions) do not exist, but they are reported in our offsets.

It also seems like merging #503 has unexpectedly filled in gaps.

The correctness of the offsets need to be addressed.

_Originally posted by @MrAlias in https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/503#discussion_r1394463028_

MikeGoldsmith:
Following up here instead of the merged PR - guessing at an offset could be a security risk too. If for example, reading data from a HTTP headers, you could accidentally collect unintended data.

PR #781, opened 2024-04-13
Fix offsetgen

RonFed:
* Only consider child entries in DWARF lookup once we find the struct entry. 
* Fix fallback to single offset only when version is unknown.
Resolves the regression described in https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/509
Also, this should fix https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/740#discussion_r1554036803
* Add `go get <mod>@version` command to `Builder` for cases where we look for offsets that are not part of go std lib. This makes sure we are using the correct version. This fixes the problem of un-expected versions the tool found offsets for.


MrAlias (approved):
🏅

PR #790, opened 2024-04-18
Update `net/http` instrumentation to comply with semantic conventions

### LINKED ISSUE #169: Audit and update `net/http` instrumentation to comply with HTTP semantic conventions
https://github.com/open-telemetry/opentelemetry-specification/blob/84a5fa3d1ef42caae8f99f1b431504d7cdb9fffe/specification/trace/semantic_conventions/http.md

# Fixes needed

- [ ] Resolve questions in #740
- [ ] Should we handle `_OTHER` http request method?
- [ ] Conditional checks for including span status
- [ ] Do we need to set `error.type`?
- [ ] Set [`network.protocol.name`](http://network.protocol.name/) if not `http`
- [ ] Set `url.full`

damemi:
Fixes #169 

This does the following:

* Sets `span.Status` for 4xx and 5xx codes in Client, and just 5xx codes in Server
* Sets `protocol.name` when the protocol is not `HTTP` for both Client and Server
* Generates a `url.full` attribute from the protocol+host+path for Client
* Adds Probe tests to Client and Server instrumentation for changes
* Adds a new `opentelemetry/controller_test.go` to check that the output gets exported as expected
* Updates fixtures for HTTP and Gin to show new `url.full` field

This PR adds changes to the net/http eBPF instrumentation to copy the full set of fields from the [`*net/url.URL`](https://pkg.go.dev/net/url#URL) object in the Request to reconstruct a `url.full` attribute. It copies each field individually because the object is a pointer to an address in user space which would not be accessible by the auto-instrumentation agent. The agent reconstructs the Go `URL` object so that it can call [`url.String()`](https://pkg.go.dev/net/url#URL.String), which uses logic and control flags in the original object to build the string representation (rather than building the string manually in the eBPF program).


internal/pkg/instrumentation/bpf/net/http/client/probe.go
```diff
@@ -195,21 +196,33 @@ func convertEvent(e *event) []*probe.SpanEvent {
...
-	return []*probe.SpanEvent{
-		{
-			SpanName:          method,
-			StartTime:         int64(e.StartTime),
-			EndTime:           int64(e.EndTime),
-			SpanContext:       &sc,
-			Attributes:        attrs,
-			ParentSpanContext: pscPtr,
-		},
+	fullURL := fmt.Sprintf("%s://%s%s", scheme, serverAddr.Value.AsString(), path)
```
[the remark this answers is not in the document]
damemi: @RonFed can we get a pointer to the URL struct from the request? https://pkg.go.dev/net/http#Request store that in the Event, and just call `String()` on that?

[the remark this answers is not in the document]
RonFed: Since that pointer will be to some memory in the instrumented process, we won't be able to read from it here. I think we should add the relevant fields of the URL struct to `offsetgen` and read them similarly to the other fields.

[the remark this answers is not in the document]
damemi: Makes sense, they are all pretty straightforward fields except for `Userinfo` which is also a pointer and we'll have to rely on the unexported fields from that struct

internal/pkg/inject/offset_results.json
```diff
@@ -864,7 +864,9 @@
                       "0.10.0",
                       "0.11.0",
                       "0.12.0",
-                      "0.13.0"
+                      "0.13.0",
+                      "1.25.0",
```
RonFed, line 868: That is strange. These offsets are cached and should not have been re-evaluated. This is the reason for the `otelglobal` test failing.

RonFed: Ran it locally without the cache and it didn't produce this diff.

damemi: using the cached version fixed it, so there is either an issue with my local setup or a problem with generating the offsets for otelglobal

damemi: @RonFed ah didn't see your comment, must be something on my end then. Will try to look into it

damemi: @RonFed sorry but just to check, did you also try deleting the `.tools` directory to re-build offsetgen when you tested it?

RonFed: yes

internal/pkg/instrumentation/bpf/net/http/client/bpf/probe.bpf.c
```diff
@@ -6,9 +6,16 @@
...
 
-#define MAX_HOSTNAME_SIZE 256
+#define MAX_HOSTNAME_SIZE 128
 #define MAX_PROTO_SIZE 8
 #define MAX_PATH_SIZE 100
+#define MAX_SCHEME_SIZE 8
+#define MAX_OPAQUE_SIZE 8
+#define MAX_RAWPATH_SIZE 8
+#define MAX_RAWQUERY_SIZE 128
+#define MAX_FRAGMENT_SIZE 50
```
RonFed, line 16: I think it is preferred to have these sizes as powers of multiples of 8 if possible, to avoid padding issues which can cause the pref buffer reader to fail to read the entries successfully.

internal/pkg/opentelemetry/controller_test.go
```diff
@@ -0,0 +1,239 @@
...
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	"go.opentelemetry.io/otel/sdk/trace/tracetest"
+	semconv "go.opentelemetry.io/otel/semconv/v1.25.0"
+	"go.opentelemetry.io/otel/trace"
+
+	"go.opentelemetry.io/auto/internal/pkg/instrumentation/probe"
+)
+
+// copied from instrumentation.go.
+func instResource() *resource.Resource {
```
RonFed, line 43: Is there a way to avoid the duplication? Mabye move the `res()` function to another location?

internal/pkg/instrumentation/bpf/net/http/client/bpf/probe.bpf.c
damemi: sounds good, I'll update. one thing with this approach though is that we're hitting the 512 byte stack limit pretty easily and can't support very long values for a lot of these fields. would like to open a follow up after this to refactor this section

internal/pkg/inject/offset_results.json
damemi: Update: I ran this a couple more times from "fresh" and still got the diff, but it looks like clearing my docker image cache fixed it. Maybe something was up in one of the images I had pulled from a previous run

internal/pkg/opentelemetry/controller_test.go
damemi: the `res()` function is pretty buried within the instrumentation.go code, so I think it would take a bigger refactor of that code to pull it out in any easily re-usable way.

Besides that, this code isn't testing that function anyway. This copy doesn't include all of the `instConfig.res()` code, because this test doesn't need to rely on build info or any of the other fields from an instConfig. So pulling it directly in here isn't relevant to these tests

damemi:
I think test failure is unrelated and fixed by https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/820

internal/pkg/instrumentation/bpf/net/http/client/bpf/probe.bpf.c
```diff
@@ -120,6 +146,49 @@ int uprobe_Transport_roundTrip(struct pt_regs *ctx) {
         bpf_printk("uprobe_Transport_roundTrip: Failed to get path from Request.URL");
     }
 
+    // get scheme from Request.URL
```
grcevski, line 149: Hi @damemi, based on our discussion from the SIG call today, it seems that the code is already using a BPF per-cpu array to store the httpRequest. Do you remember where did you get a stack limit error if the sizes of the arrays were made bigger?

grcevski: I wonder if the error might've been related to stack alignment?

damemi: @grcevski I just tried reproducing it, but I had to add a lot more space to each field than I thought. I'll push my change so you can see the failure, but what it comes up with is this:

```
#0 8.736 bpf/probe.bpf.c:125:5: error: A call to built-in function 'memset' is not supported.
#0 8.736     __builtin_memset(httpReq, 0, sizeof(struct http_request_t));
#0 8.736     ^
#0 8.785 1 error generated.
#0 8.791 Error: can't execute clang: exit status 1
#0 8.792 exit status 1
#0 8.795 internal/pkg/instrumentation/bpf/net/http/client/probe.go:41: running "go": exit status 1
```

googling it I found github comments pointing to the 512 byte limit as causing that, but it could totally be something else I'm not aware of

damemi: @grcevski you can see the failure [here](https://github.com/open-telemetry/opentelemetry-go-instrumentation/actions/runs/8991111452/job/24697949058?pr=790) from the commit I just pushed https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/790/commits/4a9ee82a8eb30a722bc7a484ec4a93334c567cc6

grcevski: Ah I think it's the use of `__buitin_memset` which can only clear up to 512 bytes. I think to workaround this problem we can avoid using memset and manually set each scalar field to 0 and each character array 0th byte to '\0'. So essentially, manually initialize all fields. It should be less operations too, since we'll be setting a lot less bytes to 0, e.g. maybe 30?

Perhaps this shouldn't hold up the PR, we can follow-up with another change if we wanted to increase the sizes.

MrAlias: SGTM

CHANGELOG.md
```diff
@@ -13,7 +13,7 @@ OpenTelemetry Go Automatic Instrumentation adheres to [Semantic Versioning](http
 - kafka-go instrumentation ([#709](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/709))
 - Support `go.opentelemetry.io/otel@v1.26.0`. ([#796](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/796))
 - Support HTTP server path template added in Go version 1.22.
-The `http.route` attribute is included and the span name updated to use this information. ([#740](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/740))
+- The `http.route` attribute is included and the span name updated to use this information. ([#740](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/740))
```
MrAlias, line 16: I think this was intended to be a second sentence.

```suggestion
  The `http.route` attribute is included and the span name updated to use this information. ([#740](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/740))
```

MrAlias:
@RonFed @grcevski please take another look at this and leave a review.

grcevski (approved):
LGTM!

RonFed (approved):
Added a comment about the Makefile change, other than that looks great.

Makefile
```diff
@@ -26,7 +26,7 @@ $(TOOLS):
 	@mkdir -p $@
 $(TOOLS)/%: | $(TOOLS)
 	cd $(TOOLS_MOD_DIR) && \
-	$(GOCMD) build -o $@ $(PACKAGE)
+	$(GOCMD) build  -buildvcs=false -o $@ $(PACKAGE)
```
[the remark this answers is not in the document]
damemi: I opened https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/814 to describe this a bit more, but I couldn't get `make docker-offsets` to work without it. 

Maybe I could make some permissions changes locally, but I think I have a pretty standard git setup. Since this should almost always be run in a container, it should work as the default path.

I don't think there's any benefit to VCS stamping for our tools, so I don't see a downside to adding it. What do you think?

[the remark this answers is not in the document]
MrAlias: Tracking with #814. Going to resolve this.

damemi:
ready to merge @MrAlias

PR #810, opened 2024-04-30
Support empty url path in HTTP client probe

RonFed:
Currently, if we can't read the URL.path we drop the span. This PR changes this to support an empty path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,8 @@ OpenTelemetry Go Automatic Instrumentation adheres to [Semantic Versioning](http
 ### Fixed
 
 - Change HTTP client span name to `{http.request.method}` ([#775](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/775))
-- Support empty url path in HTTP client probe ([#810](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/810))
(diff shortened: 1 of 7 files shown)


RonFed:
The go part will probably change anyway in https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/790, but it is worth avoiding canceling the span on the eBPF side.

damemi (approved):

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ OpenTelemetry Go Automatic Instrumentation adheres to [Semantic Versioning](http
 ### Fixed
 
 - Change HTTP client span name to `{http.request.method}` ([#775](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/775))
+- Support empty url path in HTTP client probe ([#810](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/810))
```
MrAlias, line 19:
```suggestion
- Don't set empty URL path in HTTP client probe. ([#810](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/810))
- Don't fail HTTP client probe attribute resolution on empty URL path. ([#810](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/810))
```

MrAlias (approved):

PR #811, opened 2024-04-30
Extract go version and compiler from instrumneted process

RonFed:
The `process.runtime.version` and `process.runtime.name` resource attributes should be extracted from the instrumented process and not from the agent one.
The rest of the resource attributes such as OS and ARCH are common to the agent and the app.


MrAlias (approved):
:tada:

PR #820, opened 2024-05-06
Commit `go.mod` for e2e test apps

damemi:
Similar to https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/650 for the public samples, this commits the `go.mod`/`go.sum` files for each e2e sample app.

I hit an issue with this in [this test failure](https://github.com/open-telemetry/opentelemetry-go-instrumentation/actions/runs/8973363998/job/24643354707?pr=790#step:6:128) where our gRPC e2e was failing with:

```
 > [4/4] RUN go mod init go.opentelemetry.io/auto/internal/test/e2e/grpc && go mod tidy && go build -o main:
0.702 go: found google.golang.org/grpc/examples/helloworld/helloworld in google.golang.org/grpc/examples v0.0.0-20240506165903-befc29de931d
0.704 go: downloading google.golang.org/protobuf v1.33.0
0.708 go: downloading google.golang.org/genproto/googleapis/rpc v0.0.0-20240318140521-94a12d6c2237
0.733 go: downloading google.golang.org/genproto v0.0.0-20240227224415-6ceb2ff114de
0.946 go: downloading github.com/google/go-cmp v0.6.0
0.982 go: downloading golang.org/x/net v0.23.0
1.256 go: downloading golang.org/x/sys v0.18.0
1.505 go: downloading golang.org/x/text v0.14.0
14.86 # google.golang.org/grpc/examples/helloworld/helloworld
Error: 14.86 /go/pkg/mod/google.golang.org/grpc/examples@v0.0.0-20240506165903-befc29de931d/helloworld/helloworld/helloworld_grpc.pb.go:33:16: undefined: grpc.SupportPackageIsVersion9
```

Digging into it, this was because the `google.golang.org/grpc/examples` dependency for our e2e was [updated 3 days ago](https://github.com/grpc/grpc-go/blame/4879d51a597afb5249e03a333a7a8da28ebe98ed/examples/helloworld/helloworld/helloworld_grpc.pb.go#L32-L33). But, because we don't pin a specific version of this package (and they don't actually publish tagged versions of the `examples/helloworld` package), our e2es will just pull from latest. This causes a conflict trying to depend on unreleased code.

This PR commits all of the `go.mod` files for our e2es and removes that step from their dockerfiles. We should be doing this anyway for consistency in our tests, rather than running a fresh `go mod init` within the test.

PR #821, opened 2024-05-06
Update offsets

damemi:


damemi:
Mutually blocked with the test failure fixed by https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/820, so moving this commit there

PR #842, opened 2024-05-13
Add support to custom log level through command line flag and environment variable

### LINKED ISSUE #691: Add log verbosity configuration and move the log in controller.go to debug level
Currently, we print a log message for each span: 
https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/61ff124646105a2058863b17da1815d8d42d78b3/internal/pkg/opentelemetry/controller.go#L54

We should change that to have a debug level.
We don't have a configuration for verbosity. We probably should add support for configuration via:
1. command line arguments
2. `InstrumentationOption`
3. Enviroment variable (?)

This will need to be modified:
https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/61ff124646105a2058863b17da1815d8d42d78b3/instrumentation.go#L70-L83

vitorhugoro1:
- Added on `InstrumentationOption` the option to define log level.
- Implemented a custom severity enum, called level.go, with options to set log levels based on text using `zap.Level` as reference.
- Added on CLI the flag `-log-level` that accept one of `LogLevel` value, which can be see on [level.go](https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/91e4e34ef58b7a05ca1e72110b6477f5cf942cbe/level.go) ex. `otel-go-instrumentation -log-level error`, the default value is `info`.
- Added support to `OTEL_LOG_LEVEL` which implements `LogLevel` to set visibility level through environment variable.
-  Changed multiples logs level to Debug to improve setup when using in production environment.  

- Closes #691


instrumentation.go
```diff
@@ -463,3 +474,12 @@ func WithResourceAttributes(attrs ...attribute.KeyValue) InstrumentationOption {
 		return c, nil
 	})
 }
+
+// WithLogLevel returns an [InstrumentationOption] that will configure
+// an [Instrumentation] with the logger level visibility defined as inputed.
+func WithLogLevel(level string) InstrumentationOption {
+	return fnOpt(func(ctx context.Context, c instConfig) (instConfig, error) {
```
grcevski, line 481: What do you think about parsing the log level here and erroring out if the user has specified something invalid? It would be nice to know for example if I've miss-spelled "debug", instead of silently switching to "info".

instrumentation_test.go
```diff
@@ -172,6 +172,14 @@ func TestWithResourceAttributes(t *testing.T) {
 	})
 }
 
+func TestWithLogLevel(t *testing.T) {
+	c, err := newInstConfig(context.Background(), []InstrumentationOption{WithLogLevel("error")})
```
grcevski, line 176: Thanks for adding these tests! Do you mind extending the testing to cover what happens if invalid options are specified or the configuration doesn't contain logging level?

This is related to my earlier comment that parsing of the log level should be done here, so that invalid command line input is detected.

instrumentation.go
```diff
@@ -463,3 +474,12 @@ func WithResourceAttributes(attrs ...attribute.KeyValue) InstrumentationOption {
 		return c, nil
 	})
 }
+
+// WithLogLevel returns an [InstrumentationOption] that will configure
+// an [Instrumentation] with the logger level visibility defined as inputed.
+func WithLogLevel(level string) InstrumentationOption {
```
MrAlias, line 480: Instead of accepting a string with special values, we should introduce a `LogLevel` type with `enum` values. We might want to use the [OTel levels](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#field-severitynumber). For example:

https://github.com/open-telemetry/opentelemetry-go/blob/main/log/severity.go

vitorhugoro1: Do you think we need to create something which uses numbers as reference like the one you linked?

MrAlias: That seems reasonable to me.

MrAlias: We already do something like this for offsetgen:

https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/e85ddd9de6c5829d6736d7ccefbc3733f2cb8ed8/internal/tools/inspect/cmd/offsetgen/main.go#L59

We could provided human readable flags for the CLI though (i.e. `--log-level info`).

vitorhugoro1: Yeah, I usually prefer something human readable to CLI because is more quick to identify.

instrumentation.go
RonFed: It looks like [ParseAtomicLevel](https://pkg.go.dev/go.uber.org/zap#ParseAtomicLevel) does some validation if we choose to use it (currently it looks like it is used in `newLogger`)

RonFed:
@vitorhugoro1 Do you plan on changing the level of the log line I mentioned in the issue?
We can use debug level in our tests in https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/e85ddd9de6c5829d6736d7ccefbc3733f2cb8ed8/.github/workflows/e2e/k8s/sample-job.yml#L29

pellared (commented):
Shouldn't we support setting via `OTEL_LOG_LEVEL` env var?

Reference: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md#general-sdk-configuration

MrAlias:
> Shouldn't we support setting via `OTEL_LOG_LEVEL` env var?
> 
> Reference: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md#general-sdk-configuration

Please see https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/691

That appears out of scope for this PR.

pellared:
> > Shouldn't we support setting via `OTEL_LOG_LEVEL` env var?
> > Reference: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md#general-sdk-configuration
> 
> That is tracked in #691. It is out of scope for this PR.

This PR says

> Closes https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/691

vitorhugoro1:
> > Shouldn't we support setting via `OTEL_LOG_LEVEL` env var?
> > Reference: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md#general-sdk-configuration
> 
> That is tracked in #691. It is out of scope for this PR.

Yeah this PR will implement this change too, I take the approach just with flags because is the way I usually use log levels on container

instrumentation.go
vitorhugoro1: Added the validation, what do you think?

instrumentation_test.go
vitorhugoro1: Created the test to cover the validation of a wrong logging level

instrumentation.go
vitorhugoro1: @MrAlias I have create a simple Level Enum for logs, what do you think?

vitorhugoro1:
> @vitorhugoro1 Do you plan on changing the level of the log line I mentioned in the issue? In addition, we can use debug level in our tests at
> 
> https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/e85ddd9de6c5829d6736d7ccefbc3733f2cb8ed8/.github/workflows/e2e/k8s/sample-job.yml#L29

I changed some log level, not just in the line you mentioned but in another which is more suitable to debugging not exactly on a production setup.

Changed the sample job too to use debug level by default

vitorhugoro1:
> Shouldn't we support setting via `OTEL_LOG_LEVEL` env var?
> 
> Reference: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md#general-sdk-configuration

Added the supporting via environment with the name you suggest

instrumentation.go
grcevski: Awesome!

internal/pkg/instrumentation/manager.go
```diff
@@ -199,23 +199,23 @@ func (m *Manager) load(target *process.TargetDetails) error {
 
 	// Load probes
 	for name, i := range m.probes {
-		m.logger.Info("loading probe", "name", name)
+		m.logger.V(-1).Info("loading probe", "name", name)
```
grcevski, line 202: I think this might be useful to keep as info, to know which probes are activated.

internal/pkg/process/allocate.go
```diff
@@ -81,7 +81,7 @@ func remoteAllocate(logger logr.Logger, pid int, mapSize uint64) (uint64, error)
 	}
 
 	defer func() {
-		logger.Info("Detaching from process", "pid", pid)
+		logger.V(-1).Info("Detaching from process", "pid", pid)
```
grcevski, line 84: I'd keep this as info too.

internal/pkg/process/discover.go
```diff
@@ -70,16 +70,16 @@ func (a *Analyzer) DiscoverProcessID(ctx context.Context, target *TargetArgs) (i
...
 		select {
 		case <-ctx.Done():
-			a.logger.Info("stopping process id discovery due to kill signal")
+			a.logger.V(-1).Info("stopping process id discovery due to kill signal")
 			return 0, ErrInterrupted
 		case <-t.C:
 			pid, err := a.findProcessID(target, proc)
 			if err == nil {
-				a.logger.Info("found process", "pid", pid)
+				a.logger.V(-1).Info("found process", "pid", pid)
```
grcevski, line 78: Same here, keep this under Info.

internal/pkg/process/allocate.go
vitorhugoro1: Changed to info

internal/pkg/process/discover.go
vitorhugoro1: Changed to info

internal/pkg/instrumentation/manager.go
vitorhugoro1: Changed to info

cli/main.go
```diff
@@ -68,8 +68,11 @@ func newLogger() logr.Logger {
 
 func main() {
 	var globalImpl bool
+	var logLevel string
 
 	flag.BoolVar(&globalImpl, "global-impl", false, "Record telemetry from the OpenTelemetry default global implementation")
+	flag.StringVar(&logLevel, "log-level", "info", "Define log visibility level")
```
RonFed, line 74:
```suggestion
	flag.StringVar(&logLevel, "log-level", "", "Define log visibility level")
```

I think we should not set the default value here as info. Doing so will cause `OTEL_LOG_LEVEL=DEBUG` to be ignored since we add `WithLogLevel` after `WhitEnv` this is expected, but from a user who is running the agent image point of view this is not expected.
In addition we set the default level to info in `newInstConfig` if it is not define but env or by option.

cli/main.go
```diff
@@ -98,6 +101,8 @@ func main() {
 		instOptions = append(instOptions, auto.WithGlobal())
 	}
 
+	instOptions = append(instOptions, auto.WithLogLevel(logLevel))
```
RonFed, line 104:
```suggestion
	if logLevel != "" {
		instOptions = append(instOptions, auto.WithLogLevel(logLevel))
	}
```

This relates to my other comment in this file

cli/main.go
vitorhugoro1: Yeah, makes sense what you say, changed to use like you recommended.

I have already set info as default on `newInstConfig`.

instrumentation.go
```diff
@@ -371,6 +385,11 @@ func WithEnv() InstrumentationOption {
 			c.serviceName = name
 			c.additionalResAttrs = append(c.additionalResAttrs, attrs...)
 		}
+		if l, ok := lookupEnv(envLogLevelKey); ok {
```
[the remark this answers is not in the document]
vitorhugoro1: Updated

cli/main.go
vitorhugoro1: Changed

RonFed (approved):

vitorhugoro1:
@pellared @MrAlias @grcevski it's okay now for you guys?

grcevski (approved):
LGTM!

internal/pkg/log/level.go
```diff
@@ -0,0 +1,104 @@
...
+	// Log Level [debug].
+	LevelDebug Level = "debug"
+	// Log Level [info].
+	LevelInfo Level = "info"
+	// Log Level [warn].
+	LevelWarn Level = "warn"
+	// Log Level [error].
+	LevelError Level = "error"
+	// Log Level [dpanic].
+	LevelDPanic Level = "dpanic"
```
MrAlias, line 35: What is dpanic?

internal/pkg/log/level.go
```diff
@@ -0,0 +1,104 @@
...
+	// Log Level [panic].
+	LevelPanic Level = "panic"
+	// Log Level [fatal].
+	LevelFatal Level = "fatal"
```
MrAlias, line 39: What is the difference between panic and fatal? Shouldn't these both be terminations of the application?

instrumentation.go
MrAlias: > @MrAlias I have create a simple Level Enum for logs, what do you think?

This still accepts a `string`.

```suggestion
func WithLogLevel(level Level) InstrumentationOption {
```

internal/pkg/log/level.go
```diff
@@ -0,0 +1,104 @@
...
+
+package log
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+)
+
+type Level string
```
MrAlias, line 23: Missing comment.

internal/pkg/log/level.go
```diff
@@ -0,0 +1,104 @@
...
+	LevelInfo Level = "info"
+	// Log Level [warn].
+	LevelWarn Level = "warn"
+	// Log Level [error].
+	LevelError Level = "error"
+	// Log Level [dpanic].
+	LevelDPanic Level = "dpanic"
+	// Log Level [panic].
+	LevelPanic Level = "panic"
+	// Log Level [fatal].
```
MrAlias, line 38: These comments do not explain the levels. Please update so they communicate the logging strata they intend to enable.

internal/pkg/log/level.go
MrAlias, line 23: This needs to be in the exported `auto` package so users can reference it.

MrAlias (changes requested):

internal/pkg/log/level.go
vitorhugoro1: These two is to replicate what we see on zapcore.levels which is our base config

instrumentation.go
vitorhugoro1: @MrAlias  This case here expect an string because is the input when use flag or from env, so for properly work is these scenarios this need to be string

internal/pkg/log/level.go
MrAlias: I would replicate [OTel logging levels](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#field-severitynumber) if we are going to model this on a preexisting structure.

internal/pkg/log/level.go
vitorhugoro1: I will remove this cases

internal/pkg/log/level.go
vitorhugoro1: I will remove these too

internal/pkg/log/level.go
vitorhugoro1: Added comment

internal/pkg/log/level.go
vitorhugoro1: Added comments to levels

internal/pkg/log/level.go
vitorhugoro1: I removed these options

internal/pkg/log/level.go
vitorhugoro1: I removed these options because they don't seem to make sense for the project

instrumentation.go
MrAlias: The environment variable needs to be parsed before it is passed to this function then.

We need to use syntax here to ensure the best possible chance values passed will be valid. Using an arbitrary string with magic values is not appropriate.

internal/pkg/log/level.go
MrAlias: This lives in an internal package currently. It needs to be moved to the user accessible `go.opentelemetry.io/auto` package instead.

vitorhugoro1: @MrAlias nice, changed to `auto` package the level

instrumentation.go
vitorhugoro1: @MrAlias changed `WithLogLevel` to work on the behaviour you said

vitorhugoro1:
@MrAlias any things needed here yet?

RonFed:
@vitorhugoro1 bump 😄

instrumentation.go
```diff
@@ -463,3 +482,17 @@ func WithResourceAttributes(attrs ...attribute.KeyValue) InstrumentationOption {
 		return c, nil
 	})
 }
+
+// WithLogLevel returns an [InstrumentationOption] that will configure
+// an [Instrumentation] with the logger level visibility defined as inputed.
+func WithLogLevel(level Level) InstrumentationOption {
+	return fnOpt(func(ctx context.Context, c instConfig) (instConfig, error) {
+		if err := level.UnmarshalText([]byte(level.String())); err != nil {
```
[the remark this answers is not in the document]
vitorhugoro1: Added a validate method to cover this case

instrumentation.go
```diff
@@ -371,6 +385,11 @@ func WithEnv() InstrumentationOption {
 			c.serviceName = name
 			c.additionalResAttrs = append(c.additionalResAttrs, attrs...)
 		}
+		if l, ok := lookupEnv(envLogLevelKey); ok {
+			var e error
+			c.logLevel, e = ParseLevel(l)
```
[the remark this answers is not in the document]
vitorhugoro1: Changed

instrumentation.go
```diff
@@ -204,6 +213,10 @@ func newInstConfig(ctx context.Context, opts []InstrumentationOption) (instConfi
 		c.sampler = trace.AlwaysSample()
 	}
 
+	if c.logLevel == "" {
```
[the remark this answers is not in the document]
vitorhugoro1: Changed

level.go
```diff
@@ -0,0 +1,87 @@
...
+	"errors"
+	"fmt"
+)
+
+// Level defines the log level which instrumentation uses.
+type Level string
+
+const (
+	// Log Level [debug] Print all logs generate by instrumentation.
+	LevelDebug Level = "debug"
```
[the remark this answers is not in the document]
vitorhugoro1: Applied the suggested

instrumentation.go
```diff
@@ -68,8 +70,17 @@ type Instrumentation struct {
 // binary or pid.
 var errUndefinedTarget = fmt.Errorf("undefined target Go binary, consider setting the %s environment variable pointing to the target binary to instrument", envTargetExeKey)
 
-func newLogger() logr.Logger {
-	zapLog, err := zap.NewProduction()
+func newLogger(logLevel Level) logr.Logger {
+	level, err := zap.ParseAtomicLevel(logLevel.String())
+	if err != nil {
+		level, _ = zap.ParseAtomicLevel(LevelInfo.String())
```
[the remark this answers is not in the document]
vitorhugoro1: Added the log

level.go
```diff
@@ -0,0 +1,87 @@
...
+func (l Level) String() string {
+	switch l {
+	case LevelDebug:
+		return "debug"
+	case LevelInfo:
+		return "info"
+	case LevelWarn:
+		return "warn"
+	case LevelError:
+		return "error"
```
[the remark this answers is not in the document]
vitorhugoro1: Changed and added the LevelUndefined too

level.go
```diff
@@ -0,0 +1,87 @@
...
+	LevelDebug Level = "debug"
+	// Log Level [info].Print most of logs where these is running status of instrumentation.
+	LevelInfo Level = "info"
+	// Log Level [warn].Print errors and logs which need attention on instrumentation.
+	LevelWarn Level = "warn"
+	// Log Level [error].Print errors generated by instrumentation.
+	LevelError Level = "error"
+)
+
+func (l Level) String() string {
```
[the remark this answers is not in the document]
vitorhugoro1: Changed

level.go
```diff
@@ -0,0 +1,87 @@
...
+type Level string
+
+const (
+	// Log Level [debug] Print all logs generate by instrumentation.
+	LevelDebug Level = "debug"
+	// Log Level [info].Print most of logs where these is running status of instrumentation.
+	LevelInfo Level = "info"
+	// Log Level [warn].Print errors and logs which need attention on instrumentation.
+	LevelWarn Level = "warn"
+	// Log Level [error].Print errors generated by instrumentation.
```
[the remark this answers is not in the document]
vitorhugoro1: Changed

level.go
```diff
@@ -0,0 +1,87 @@
...
+package auto
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+)
+
+// Level defines the log level which instrumentation uses.
+type Level string
```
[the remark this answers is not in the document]
vitorhugoro1: Makes sense, changed to LogLevel

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ OpenTelemetry Go Automatic Instrumentation adheres to [Semantic Versioning](http
 - Support Go `v1.21.10`. ([#824](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/824))
 - Support Go `v1.22.3`. ([#824](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/824))
 - Support `google.golang.org/grpc` `1.65.0-dev`. ([#827](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/827))
+- Add support to log level through command line flag. ([#842](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/842))
```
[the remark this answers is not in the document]
vitorhugoro1: Moved to unreleased section and added the recommended

vitorhugoro1:
Sorry for the late guys, applied the changes @MrAlias suggested and rebase with the main branch

instrumentation.go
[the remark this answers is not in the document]
MrAlias: Is there a missing commit? This still stores the returned error as `_`.

level.go
```diff
@@ -0,0 +1,94 @@
...
+	"errors"
+	"fmt"
+)
+
+// Level defines the log level which instrumentation uses.
+type LogLevel string
+
+const (
+	// LevelUndefined is an unset log level, it should not be used.
+	LevelUndefined LogLevel = ""
```
MrAlias, line 28: Having this unexported will help prevent users from accidentally using it. It is only added to easily test unset or undefined logging levels.

```suggestion
	// levelUndefined is an unset log level, it should not be used.
	levelUndefined LogLevel = ""
```

level.go
```diff
@@ -0,0 +1,94 @@
...
+
+package auto
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+)
+
+// Level defines the log level which instrumentation uses.
```
MrAlias, line 23:
```suggestion
// LogLevel defines the log level which instrumentation uses.
```

level.go
```diff
@@ -0,0 +1,94 @@
...
+	// LevelUndefined is an unset log level, it should not be used.
+	LevelUndefined LogLevel = ""
+	// LevelDebug sets the logging level to log all messages.
+	LevelDebug LogLevel = "debug"
+	// LevelInfo sets the logging level to log only informational, warning, and error messages.
+	LevelInfo LogLevel = "info"
+	// LevelWarn sets the logging level to log only warning and error messages.
+	LevelWarn LogLevel = "warn"
+	// LevelError sets the logging level to log only error messages.
+	LevelError LogLevel = "error"
```
MrAlias, line 36:
```suggestion
	// LogLevelDebug sets the logging level to log all messages.
	LogLevelDebug LogLevel = "debug"
	// LogLevelInfo sets the logging level to log only informational, warning, and error messages.
	LogLevelInfo LogLevel = "info"
	// LogLevelWarn sets the logging level to log only warning and error messages.
	LogLevelWarn LogLevel = "warn"
	// LogLevelError sets the logging level to log only error messages.
	LogLevelError LogLevel = "error"
```

level.go
```diff
@@ -0,0 +1,94 @@
...
+)
+
+// String returns the string encoding of the Level l.
```
MrAlias, line 39:
```suggestion
// String returns the string encoding of the LogLevel l.
```

level.go
```diff
@@ -0,0 +1,94 @@
...
+		*l = LevelError
+	default:
+		return false
+	}
+
+	return true
+}
+
+// ParseLevel return a new LogLevel using text, and will return err if inputted text is not accepted.
+func ParseLevel(text string) (LogLevel, error) {
```
MrAlias, line 88:
```suggestion
// ParseLogLevel return a new LogLevel parsed from text. A non-nil error is returned if text is not a valid LogLevel.
func ParseLogLevel(text string) (LogLevel, error) {
```

level.go
```diff
@@ -0,0 +1,94 @@
...
+	case "warn":
+		*l = LevelWarn
+	case "error":
+		*l = LevelError
+	default:
+		return false
+	}
+
+	return true
+}
```
MrAlias, line 85: This makes redundant calls to `unmarshalText`. Refactoring `validate` and `unmarshalText` will simplify:

```suggestion
// UnmarshalText applies the LogLevel type when inputted text is valid.
func (l *LogLevel) UnmarshalText(text []byte) error {
	*l = LogLevel(bytes.ToLower(text))
	return l.validate()
}

var errInvalidLogLevel = errors.New("invalid LogLevel")

func (l *LogLevel) validate() error {
	if l == nil {
		return errors.New("nil LogLevel")
	}

	switch *l {
	case LevelDebug, LevelInfo, LevelWarn, LevelError:
		// Valid.
	default:
		return fmt.Errorf("%w: %s", errInvalidLogLevel, l.String())
	}
	return nil
}
```

`TestValidate` can then be updated to just:

```go
func TestValidate(t *testing.T) {
	l := LogLevel("notexist")
	assert.ErrorIs(t, l.validate(), errInvalidLogLevel)
}
```

instrumentation.go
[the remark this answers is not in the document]
vitorhugoro1: Oh this err is from zapcore itself, here I manually force the info. The validation for change to Info is make on the added line 98, where I catch if the first `ParseAtomicLevel` generated some error and show the message to user

level.go
vitorhugoro1: Changed

level.go
vitorhugoro1: Changed

level.go
vitorhugoro1: Changed

level.go
vitorhugoro1: Changed

level.go
vitorhugoro1: Changed

level.go
vitorhugoro1: Applied the suggestion and simplified TestValidate

instrumentation.go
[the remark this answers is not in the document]
MrAlias: Understood. The issue is though that you will have a user that, say, sets the "trace" logging level expecting very detailed logs. We need to tell that user somehow that we have fallen back to using `info` logging and that error is a good way to communicate this.

instrumentation.go
```diff
@@ -83,6 +94,10 @@ func newLogger() logr.Logger {
 		logger = zapr.NewLogger(zapLog)
 	}
 
+	if logErr != nil {
+		logger.V(2).Info("error to parse log level, changed to LevelInfo by default")
```
MrAlias, line 98:
```suggestion
		logger.Error("invalid log level; using LevelInfo instead", zap.Error(logErr), zap.String("input", logLevel.String()))
```

instrumentation.go
```diff
@@ -383,6 +401,16 @@ func WithEnv() InstrumentationOption {
 				c.globalImpl = boolVal
 			}
 		}
+		if l, ok := lookupEnv(envLogLevelKey); ok {
+			var e error
+			level, e := ParseLogLevel(l)
+
+			if err == nil {
```
MrAlias, line 408:
```suggestion
			if e == nil {
```

instrumentation.go
```diff
@@ -490,3 +518,17 @@ func WithLoadedIndicator(indicator chan struct{}) InstrumentationOption {
 		return c, nil
 	})
 }
+
+// WithLogLevel returns an [InstrumentationOption] that will configure
+// an [Instrumentation] with the logger level visibility defined as inputed.
```
MrAlias, line 523:
```suggestion
// an [Instrumentation] to use the provided logging level.
```

level.go
```diff
@@ -0,0 +1,79 @@
...
+	"bytes"
+	"errors"
+	"fmt"
+)
+
+// LogLevel defines the log level which instrumentation uses.
+type LogLevel string
+
+const (
+	// LevelUndefined is an unset log level, it should not be used.
```
MrAlias, line 27:
```suggestion
	// logLevelUndefined is an unset log level, it should not be used.
```

level_test.go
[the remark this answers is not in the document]
MrAlias: Bump.

level_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+package auto
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestLevel(t *testing.T) {
```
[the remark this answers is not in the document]
MrAlias: Bump.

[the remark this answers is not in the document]
vitorhugoro1: Changed

level_test.go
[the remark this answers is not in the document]
vitorhugoro1: Added tests to cover ParseLogLevel

instrumentation.go
vitorhugoro1: Nice, changed!

instrumentation.go
vitorhugoro1: Changed

level.go
vitorhugoro1: Changed

RonFed:
@MrAlias can you please have another look?

vitorhugoro1:
@MrAlias can you review again, please?

PR #844, opened 2024-05-17
Collect user provided tracer name, version and schemaURL

RonFed:
When the `WithGlobal` InstrumentationOption is used, use the user provided tracer name.


internal/pkg/opentelemetry/controller.go
```diff
@@ -35,17 +35,32 @@ type Controller struct {
...
 
-func (c *Controller) getTracer(pkg string) trace.Tracer {
-	t, exists := c.tracersMap[pkg]
+func (c *Controller) getTracer(pkg, tracerName string) trace.Tracer {
+	var (
+		newTracer trace.Tracer
+		tracerKey = pkg
+	)
+
+	if tracerName != "" {
```
MrAlias, line 44:
```suggestion
	tracerKey := pkg
	if tracerName != "" {
```

internal/pkg/opentelemetry/controller.go
```diff
@@ -35,17 +35,32 @@ type Controller struct {
...
 	if exists {
 		return t
 	}
 
-	newTracer := c.tracerProvider.Tracer(
-		"go.opentelemetry.io/auto/"+pkg,
-		trace.WithInstrumentationVersion(c.version),
-	)
-	c.tracersMap[pkg] = newTracer
+	if tracerName != "" {
```
MrAlias, line 53:
```suggestion
	var newTracer trace.Tracer
	if tracerName != "" {
```

MrAlias (changes requested):
The `Tracer` name, version, and scope are the three [identifying fields](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#get-a-tracer). By only using the tracer name here we are going to have collisions with the wrong tracer being used.  The version and schema URL need to also be determined and used in the tracer resolution path.

MrAlias (approved):
:tada:

RonFed:
@MrAlias I think this will break once the new version of `otel-go` is released (since the tracers' map key size will change), so maybe it is better to wait with this PR. Once the new version is out I will add code to adjust the key size

MrAlias:
Working on otel release here: https://github.com/open-telemetry/opentelemetry-go/pull/5569#

MrAlias:
@RonFed [Releases v1.28.0/v0.50.0/v0.4.0](https://github.com/open-telemetry/opentelemetry-go/releases/tag/v1.28.0) was made yesterday. This should be unblocked.

internal/pkg/opentelemetry/controller.go
```diff
@@ -30,21 +30,35 @@ type Controller struct {
...
+	tracersMap     map[tracerID]trace.Tracer
 	bootTime       int64
 }
 
-func (c *Controller) getTracer(pkg string) trace.Tracer {
-	t, exists := c.tracersMap[pkg]
+type tracerID struct{ name, version, schema string }
+
+func (c *Controller) getTracer(pkg, tracerName, version, schema string) trace.Tracer {
+	tID := tracerID{name: pkg, version: c.version}
```
MrAlias, line 40: Doesn't this need to include the schema?

```suggestion
	tID := tracerID{name: pkg, version: c.version, schema: schema}
```

internal/pkg/opentelemetry/controller.go
```diff
@@ -30,21 +30,35 @@ type Controller struct {
...
+	if tracerName != "" {
+		tID = tracerID{name: tracerName, version: version, schema: schema}
```
MrAlias, line 42: nit

```suggestion
		tID.name = tracerName
```

internal/pkg/opentelemetry/controller.go
```diff
@@ -30,21 +30,35 @@ type Controller struct {
...
-	c.tracersMap[pkg] = newTracer
+	var newTracer trace.Tracer
+	if tracerName != "" {
+		// If the user has provided a tracer, use it.
+		newTracer = c.tracerProvider.Tracer(tracerName, trace.WithInstrumentationVersion(version), trace.WithSchemaURL(schema))
+	} else {
+		newTracer = c.tracerProvider.Tracer(
+			"go.opentelemetry.io/auto/"+pkg,
+			trace.WithInstrumentationVersion(c.version),
+		)
```
MrAlias, line 58:
```suggestion
			trace.WithInstrumentationVersion(c.version),
			trace.WithSchemaURL(schema))
		)
```

internal/pkg/opentelemetry/controller.go
RonFed: I don't think this is correct. The `version`  argument is the user-defined version we collected in the probe, and `c.version` is our distro version. So the assumption here is that if we have a user-provided tracer name we should also take the user-provided version and schema.
It is a bit confusing.

internal/pkg/opentelemetry/controller.go
RonFed: Relates to the above comment. `schema` here is the user-provided schema (if exists). Hence, if we provide a schema here it should be something related to our distro (since this case deals with all the fully auto-generated spans).

MrAlias: That makes sense. Can we then set `semconv.SchemaURL` from [`semconv/v1.26.0`](https://pkg.go.dev/go.opentelemetry.io/otel/semconv/v1.26.0)?

RonFed: We set it here: https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/c6620c7f9f80bcd1ad6fc59278a469d0a743da15/instrumentation.go#L285

I'm not sure about the differences between schema URL for the tracerProvider and the tracer, and in our case will they be the same or not?

MrAlias: That schema URL is for resource which indicates what semantic conventions the resource itself contain: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/resource/README.md

The `Tracer` schema URL indicates what semantic convention version the span data is recorded with. This is the field that populates this in OTLP: https://github.com/open-telemetry/opentelemetry-proto-go/blob/109641596aab69d0bec6e1f2856ecf749de88b67/otlp/trace/v1/trace.pb.go#L386-L390

RonFed: I see.
Setting the schema URL here from the latest semconv assumes all the probes are updated to the latest semconv right? This is currently the case but will probably need some maintenance, and each probe can have a different version as well.

MrAlias: > This is currently the case but will probably need some maintenance, and each probe can have a different version as well.

True. We will need to be diligent about this and likely need to build some testing to ensure this is accurate.

internal/pkg/opentelemetry/controller.go
```diff
@@ -19,21 +19,36 @@ type Controller struct {
...
+	tracersMap     map[tracerID]trace.Tracer
 	bootTime       int64
 }
 
-func (c *Controller) getTracer(pkg string) trace.Tracer {
-	t, exists := c.tracersMap[pkg]
+type tracerID struct{ name, version, schema string }
+
+func (c *Controller) getTracer(pkg, tracerName, version, schema string) trace.Tracer {
+	tID := tracerID{name: pkg, version: c.version}
```
MrAlias, line 29:
```suggestion
	// Default Tracer ID, if the user does not provide one.
	tID := tracerID{name: pkg, version: c.version}
```

internal/test/e2e/otelglobal/main.go
```diff
@@ -10,9 +10,14 @@ import (
...
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/codes"
+	"go.opentelemetry.io/otel/trace"
 )
 
-var tracer = otel.Tracer("trace-example")
+var tracer = otel.Tracer(
+	"trace-example",
+	trace.WithInstrumentationVersion("v1.23.42"),
+	trace.WithSchemaURL("https://some_shcema"),
```
MrAlias, line 19: nit:

```suggestion
	trace.WithSchemaURL("https://some_schema"),
```

internal/test/e2e/otelglobal/traces.json
```diff
@@ -50,9 +50,10 @@
       "schemaUrl": "https://opentelemetry.io/schemas/1.26.0",
       "scopeSpans": [
         {
+          "schemaUrl": "https://some_shcema",
```
MrAlias, line 53:
```suggestion
          "schemaUrl": "https://some_schema",
```

MrAlias (approved):
Looks good to me overall. Minor nits.

MrAlias:
Looks like some tests need to be updated for the added schema URLs.

PR #863, opened 2024-05-31
Add make target 'codespell'

rockdaboot:


Makefile
```diff
@@ -132,6 +132,14 @@ verify-licenses: generate $(GOLICENSES)
       exit 1; \
     fi; \
 
+.PHONY: spellcheck
+spellcheck:
+	@codespell $(CODESPELL_ARGS) -Lmmaped,ect $$(git ls-files | grep -Ev '^CHANGELOG.md$$|(go.mod|go.sum)$$|^LICENSES/')
```
MrAlias, line 137: This assumes codespell is installed on the system running this target. This is not a valid assumption.

We should do something similar to what was done in https://github.com/open-telemetry/opentelemetry-go/pull/3996.

rockdaboot: I didn't like the spilling of `venv` etc into the work directory. So I did the docker based codespell slightly different.

pellared: I prefer the way it is done in https://github.com/open-telemetry/opentelemetry-go/pull/3996.

Reasons:
- pins the version of codespell
- reruns should be faster

rockdaboot: >     * pins the version of codespell

Pinning versions comes with the cost of manual maintaining these version and deciding for upgrades ("polling mode"). And when you don't do this regularly, you'll get of of sync with upstream, lose functionality or get it too late. And it doesn't scale... the more projects you maintain, the more time you spend in updating versions instead of doing more important or more interesting things.
From my experience the "event driven mode" is more effective in terms of workload. When automatically going with the latest (docker images, packages etc), you get notified if something in the CI breaks and you can fix it. Often, version updates don't break things and you as the maintainer have no additional work to be done.

But if you prefer version pinning (also for the docker image?), I'll do it (your house, your rules).

>     * reruns should be faster

I'll give it a spin, for me it is just a few seconds, for the CI it should be the same. (User won't run it anyways, do you expect developers to run this manually often?)

rockdaboot:
Extracted the actual fixes of spelling errors to #867

Makefile
rockdaboot: Also:
- `venv` clutters the repository directory
- `venv` contains files owned by root, you can't just wipe it
- `make codespell` outputs this error (might be something on my side):
```
docker: Error response from daemon: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec: "venv/bin/codespell.PHONY:": stat venv/bin/codespell.PHONY:: no such file or directory: unknown.
make: *** [Makefile:247: codespell] Error 127
```

pellared: > when you don't do this regularly

That is why we use dependabot/renovate

Pinning the version gives other benefits such as:

- if there is a vulnerability we know in which places we could be affected
- We can go back in git history and the build would be reproducible
- The build will not start failing when e.g. a new version detects a new typo or when there is some bug in codespell

rockdaboot:
> Extracted the actual fixes of spelling errors to #867

Now, the `codespell` make target is running locally. It is same as in https://github.com/open-telemetry/opentelemetry-go/blob/main/Makefile.

.codespellrc
```diff
@@ -0,0 +1,10 @@
+# https://github.com/codespell-project/codespell
+[codespell]
+builtin = clear,rare,informal
+check-filenames =
+check-hidden =
+ignore-words = .codespellignore
+interactive = 1
+skip = .git,go.mod,go.sum,CHANGELOG.md,LICENSES,venv
```
MrAlias, line 8: Can we add `internal/include/libbpf` here? It was mentioned in the last SIG meeting that that directory is just copied from upstream. Fixing upstream spelling errors here will just result in thrash.

cc @grcevski

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ OpenTelemetry Go Automatic Instrumentation adheres to [Semantic Versioning](http
 - Support `golang.org/x/net` v0.26.0. ([#871](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/871))
 - Support Go `v1.21.11`. ([#871](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/871))
 - Support Go `v1.22.4`. ([#871](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/871))
+- Add make target 'codespell'. ([#863](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/863))
```
MrAlias, line 30: This needs to be moved to the "Unreleased" section.

MrAlias (commented):
Looks good overall. One question about ignoring copied files though.

.codespellrc
MrAlias: Related: https://github.com/libbpf/libbpf/pull/819

rockdaboot: Fixed

CHANGELOG.md
rockdaboot: Fixed

Makefile
```diff
@@ -208,3 +208,40 @@ check-clean-work-tree:
...
+
+# Install python packages into the virtual environment.
+$(PYTOOLS)/%: $(PYTOOLS)
+	@$(DOCKERPY) $(PIP) install -r requirements.txt
+
+CODESPELL = $(PYTOOLS)/codespell
+$(CODESPELL): PACKAGE=codespell
+
+.PHONY: codespell
+codespell: $(CODESPELL)
```
RonFed, line 246: Maybe worth adding the `codespell` as dependency for the `precommit` target

pellared (approved):

Makefile
rockdaboot: Thanks, done

grcevski (approved):
LGTM!

PR #888, opened 2024-06-13
HTTP client: Use Host field in URL if the Request Host is not present

RonFed:
From the `net/http` documentation about the `Request` struct:

> 	// For client requests, the URL's Host specifies the server to
	// connect to, while the Request's Host field optionally
	// specifies the Host header value to send in the HTTP
	// request.
	URL *url.URL

> 	// For client requests, Host optionally overrides the Host
	// header to send. If empty, the Request.Write method uses
	// the value of URL.Host. Host may contain an international
	// domain name.
	Host string

Our current implementation can cause the host field to not be collected properly in some cases.
This can be reproduced using our e2e test by changing
```
resp, err := http.Get("http://user@localhost:8080/hello/42?query=true#fragment")
```
to
```
url, _ := url.Parse("http://user@localhost:8080/hello/42?query=true#fragment")
req := &http.Request{
  Method: http.MethodGet,
  URL:    url,
  }
req = req.WithContext(context.Background())
resp, err := http.DefaultClient.Do(req)
```


edeNFed (approved):

damemi (approved):
Thanks @RonFed, I actually wasn't sure if there was a difference between the URL Host and Request Host in https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/790 so glad you looked into it!

PR #892, opened 2024-06-19
Add build constraints for linux/non-linux systems

RonFed:
This PR adds build constraints on files that contain Linux-specific logic or constants.
This allows running tests locally such as controller and manager on non-Linux systems.


damemi:
@RonFed nit but looks like a couple of these new files are missing newlines at the end fyi

edeNFed (approved):

Issue #334, opened 2023-09-17
Performance Testing & Benchmarking Go Auto Instrumentation


dineshg13:
When go auto instrumentation, we need to make performance trade-offs very clear to the end user.  
We need to run performance tests to measure the throughput of the go auto instrumentation supports. We have two knobs that we can tweak 
1. [Incoming Events buffer](https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/b1126ad1977e8af7c561c64a4c8595188b329f27/internal/pkg/instrumentors/manager.go#L52)
2. [Memory size](https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/b1126ad1977e8af7c561c64a4c8595188b329f27/internal/pkg/process/analyze.go#L32)

We should derive recommended values for the above and also make them configurable.

RonFed:
Related to this topic but more on the ebpf side.
As a general note, I think the ebpf code performance has a more "direct" impact on the probed code relative to the performance of our Go code.
Currently, we use perf buffer to transfer the events from ebpf to the user code in Go.
* I think that we should switch to ring buffer instead of perf buffer as it provides better performance in almost all scenarios as explained in [this blog post](https://nakryiko.com/posts/bpf-ringbuf/) and in more detail in [this patch](https://patchwork.ozlabs.org/project/netdev/patch/20200529075424.3139988-5-andriin@fb.com/)
* A key take-away from the above links is that the biggest performance impact by using the ring/perf buffer is the process in which the kernel signals the user code (blocking on reading from the buffer) that an event is ready. Today we signal for each event. In practice, a much more efficient approach would be to not signal each time we push an event to the buffer, but take the 'sampled' approach discussed in the links - to wake the Go user code every X events, where X may be a configurable parameter. Or for example, X can be a percentage of the buffer - if more than X% of the buffer is full, signal.
* I did some basic testing using `bpftool prog profile` and it looks like when we signal from ebpf code to the user code, the time spent in the uprboe increase drastically. Hence, implementing the sampled ring buffer approach sounds like a good idea to me.
* Another point to consider is what happens if the events throughput decreases and we wait a long time for X events - without signaling the Go code. For this case, I did this [Pr](https://github.com/cilium/ebpf/pull/1111) - which adds the ability to set a timeout for flushing the buffer by the user code in case it didn't get a signal. This timeout can be another configurable parameter.

PR #918, opened 2024-07-06
Refactor span output to a common function

RonFed:
Introduce the `output_span_event` function which is used by all probes to send spans to user space.
This has the following advantages:
* Spans that are not sampled will not be sent to user space, and the new function contains that check.
* The perf buffer is defined in a single place, similarly to what was done in user space in #899. This allows for future improvements such as using ring buffer instead of perf buffer if the kernel supports it (as suggested in https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/334#issuecomment-1727796027).


edeNFed (approved):

internal/pkg/instrumentation/bpf/github.com/segmentio/kafka-go/producer/bpf/probe.bpf.c
```diff
@@ -201,10 +194,16 @@ int uprobe_WriteMessages(struct pt_regs *ctx) {
         // Optionally collect the topic, and always collect key
         collect_kafka_attributes(msg_ptr, &kafka_request->msgs[i], !global_topic);
         // Generate span id for each message
-        generate_random_bytes(kafka_request->msgs[i].SpanID, SPAN_ID_SIZE);
-        __builtin_memcpy(current_sc.SpanID, kafka_request->msgs[i].SpanID, SPAN_ID_SIZE);
+        if (i > 0) {
+            generate_random_bytes(kafka_request->msgs[i].sc.SpanID, SPAN_ID_SIZE);
```
MrAlias, line 198: Why not generate a span ID for the first message?

RonFed: The span context (including the span ID) is currently created for the first message before this loop. Hence for each message (except the first one) we generate a new span ID and copy the trace ID and trace flags from the first one.

In the next PR, I plan to do a similar refactor for a `start_span` function which will contain similar logic to what is done in https://github.com/open-telemetry/opentelemetry-go/blob/82184aa44bb892ad95d98e09196a7c331f17797e/sdk/trace/tracer.go#L65

MrAlias (approved):

PR #957, opened 2024-07-17
Update libbpf to v1.4.5

MrAlias:
- Add `update.sh` to manage updates of the `libbpf` copy
- Remove all unused files copied from libbpf
- Move modifications of `libbpf/bpf_tracing.h` to `arguments.h`
- Run `update.sh` to sync libbpf to v1.4.5


internal/include/libbpf/update.sh
RonFed, line 1: Do we need to call this in CI? or create an action for this?

damemi: +1 a CI check would be good to prevent accidental changes to the libbpf copy. Or @RonFed do you mean something to automate the update itself, like a dependabot action?

RonFed: I was thinking about something like dependabot, but preventing updates to the libbpf copy is also a good idea.

MrAlias: I think both suggestions are good. What do you both feel about addressing them in follow-up PRs?

I can create issues to track them if you agree.

RonFed: Sounds good to me.

MrAlias: - https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/960
- https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/961

RonFed (approved):

PR #980, opened 2024-07-30
Fix shutdown panic

### LINKED ISSUE #976: Panic: send on closed channel
### Describe the bug

When terminating the auto-instrumentation CLI using `^C` the following panic occurs:

```terminal
^Cpanic: send on closed channel

goroutine 43 [running]:
go.opentelemetry.io/auto/internal/pkg/instrumentation/probe.(*Base[...]).Run(0xe2b700, 0xc0000220c0)
	/home/tyler/go/src/go.opentelemetry.io/auto/internal/pkg/instrumentation/probe/probe.go:211 +0x2d1
go.opentelemetry.io/auto/internal/pkg/instrumentation.(*Manager).Run.func1({0xe23790?, 0xc0003ca500?})
	/home/tyler/go/src/go.opentelemetry.io/auto/internal/pkg/instrumentation/manager.go:153 +0x64
created by go.opentelemetry.io/auto/internal/pkg/instrumentation.(*Manager).Run in goroutine 1
	/home/tyler/go/src/go.opentelemetry.io/auto/internal/pkg/instrumentation/manager.go:151 +0x18a
```

### Environment

 - OS: Linux xi 6.10.2-arch1-1
 - Go Version: 1.22.5
 - Version: dfd43b3

### To Reproduce

1. Run the `user` service from https://github.com/MrAlias/otel-auto-demo using `go build -o $GOPATH/bin/user ./... && $GOPATH/bin/user`
2. Run the auto-instrumentation cli: `OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4318 OTEL_GO_AUTO_TARGET_EXE=$GOPATH/bin/user OTEL_SERVICE_NAME=user OTEL_GO_AUTO_GLOBAL=true sudo -E ./cli`
3. Exit the CLI with `<CTRL>C`

### Expected behavior

No panic and a clean shutdown

MrAlias:
Resolve #976 

- Remove the `Close` method from the `Manager`. The unified way to stop `Run` is to have the passed context be done.
- Do not hold the events channel or wait group as fields of a `Manager`. They are local objects scoped to the `Run` method, move them entirely to that scope.
- Wait for all probes to close in `Run` before closing the `chan` they send on. This prevents a panic from the send on a closed `chan`.


internal/pkg/instrumentation/manager.go
```diff
@@ -161,29 +161,26 @@ func (m *Manager) Run(ctx context.Context, target *process.TargetDetails) error
...
-		case <-m.done:
-			m.logger.V(1).Info("shutting down all probes due to signal")
+			m.logger.V(1).Info("Shutting down all probes")
 			err := m.cleanup(target)
-			m.closingErrors <- err
-			return nil
-		case e := <-m.incomingEvents:
+
+			wg.Wait()
+			close(eventCh)
```
RonFed, line 168: Maybe worth adding a comment here about the importance of closing this channel only after the probes are closed to avoid panic

internal/pkg/instrumentation/manager_test.go
```diff
@@ -186,3 +189,87 @@ func fakeManager(t *testing.T) *Manager {
 
 	return m
 }
+
+func TestRunStopping(t *testing.T) {
```
RonFed, line 193: Why not do this test for the `Instrumentation` object? That way the test will cover more code

instrumentation.go
```diff
@@ -57,6 +58,10 @@ type Instrumentation struct {
 	target   *process.TargetDetails
 	analyzer *process.Analyzer
 	manager  *instrumentation.Manager
+
+	stopMu  sync.Mutex
+	stop    context.CancelFunc
+	stopped chan struct{}
```
RonFed, line 64: nit: we could group these 3 fields into a "Closing" struct

RonFed (approved):

internal/pkg/instrumentation/manager_test.go
MrAlias: Yeah, that was the original plan. However, the `Instrumentation` type is in a different package and setting these testing overrides is not possible. We would need to create a "true" fake program to instrument or develop other testing override ability.

I'm planning to take a look at that in a follow-up, possibly. But didn't want to block this fix on that significant work.

edeNFed (approved):
lgtm @MrAlias can we merge this?

internal/pkg/instrumentation/manager_test.go
MrAlias: Tracking in https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/1006

instrumentation.go
MrAlias: I looked into this, but am not sure it is worth it. There is a lot of coordination between the setup and tear-down of these within the context of the `Intrumentation` methods that they are pretty tightly coupled. Without doing a more complete re-design on the signaling within the `Instrumentation`, this is probably going to be the best we can do.

I'm not opposed to that re-design. Just, I would want to address it in a separate issue.

Going to resolve this, and track the work in https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/1006.

Issue #229, opened 2023-07-21
Document supported environment variables


pellared:
We should have a dedicated page with all possible configuration settings. Example from .NET Automatic Instrumentation: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/main/docs/config.md

Here is how we can find some env vars coming from this repo: https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-go-instrumentation%20%22OTEL_&type=code

We need to remember to also document config which comes from the dependencies e.g. 
- https://pkg.go.dev/go.opentelemetry.io/contrib/exporters/autoexport  (only the ones related with tracing)
- https://pkg.go.dev/go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc
- https://pkg.go.dev/go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracehttp

I think these env vars are supported as well (documentation for these env vars: https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/):
- `OTEL_SPAN_ATTRIBUTE_COUNT_LIMIT`
- `OTEL_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT`
- `OTEL_SPAN_EVENT_COUNT_LIMIT`
- `OTEL_SPAN_LINK_COUNT_LIMIT`
- `OTEL_EVENT_ATTRIBUTE_COUNT_LIMIT`
- `OTEL_LINK_ATTRIBUTE_COUNT_LIMIT`
- ~~`OTEL_ATTRIBUTE_VALUE_LENGTH_LIMIT`~~ (IMO it is better not to document this env var as it is [not documented in OTel Go SDK](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/trace#NewSpanLimits) also other env vars offer more fine-grained control)
- ~~`OTEL_ATTRIBUTE_COUNT_LIMIT`~~ (IMO it is better not to document this env var as it is [not documented in OTel Go SDK](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/trace#NewSpanLimits) also other env vars offer more fine-grained control)

MA3CIN:
Bump - As a new user, this is still very much needed. Perhaps some info could be added in the "Configuration" section of the README?

damemi:
Also from #233:

[...]

PR #981, opened 2024-07-31
Add env var page to docs

pauljwil:
Fixes https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/229.


docs/configuration.md
```diff
@@ -0,0 +1,127 @@
+# Configuration
+
+## Configuration methods
```
pauljwil, line 3: @MrAlias Should we add content here akin the [.NET docs](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/main/docs/config.md#configuration-methods)? If so, can you provide rough content for me to add, or point to relevant resources?

pauljwil:
Forgot to format everything into markdown tables. Will submit another commit shortly.

docs/configuration.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Global settings
+
+| Environment variable        | Description                                                                | Default value |
+|-----------------------------|----------------------------------------------------------------------------|---------------|
+| `OTEL_GO_AUTO_TARGET_EXE`   | Sets the binary for the Go application to be instrumented.                 |               |
```
pauljwil, line 9: Everywhere the default value column is blank, I could not find a default value. @MrAlias Please supply these if you know them.

docs/configuration.md
```diff
@@ -0,0 +1,53 @@
+# Configuration
+
+## Configuration methods
+
+## Global settings
```
pauljwil, line 5: @MrAlias Let me know if there is a better way to structure the env vars into separate sections. I based these mostly on what I found in the .NET docs.

docs/configuration.md
pellared: The environmental variables listed below are used to configure the auto-instrumentation binary: https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/main/cli/main.go.

For the binary one can also set the logging level via `-logLevel` and enable recording telemetry from the OpenTelemetry default global implementation the global via `global-impl` CLI flags.

Users can also chose build their own Go automatic instrumentation binaries using https://pkg.go.dev/go.opentelemetry.io/auto and add support for more/other configurations.

docs/configuration.md
```diff
@@ -0,0 +1,53 @@
...
+
+| Environment variable        | Description                                                                | Default value |
+|-----------------------------|----------------------------------------------------------------------------|---------------|
+| `OTEL_GO_AUTO_TARGET_EXE`   | Sets the binary for the Go application to be instrumented.                 |               |
+| `OTEL_GO_AUTO_GLOBAL`       | Enables the OpenTelemetry global implementation.                          |               |
```
pellared, line 10: Here is a more accurate description: https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/d730a91aaa5485951d9f8b3883b64936d021e693/cli/main.go#L62

docs/configuration.md
pellared, line 9: `OTEL_GO_AUTO_TARGET_EXE` is a required environment variable (by default it is empty).

docs/configuration.md
pellared, line 10: Default: `false`

docs/configuration.md
```diff
@@ -0,0 +1,53 @@
...
+| `OTEL_EXPORTER_OTLP_TIMEOUT`                | Maximum time in milliseconds the OTLP exporter waits for each batch export.                                                                                                                                                                                                                                                                                                  | `"10000"`                   |
+| `OTEL_EXPORTER_OTLP_TRACES_TIMEOUT`         | Maximum time in milliseconds the OTLP exporter waits for each batch export. The value of this variable takes precedence over `OTEL_EXPORTER_OTLP_TIMEOUT`.                                                                                                                                                                                                                   | `"10000"`                   |
```
pellared, line 45:
```suggestion
| `OTEL_EXPORTER_OTLP_TRACES_TIMEOUT`         | Maximum time in milliseconds the OTLP exporter waits for each batch export. The value of this variable takes precedence over `OTEL_EXPORTER_OTLP_TIMEOUT`.                                                                                                                                                                                                                   | `10000`                   |
```

docs/configuration.md
```diff
@@ -0,0 +1,53 @@
...
+| `OTEL_EXPORTER_OTLP_TRACES_HEADERS`         | Key-value pairs used as gRPC metadata associated with gRPC requests. The value must be represented in a format matching the W3C Baggage HTTP Header Content Format, except that additional semi-colon delimited metadata is not supported. Example value: `"key1=value1,key2=value2"`. The value of this variable takes precedence over `OTEL_EXPORTER_OTLP_HEADERS`.           | `none`                      |
+| `OTEL_EXPORTER_OTLP_TIMEOUT`                | Maximum time in milliseconds the OTLP exporter waits for each batch export.                                                                                                                                                                                                                                                                                                  | `"10000"`                   |
```
pellared, line 44:
```suggestion
| `OTEL_EXPORTER_OTLP_TIMEOUT`                | Maximum time in milliseconds the OTLP exporter waits for each batch export.                                                                                                                                                                                                                                                                                                  | `10000`                   |
```

docs/configuration.md
```diff
@@ -0,0 +1,53 @@
...
+| `OTEL_EXPORTER_OTLP_CLIENT_KEY`             | The filepath to the client's private key to use for mTLS communication in PEM format.                                                                                                                                                                                                                                                                                       | `none`                      |
+| `OTEL_EXPORTER_OTLP_TRACES_CLIENT_KEY`      | The filepath to the client's private key to use for mTLS communication in PEM format. The value of this variable takes precedence over `OTEL_EXPORTER_OTLP_CLIENT_KEY`.                                                                                                                                                                                                     | `none`                      |
```
pellared, line 53: Default should be blank (unset). `none` is an actual possible value.

docs/configuration.md
```diff
@@ -0,0 +1,53 @@
...
+| `OTEL_EXPORTER_OTLP_HEADERS`                | Key-value pairs used as gRPC metadata associated with gRPC requests. The value must be represented in a format matching the W3C Baggage HTTP Header Content Format, except that additional semi-colon delimited metadata is not supported. Example value: `"key1=value1,key2=value2"`.                                                                                       | `none`                      |
+| `OTEL_EXPORTER_OTLP_TRACES_HEADERS`         | Key-value pairs used as gRPC metadata associated with gRPC requests. The value must be represented in a format matching the W3C Baggage HTTP Header Content Format, except that additional semi-colon delimited metadata is not supported. Example value: `"key1=value1,key2=value2"`. The value of this variable takes precedence over `OTEL_EXPORTER_OTLP_HEADERS`.           | `none`                      |
```
pellared, line 43: Default should be blank (unset). `none` is an actual possible value.

docs/configuration.md
```diff
@@ -0,0 +1,53 @@
...
+| `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL`        | Defines the OTLP exporter's transport protocol for the traces signal. Supported values: `grpc`, `http/protobuf`. The value of this environment variable takes precedence over `OTEL_EXPORTER_OTLP_PROTOCOL`.                                                                                                                                                                                                                                                             | `http/protobuf`             |
+| `OTEL_EXPORTER_OTLP_ENDPOINT`               | The endpoint where the exporter sends the telemetry data. The value must contain a host, and can additionally include a port, scheme, and path. You can use either `http` or `https`. The value must not contain a query string or fragment.                                                                                                                                | `https://localhost:4317`    |
```
pellared, line 38: Actually, I think the SDK docs are misleading. The default is `http://localhost:4318`  for http/protobuf protocol and `http://localhost:4317` for gRPC protocol (the connection is insecure by default).

docs/configuration.md
```diff
@@ -0,0 +1,53 @@
...
+| `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`        | The endpoint where the exporter sends the telemetry data. The value must contain a host, and can additionally include a port, scheme, and path. You can use either `http` or `https`. The value must not contain a query string or fragment. The value of this environment variable takes precedence over `OTEL_EXPORTER_OTLP_ENDPOINT`.                                  | `https://localhost:4317`    |
```
pellared, line 39: Actually, I think the SDK docs are misleading. The default is `http://localhost:4318/v1/traces`  for http/protobuf protocol and `http://localhost:4317` for gRPC protocol (the connection is insecure by default).

docs/configuration.md
pauljwil: Should `global-impl` have a dash at the beginning (`-global-impl`)?

pellared: Yes, it should.

docs/configuration.md
pellared: lgtm

docs/configuration.md
```diff
@@ -0,0 +1,57 @@
...
+Use the environment variables in the tables below to configure the [automatic instrumentation binary](https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/main/cli/main.go).
+
+Alternatively, you can add support for additional or different configurations by building your own Go automatic instrumentation binary using the [OpenTelemetry Go Automatic Instrumentation package](https://pkg.go.dev/go.opentelemetry.io/auto).
+
+## Global settings
+
+| Environment variable        | Description                                                                | Default value |
+|-----------------------------|----------------------------------------------------------------------------|---------------|
+| `OTEL_GO_AUTO_TARGET_EXE`   | Sets the binary for the Go application to be instrumented. Required.       | Unset         |
+| `OTEL_GO_AUTO_GLOBAL`       | Records telemetry from the OpenTelemetry default global implementation. As an alternative to using the environment variable, you can use the `-logLevel` CLI flag.    | `false`       |
```
pellared, line 14:
```suggestion
| `OTEL_GO_AUTO_GLOBAL`       | Records telemetry from the OpenTelemetry default global implementation. As an alternative to using the environment variable, you can use the `-global-impl` CLI flag.    | `false`       |
```

docs/configuration.md
```diff
@@ -0,0 +1,57 @@
...
+| `OTEL_LOG_LEVEL`            | Sets the log level. Supported values: `none`, `error`, `warn`, `info`, `debug`. As an alternative to using the environment variable, you can use the `global-impl` CLI flag. | `info`        |
```
pellared, line 15:
```suggestion
| `OTEL_LOG_LEVEL`            | Sets the log level. Supported values: `none`, `error`, `warn`, `info`, `debug`. As an alternative to using the environment variable, you can use the `-logLevel` CLI flag. | `info`        |
```

docs/configuration.md
```diff
@@ -0,0 +1,57 @@
...
+| `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL`        | Defines the OTLP exporter's transport protocol for the traces signal. Supported values: `grpc`, `http/protobuf`. The value of this environment variable takes precedence over `OTEL_EXPORTER_OTLP_PROTOCOL`.                                                                                                                                                                 | `http/protobuf`             |
+| `OTEL_EXPORTER_OTLP_ENDPOINT`               | The endpoint where the exporter sends the telemetry data. The value must contain a host, and can additionally include a port, scheme, and path. You can use either `http` or `https`. The value must not contain a query string or fragment.                                                                                                                                | http/protobuf: `https://localhost:4318`, gRPC: `http://localhost:4317`    |
```
pellared, line 42:
```suggestion
| `OTEL_EXPORTER_OTLP_ENDPOINT`               | The endpoint where the exporter sends the telemetry data. The value must contain a host, and can additionally include a port, scheme, and path. You can use either `http` or `https`. The value must not contain a query string or fragment.                                                                                                                                | http/protobuf: `https://localhost:4318`, grpc: `http://localhost:4317`    |
```

docs/configuration.md
```diff
@@ -0,0 +1,57 @@
...
+| `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`        | The endpoint where the exporter sends the telemetry data. The value must contain a host, and can additionally include a port, scheme, and path. You can use either `http` or `https`. The value must not contain a query string or fragment. The value of this environment variable takes precedence over `OTEL_EXPORTER_OTLP_ENDPOINT`.                                  | http/protobuf: `https://localhost:4318/v1/traces`, gRPC: `http://localhost:4317`    |
```
pellared, line 43:
```suggestion
| `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`        | The endpoint where the exporter sends the telemetry data. The value must contain a host, and can additionally include a port, scheme, and path. You can use either `http` or `https`. The value must not contain a query string or fragment. The value of this environment variable takes precedence over `OTEL_EXPORTER_OTLP_ENDPOINT`.                                  | http/protobuf: `https://localhost:4318/v1/traces`, grpc: `http://localhost:4317`    |
```

README.md
```diff
@@ -47,6 +47,10 @@ You can also try the [Tutorial](docs/tutorial) for a guide on setting up a sampl
 
 For technical and design info, see [How it works](docs/how-it-works.md).
 
+## Configuration
+
+For a list of supported environment variables that you can use to configure your Go automatic instrumentation, see [Configuration](docs/configuration.md).
```
pellared, line 52: I think this text can be updated as this is not only about env vars. Maybe some more generic?

E.g.

see [Configuration](docs/configuration.md)
```suggestion
See [configuration documentation](docs/configuration.md) to learn how to configure Go Automatic Instrumentation.
```

pellared:
@open-telemetry/go-instrumentation-approvers, PTAL

docs/configuration.md
RonFed, line 1: Some of these environment variables are documented in https://opentelemetry.io/docs/languages/sdk-configuration/otlp-exporter or in https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables.
Maybe it will be better to point to those pages when applicable and keep here only the configuration specific to this project? If one of the env vars documentation is changed having a pointer to the official OTel docs will reduce the chance the docs here will become not up to date.

pellared: 1. Those env vars are mostly stable
2. Some of the env vars are not supported
3. Having everything in one page makes it simpler for the users
4. The spec and actual implementation can get out of sync

For reference in opentelemetry.io some existing automatic instrumentations are having dedicated configuration pages: 
- https://opentelemetry.io/docs/zero-code/net/configuration/
- https://opentelemetry.io/docs/languages/java/configuration/

For .NET, we are maintaining https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/main/docs/config.md and syncing https://opentelemetry.io/docs/zero-code/net/configuration during the release process.

docs/configuration.md
```diff
@@ -0,0 +1,57 @@
+# Configuration
+
+## Configuration methods
+
+Use the environment variables in the tables below to configure the [automatic instrumentation binary](https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/main/cli/main.go).
+
+Alternatively, you can add support for additional or different configurations by building your own Go automatic instrumentation binary using the [OpenTelemetry Go Automatic Instrumentation package](https://pkg.go.dev/go.opentelemetry.io/auto).
+
+## Global settings
```
RonFed, line 9: Maybe worth adding a section about instrumentation specific configuration? Currently the only example we have that I can think of is to include the SQL query in database/sql. But we are planning to have more in the future. If it is out of scope for this PR it's fine.

pellared: Can you share what config option is missing so that it can be added (here or separate PR)?

docs/configuration.md
```diff
@@ -0,0 +1,57 @@
...
+
+| Environment variable        | Description                                                                | Default value |
+|-----------------------------|----------------------------------------------------------------------------|---------------|
+| `OTEL_GO_AUTO_TARGET_EXE`   | Sets the binary for the Go application to be instrumented. Required.       | Unset         |
```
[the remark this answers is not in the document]
pellared: Is the `Configuration methods` section not describing it already?

docs/configuration.md
RonFed: `OTEL_GO_AUTO_INCLUDE_DB_STATEMENT`

docs/configuration.md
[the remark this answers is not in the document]
RonFed: Right, just the word `Required` maybe should replaced/removed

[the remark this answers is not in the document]
pellared: @RonFed 

> Use the environment variables in the tables below to configure the [automatic instrumentation binary](https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/main/cli/main.go).

and 

> It is only required when running the instrumentation as a standalone binary. 

Therefore, I think it is better to mention that it is required. Or am I missing something?

docs/configuration.md
RonFed: I think it might be clearer to have a separate section for instrumentation configuration (as the above option is not configuring the exporter but the instrumentation itself - collecting an extra attribute). Currently, the above env var is the only config option we have under that category so we can change it in the future once we'll have more options.

pauljwil: IMO, it would be fine to create the section now with just the one entry. It might look a little funny to have just one config option, but I'd rather that than bury the env var in the wrong place and potentially confuse people about what it's for. But I don't have a ton of context, so we can go with whatever you recommend.

pellared: You can use https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/main/docs/config.md#instrumentation-options for inspiration.

pauljwil: I moved the env var to a new Instrumentation options section. As we fill in the new section with more env vars, we can add similar structure and detail to what @pellared linked to in the dotnet docs.

MrAlias (approved):

PR #982, opened 2024-08-04
Add sampling config for instrumentation

RonFed:
This PR includes the high-level code handling the sampling configuration.


RonFed:
@MrAlias @damemi can you please have a look?

internal/pkg/instrumentation/probe/sampling/sampling.go
```diff
@@ -0,0 +1,123 @@
...
+	// TODO
+}
+
+type TraceIDRatioConfig struct {
+	// samplingRateNumerator is the numerator of the sampling rate.
+	// see samplingRateDenominator for more information.
+	samplingRateNumerator uint64
+}
+
+func NewTraceIDRationConfig(ratio float64) (TraceIDRatioConfig, error) {
```
damemi, line 33:
```suggestion
func NewTraceIDRatioConfig(ratio float64) (TraceIDRatioConfig, error) {
```

edeNFed (approved):

sampling.go
```diff
@@ -0,0 +1,303 @@
...
+		RemoteSampled:    AlwaysOn{},
+		RemoteNotSampled: AlwaysOff{},
+		LocalSampled:     AlwaysOn{},
+		LocalNotSampled:  AlwaysOff{},
+	}
+
+	samplerName, ok := lookupEnv(tracesSamplerKey)
+	if !ok {
+		return defaultSampler, nil
+	}
```
MrAlias, line 255: To match the other environment variable resolution functionality, and to not perform wasted work, the default sampler should not be returned when the sampler environment variable is not provided.

```suggestion
	samplerName, ok := lookupEnv(tracesSamplerKey)
	if !ok {
		return nil, nil
	}
	
	defaultSampler := ParentBased{
		Root:             AlwaysOn{},
		RemoteSampled:    AlwaysOn{},
		RemoteNotSampled: AlwaysOff{},
		LocalSampled:     AlwaysOn{},
		LocalNotSampled:  AlwaysOff{},
	}
```

instrumentation.go
```diff
@@ -401,6 +406,17 @@ func WithEnv() InstrumentationOption {
 
 			err = errors.Join(err, e)
 		}
+		if s, err := newSamplerFromEnv(); err == nil {
```
MrAlias, line 409: This is implicitly setting the default. That needs to be left for `newInstConfig` (see suggestion for `newSamplerFromEnv`).

Additionally, this is shadowing `err` from the outer scope. If `err` is not `nil` in this scope it will not be returned in the outer scope function. You may want to refactor to something like this:

```suggestion
		if s, e := newSamplerFromEnv(); e != nil {
			err = errors.Join(err, e)
		} else if s != nil {
```

instrumentation.go
```diff
@@ -401,6 +406,17 @@ func WithEnv() InstrumentationOption {
...
+			err := s.validate()
+			if err != nil {
+				return c, err
+			}
+			cfg, err := s.convert()
+			if err != nil {
+				return c, err
```
MrAlias, line 416: This is dropping all the combined errors from above. The returned errors need to be joined into the already existing errors instead.

internal/pkg/instrumentation/probe/sampling/sampling.go
```diff
@@ -0,0 +1,123 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package sampling
+
+import "github.com/cilium/ebpf"
+
+type SamplerType uint64
+
+// SamplerType defines the type of a sampler.
```
MrAlias, line 10:
```suggestion
// SamplerType defines the type of a sampler.
type SamplerType uint64

```

internal/pkg/instrumentation/probe/sampling/sampling.go
```diff
@@ -0,0 +1,123 @@
...
+const (
+	// OpenTelemetry spec-defined samplers.
+
```
MrAlias, line 13:
```suggestion
// OpenTelemetry spec-defined samplers.
const (
```

internal/pkg/instrumentation/probe/sampling/sampling.go
```diff
@@ -0,0 +1,123 @@
...
+	SamplerTraceIDRatio
+	SamplerParentBased
+
+	// Custom samplers TODO.
+)
+
+// Manager is used to configure the samplers used by eBPF.
+type Manager struct {
+	// TODO
+}
```
MrAlias, line 25: This is not used in this PR. Can you remove it, `NewSamplingManager`, and the use of `NewSamplingManager` from this change? In the place where `NewSamplingManager` is called in the base probe, can you leave a comment explaining how sampling is planned to be setup and used instead?

internal/pkg/instrumentation/probe/sampling/sampling.go
```diff
@@ -0,0 +1,123 @@
...
+	LocalNotSampled  SamplerID
+}
+
+// the following are constants which are used by the eBPF code.
+// they should be kept in sync with the definitions there.
+const (
+	// since eBPF does not support floating point arithmetic, we use a rational number to represent the ratio.
+	// the denominator is fixed and the numerator is used to represent the ratio.
+	// This value can limit the precision of the sampling rate, hence setting it to a high value should be enough in terms of precision.
+	samplingRateDenominator = 1e9
```
MrAlias, line 89: Why not use the max `uint32` value (`0xFFFFFFFF`) here?

instrumentation.go
RonFed: Right. Updated

sampling.go
```diff
@@ -0,0 +1,303 @@
...
+	if err := validateParentBasedComponent(p.RemoteSampled); err != nil {
+		return err
+	}
+	if err := validateParentBasedComponent(p.RemoteNotSampled); err != nil {
+		return err
+	}
+	if err := validateParentBasedComponent(p.LocalSampled); err != nil {
+		return err
+	}
+	return validateParentBasedComponent(p.LocalNotSampled)
```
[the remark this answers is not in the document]
RonFed: done

sampling.go
RonFed: Updated

internal/pkg/instrumentation/probe/sampling/sampling.go
RonFed: Done

internal/pkg/instrumentation/probe/sampling/sampling.go
RonFed: It felt more intuitive to me since I think most use cases will be in "decimal" fractions (0.1, 0.001. etc'). In addition, it might lead to fewer rounding errors (since it is integer division) but not sure about that.

internal/pkg/instrumentation/probe/sampling/sampling.go
RonFed: Fixed

sampling.go
```diff
@@ -0,0 +1,303 @@
...
+const (
+	tracesSamplerKey    = "OTEL_TRACES_SAMPLER"
+	tracesSamplerArgKey = "OTEL_TRACES_SAMPLER_ARG"
+
+	samplerNameAlwaysOn                = "always_on"
+	samplerNameAlwaysOff               = "always_off"
+	samplerNameTraceIDRatio            = "traceidratio"
+	samplerNameParentBasedAlwaysOn     = "parentbased_always_on"
+	samplerNameParsedBasedAlwaysOff    = "parentbased_always_off"
+	samplerNameParentBasedTraceIDRatio = "parentbased_traceidratio"
```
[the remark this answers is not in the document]
RonFed: Opened #1027

instrumentation.go
```diff
@@ -401,6 +406,17 @@ func WithEnv() InstrumentationOption {
...
 			err = errors.Join(err, e)
 		}
+		if s, e := newSamplerFromEnv(); e != nil {
+			err = errors.Join(err, e)
+		} else if s != nil {
+			e := s.validate()
+			if e == nil {
+				if cfg, e := s.convert(); e == nil {
+					c.samplingConfig = cfg
+				}
```
MrAlias, line 416: This error held in `e` will be dropped. It is in a sub-scope that shadows the variable in the outer scope.

```suggestion
				c.samplingConfig, e = s.convert()
```

internal/pkg/instrumentation/probe/sampling/sampling.go
RonFed: Updated

internal/pkg/instrumentation/probe/probe.go
```diff
@@ -97,7 +98,7 @@ func (i *Base[BPFObj, BPFEvent]) Spec() (*ebpf.CollectionSpec, error) {
 }
 
 // Load loads all instrumentation offsets.
-func (i *Base[BPFObj, BPFEvent]) Load(exec *link.Executable, td *process.TargetDetails) error {
+func (i *Base[BPFObj, BPFEvent]) Load(exec *link.Executable, td *process.TargetDetails, sampler config.Sampler) error {
```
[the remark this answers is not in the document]
RonFed: I hope that in the near future the sampling config will be able to change during the lifetime of a probe (I did some preparation for this using the `config.Provider`).
The Probe API will need to change to support this (adding some `Config` function to change it in runtime).
The `Load` function seemed like a reasonable place to pass an initial configuration (a configuration that may include more things than just sampling).
What do you think?

.github/dependabot.yml
```diff
@@ -10,6 +10,15 @@ updates:
...
       day: sunday
+  - package-ecosystem: docker
+    directory: /.git/logs/refs/heads
+    labels:
+      - dependencies
+      - docker
+      - Skip Changelog
+    schedule:
+      interval: weekly
+      day: sunday
```
MrAlias, line 21: This doesn't seem right. Why are we adding a docker updated directive for our git reflog?

internal/pkg/instrumentation/probe/probe.go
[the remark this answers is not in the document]
damemi: Ah, that sounds good to me then. This isn't public yet, so we can break it, but wdyt about adding a comment here that we plan to have more options passed here?

.github/dependabot.yml
RonFed: Ah that is a mistake I'll remove that. I ran make docker-pre commit and it sometimes add this for some reason.

internal/pkg/instrumentation/probe/probe.go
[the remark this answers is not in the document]
RonFed: I added a comment about this in the `Load` function in the `Probe` interface definition

instrumentation.go
RonFed: This part was changed (the conversion is not done here)

RonFed:
@damemi @MrAlias I addressed your comments. This is ready for another review.

config/sampling.go
```diff
@@ -0,0 +1,298 @@
...
+		}
+		defaultSampler.Root = TraceIDRatio{Fraction: ratio}
+		return defaultSampler, nil
+	default:
+		return nil, errors.New("unknown sampler name")
+	}
+}
+
+// ConvertSamplerToConfig converts a Sampler its internal representation.
+func ConvertSamplerToConfig(s Sampler) (*sampling.Config, error) {
```
MrAlias, line 290: This needs to be unexported. It should not be used by end-users.

```suggestion
// convertSamplerToConfig converts a Sampler its internal representation.
func convertSamplerToConfig(s Sampler) (*sampling.Config, error) {
```

(broken suggestion based on updates needed elsewhere)

MrAlias: This change will need the `config` package to be flattened into the `auto` package or the `Sampler` interface will need to be defined and `internal` and aliased here.

I'm in favor of flattening `config` into `auto` (but I haven't completely reviewed that proposal).

RonFed: The reason I created the config package is to avoid cyclic dependencies which happened in the manager (which is aware of the Provider interface from the config package).

edeNFed: As discussed during the SIG meeting (and later with @RonFed)
I think we can merge this PR and do this change in a separate PR

MrAlias (approved):
Approving with the intent to flatten this package in a follow-up PR.

MrAlias:
`golang:1.22.7` just released. We need the new docker image and an offset update to unblock this.

PR #984, opened 2024-08-04
Reduce e2e flakiness

### LINKED ISSUE #972: Reduce e2e-test flakiness
Some of the e2e tests (mainly the `kafka-go` one) are flaky. For example: https://github.com/open-telemetry/opentelemetry-go-instrumentation/actions/runs/10131243094/job/28013685717.
The cause for this failure is inconsistancy in the batch being exported. Since the default batch timeout is 5s, there is a probability of the spans being sent in 2 different batches/ single batch - which cause the differen JSON resullt in the test.

Proposed solutions:
* The more straightforwrd solution is to define `sample-job.yaml` per e2e test. Each test will be able to configure the batch span processor using `OTEL_BSP_MAX_EXPORT_BATCH_SIZE` and `OTEL_BSP_SCHEDULE_DELAY` accoridng to the expected number of spans in the test.
* A better solution might be to add a `number_of_spans` field to the test matrix defined in https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/c80c3ab2d6eb4edafc676ab492bb7227e0a92cb5/.github/workflows/kind.yml#L14.
This field will be used to configure OTEL_BSP_MAX_EXPORT_BATCH_SIZE per test. This will require to pass this config to `sample-job`. It might be possible through a build argument to the Dockerfile, and there are maybe more options.

RonFed:
Resolves #972

### DIFF
--- a/.github/workflows/e2e/k8s/e2e-chart/templates/sample-job.yml
+++ b/.github/workflows/e2e/k8s/e2e-chart/templates/sample-job.yml
@@ -38,6 +38,10 @@ spec:
           value: "tracecontext,baggage"
         - name: OTEL_GO_AUTO_INCLUDE_DB_STATEMENT
           value: "true"
+        - name: OTEL_BSP_SCHEDULE_DELAY
+          value: "60000"
+        - name: OTEL_BSP_MAX_EXPORT_BATCH_SIZE
+          value: {{ .Values.expectedSpansCount | quote }}
         resources: {}
         securityContext:
           runAsUser: 0
--- a/.github/workflows/e2e/k8s/e2e-chart/.helmignore
+++ b/.github/workflows/e2e/k8s/e2e-chart/.helmignore
@@ -1,0 +1,23 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
+# Common VCS dirs
+.git/
+.gitignore
+.bzr/
+.bzrignore
+.hg/
+.hgignore
+.svn/
+# Common backup files
+*.swp
+*.bak
+*.tmp
+*.orig
+*~
+# Various IDEs
+.project
+.idea/
+*.tmproj
+.vscode/
--- a/.github/workflows/e2e/k8s/e2e-chart/Chart.yaml
+++ b/.github/workflows/e2e/k8s/e2e-chart/Chart.yaml
@@ -1,0 +1,24 @@
+apiVersion: v2
+name: e2e-chart
+description: A Helm chart for Kubernetes
+
+# A chart can be either an 'application' or a 'library' chart.
+#
+# Application charts are a collection of templates that can be packaged into versioned archives
+# to be deployed.
+#
+# Library charts provide useful utilities or functions for the chart developer. They're included as
+# a dependency of application charts to inject those utilities and functions into the rendering
+# pipeline. Library charts do not define any templates and therefore cannot be deployed.
+type: application
+
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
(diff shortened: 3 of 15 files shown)


.github/workflows/e2e/k8s/e2e-chart/templates/sample-job.yml
```diff
@@ -38,6 +38,10 @@ spec:
           value: "tracecontext,baggage"
         - name: OTEL_GO_AUTO_INCLUDE_DB_STATEMENT
           value: "true"
+        - name: OTEL_BSP_SCHEDULE_DELAY
+          value: "60000"
+        - name: OTEL_BSP_MAX_EXPORT_BATCH_SIZE
+          value: {{ .Values.expectedSpansCount | quote }}
```
MrAlias, line 44: I thought the plan was to leave this value alone? Setting this value only reduces it so far. If the point of this change is to fix the flakiness, leaving it where it is will avoid changes like aecea372c91677d968e1b7b4bab90d211d813748

RonFed: I do think that this will reduce the flakiness. 
The number of spans in a test is not something I expect to be changed often.
The configuration in this PR as I understand it allows to specify number of spans in a test and if we don't get that amount after a minute it will fail.

MrAlias: But the test doesn't terminate when a batch is exported, it stops when the job completes. The spans are flushed regardless of this setting. If we export 2 spans but the is set to 3 then the job will stop, the SDK will flush the batch processor, and then the evaluation will run as normal.

If we want to test missing spans we would need to do that in the bats validation.

This change will only affect things when we send more spans than the configured value. It will split things into two batches.

RonFed: I tried the approach we talked about. It seems we are not calling `tracerProvider.Shutdown`. I tried to add this during the manager closing phase and it helped in exporting the spans.
I am still not 100% sure about this not adding flakiness, since we are also using a file exporter in the e2e-test to create the JSON file, so even if the BSP is flushed there still may be a chance of the file not being ready?
@damemi

damemi: @RonFed you could try configuring the batch exporter in the collector's config https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/main/.github/workflows/e2e/k8s/collector-helm-values.yml

afaik the Go SDK should automatically flush when it receives a sigterm like we're sending, @MrAlias is that right? I don't think a manual call to Shutdown is necessary in that case

MrAlias: > afaik the Go SDK should automatically flush when it receives a sigterm like we're sending, @MrAlias is that right? I don't think a manual call to Shutdown is necessary in that case

No, an explicit call to `Shutdown` or `ForceFlush` is necessary. There is no signal capture in the SDK.

FWIW, if we weren't calling `Shutdown` on the SDK when we were shutting down, we should definitely change that.

damemi: @MrAlias ahh, that was an assumption I had when writing the e2es

MrAlias: @MrAlias is going to open a PR to add a shutdown to the opentelemetry.Controller and make sure that flushes the TracerProvicer.

That should unblock this.

MrAlias: https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/1028

RonFed:
Replaces by #1029

Issue #1008, opened 2024-08-17
markdown typo error in `docs/getting-started.md`


minimAluminiumalism:
```sh
  sudo OTEL_GO_AUTO_TARGET_EXE=/home/bin/service_executable OTEL_SERVICE_NAME=my_service OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4318 ./otel-go-instrumentation`
  ```
remove redundant ``` ` ``` behind `otel-go-instrumentation`.

minimAluminiumalism:
I've fixed this and add more instructions in https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/1009, please review it

PR #1009, opened 2024-08-17
fix typo err and add env ref

minimAluminiumalism:
fix typo err in `docs/getting-started.md` and add reference for `environment` in `docker-compose.yaml`.

Fixes https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/1008


docs/getting-started.md
```diff
@@ -61,12 +61,15 @@ network, a shared volume, and a service for the application.
       environment:
         - OTEL_EXPORTER_OTLP_ENDPOINT=http://<address_in_docker_network>:4318
         - OTEL_GO_AUTO_TARGET_EXE=<location_of_target_application_binary>
+        - OTEL_RESOURCE_ATTRIBUTES=key1=value1
```
damemi, line 64: is this required to run the agent? If not, I think it is covered by the "see more variables" link below. It would be good if this example stays minimal

damemi (approved):

docs/getting-started.md
minimAluminiumalism: Ok, I've removed that.

MrAlias (approved):

PR #1025, opened 2024-08-19
Expose OpenTelemetryController as interface option to Instrumentation

damemi:
Ref https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/1026

This does two things:

* Create an interface for OpenTelemetryController so it can be overwritten (collector receiver needs to push to next consumer as ptrace, rather than exporting otlp to an endpoint)
* Exposes the `probe.Event` struct (which is the parameter to the `Trace()` function in the OTelController interface)

### DIFF
--- a/pkg/opentelemetry/controller.go
+++ b/pkg/opentelemetry/controller.go
@@ -3,7 +3,11 @@
 
 package opentelemetry
 
-import "go.opentelemetry.io/auto/pkg/probe"
+import (
+	"context"
+
+	"go.opentelemetry.io/auto/pkg/probe"
+)
 
 // OpenTelemetryController is a controller that converts probe Events to
 // OpenTelemetry spans and exports them.
@@ -11,4 +15,7 @@ type OpenTelemetryController interface {
 	// Trace receives a probe.Event and handles conversion to OpenTelemetry
 	// format and exporting.
 	Trace(event *probe.Event)
+
+	// Shutdown shuts down the OpenTelemetry TracerProvider.
+	Shutdown(ctx context.Context) error
 }
--- a/pkg/opentelemetry/controller.go
+++ b/pkg/opentelemetry/controller.go
@@ -1,0 +1,10 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package opentelemetry
+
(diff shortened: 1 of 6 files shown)


RonFed (approved):
LGTM
As we talked in the SIG, I think we need to add some comments about the `probe.Event` struct and its fields, and maybe rename them since `SpanEvent` is a misleading name (maybe `SpanData`?)

pkg/probe/probe.go
```diff
@@ -0,0 +1,9 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package probe
+
+import "go.opentelemetry.io/auto/internal/pkg/instrumentation/probe"
+
+// Event represents a probed span event.
+type Event = probe.Event
```
MrAlias, line 9: I'm hesitant to export this type.

I just noticed that the start/end time are not actually unix timestamps, but the relative offset from startup.

I think we need to better evaluate the structure of this event type before we publish something others will rely on.  It might make the most sense if we actually use a proto representation of spans similar to the collector's `pdata` approach. That way we could avoid deserialization if an exporter speaks that protocol.

pkg/opentelemetry/controller.go
```diff
@@ -0,0 +1,14 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package opentelemetry
+
+import "go.opentelemetry.io/auto/pkg/probe"
+
+// OpenTelemetryController is a controller that converts probe Events to
+// OpenTelemetry spans and exports them.
+type OpenTelemetryController interface {
```
MrAlias, line 10: This will stutter. i.e. `opentelemetry.OpenTelemetryController`.

```suggestion
type Controller interface {
```

Though, I'm also not sure if we even need this `opentelemtry` pacakge. For my two cents, something like `auto.Controller` seems like a better name to me.

pkg/probe/probe.go
damemi: @MrAlias I pushed an update adding a wrapper struct that gives us a public abstraction from the internal types. Ideally I think we should eventually just take a good look at the internal types and make them public, but something like this could give us a nice bridge to that. Wdyt? Open to ideas

PR #1033, opened 2024-08-25
Instrument SpanContextFromContext to return the span context of the auto generated span

RonFed:


RonFed:
Open questions:
* Does this need to be under an opt-in flag? If so should it be the same flag for the global instrumentation? I am not sure since this is relevant for both "manual span which becomes auto" and for fully auto-generated spans like in the `net/http` test.
* Both tests are calling a helper function which calls `trace.SpanContextFromContext` and outputs the result in a format that will be easy for us to verify in `bats`. Should we create a module for common e2e stuff (I think this will require an extra PR to use this module in the tests).
* The probe in this PR is different from the current probes we have today because it does not generate events. This means that currently, the eBPF code must declare some dummy perf buffer that is not used, and the `probe.Run` function is not reading anything. Should we modify our probe interface or create a different implementation of it which will return from `Run` immediately and won't allocate a perf buffer in `Load`? Another option is to decompose the current interface to a base and then embed this interface in the current one.

internal/pkg/instrumentation/bpf/net/http/server/bpf/probe.bpf.c
```diff
@@ -230,6 +230,7 @@ int uprobe_serverHandler_ServeHTTP(struct pt_regs *ctx)
     start_span(&start_span_params);
 
     bpf_map_update_elem(&http_server_uprobes, &key, uprobe_data, 0);
+    bpf_printk("server go_context.data: 0x%llx", go_context.data);
```
RonFed, line 233: Remove

internal/test/e2e/nethttp/go.mod
```diff
@@ -1,3 +1,9 @@
 module go.opentelemetry.io/auto/internal/test/e2e/nethttp
 
-go 1.22.0
+go 1.22.6
+
+require go.opentelemetry.io/otel/trace v1.29.0
+
+require go.opentelemetry.io/otel v1.29.0 // indirect
+
+replace go.opentelemetry.io/auto/internal/test => ../..
```
RonFed, line 9: Remove

MrAlias:
Closing for now. The alternate SDK approach should make this unneeded. Will reopen if that changes.

Issue #172, opened 2023-05-16
Audit and update `google/golang/org/grpc` instrumentation to comply with RPC semantic conventions


MrAlias:
https://github.com/open-telemetry/opentelemetry-specification/blob/84a5fa3d1ef42caae8f99f1b431504d7cdb9fffe/specification/trace/semantic_conventions/rpc.md

damemi:
Latest RPC conventions: https://github.com/open-telemetry/semantic-conventions/blob/02ecf0c71e9fa74d09d81c48e04a132db2b7060b/docs/rpc/rpc-spans.md and specifically the gRPC conventions (which extend and override RPC): https://github.com/open-telemetry/semantic-conventions/blob/02ecf0c71e9fa74d09d81c48e04a132db2b7060b/docs/rpc/grpc.md

> ## Common remote procedure call conventions

> For outgoing requests, the `SpanKind` MUST be set to `CLIENT` and for incoming requests to `SERVER`.

Correct - [client](https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/9019008bbc20036eb6591372799d773eb89c8428/internal/pkg/instrumentation/bpf/google.golang.org/grpc/client/probe.go#L34)
Correct - [server](https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/9019008bbc20036eb6591372799d773eb89c8428/internal/pkg/instrumentation/bpf/google.golang.org/grpc/server/probe.go#L33)

>Remote procedure calls can only be represented with these semantic conventions, when the names of the called service and method are known and available.

>### Span name
>
>The _span name_ MUST be the full RPC method name formatted as:
>
>```
>$package.$service/$method
>```
>
>(where $service MUST NOT contain dots and $method MUST NOT contain slashes)

We don’t seem to currently fit this in our example, see the [gRPC test spans](https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/9019008bbc20036eb6591372799d773eb89c8428/internal/test/e2e/grpc/traces.json#L111), which for example set `name` to `/helloworld.Greeter/SayHello` which has a leading `/`. However, this is coming directly from the upstream [example code](https://github.com/grpc/grpc-go/blob/4ed81800b06dd7c587dd42830a780d2dad6812f4/examples/helloworld/helloworld/helloworld_grpc.pb.go#L36) which sets the full method name to that exact string.

[...]

PR #1044, opened 2024-08-27
Instrument gRPC status code

damemi:
Ref #172


internal/pkg/instrumentation/bpf/google.golang.org/grpc/client/bpf/probe.bpf.c
```diff
@@ -105,7 +108,27 @@ int uprobe_ClientConn_Invoke(struct pt_regs *ctx)
...
+int uprobe_ClientConn_Invoke_Returns(struct pt_regs *ctx) {
+    struct go_iface go_context = {0};
+    get_Go_context(ctx, 3, 0, true, &go_context);
+    void *key = get_consistent_key(ctx, go_context.data);
+    struct grpc_request_t *grpc_span = bpf_map_lookup_elem(&grpc_events, &key);
+    if (grpc_span == NULL) {
+        bpf_printk("event is NULL in ret probe");
+        return 0;
+    }
+    if (is_register_abi()) {
```
MrAlias, line 121: Do we plan to handle the alternate code path at some point? Or are we just not planning to support the case for pre-Go 1.17?

damemi: removed this

internal/pkg/instrumentation/bpf/google.golang.org/grpc/client/bpf/probe.bpf.c
```diff
@@ -105,7 +108,34 @@ int uprobe_ClientConn_Invoke(struct pt_regs *ctx)
...
+
+    // Getting the returned response (error)
+    void *resp_ptr = get_argument(ctx, 1);
+    void *status_ptr = 0;
+    // get `s` field from Status object pointer (from returned error interface)
+    // (*Status)error{*s}
+    bpf_probe_read(&status_ptr, sizeof(status_ptr), (void *)(resp_ptr+status_s_pos));
+    // Get status code from Status.s pointer
+    // (*s)struct{Code}
+    bpf_probe_read(&grpc_span->status_code, sizeof(grpc_span->status_code), (void *)(status_ptr + status_code_pos));
```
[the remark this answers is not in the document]
RonFed: I think the problem is in this line:
`void *resp_ptr = get_argument(ctx, 1);`
Since we are interested in the data pointer of the interface, we should read the second returned argument (the firs one is the `itab` pointer)

https://github.com/golang/go/blob/aeac0b6cbfb42bc9c9301913a191bb09454d316a/src/runtime/runtime2.go#L178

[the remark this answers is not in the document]
damemi: Thanks, so would that be `void *resp_ptr = get_argument(ctx, 2)`? I couldn't figure out how we determine the argument index in the other ebpf functions, so I tried brute-forcing this with 1-9 and got the same kind of results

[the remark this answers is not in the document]
RonFed: Yes, other than that the code looks fine to me.
* When reading the `status_ptr` worth checking that the pointer we read is not NILL (if it is, avoid reading the code). I assume that in a happy case, a nil error will be returned.
* Maybe worth considering using `bpf_probe_read_user` instead of `bpf_probe_read` as it is more recommended for these cases but I'm not sure it will make a difference in terms of correctness.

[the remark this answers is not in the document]
damemi: Hm still no luck, I tried checking the 2nd argument and using bpf_probe_read_user in https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/1044/commits/540ff896453c922afe3e30066f64fdeacdfc0d88 but that's just getting 0 for both success and error. I logged the error message from the test app just to be sure and it is returning "Unavailable" which should be gRPC code `14`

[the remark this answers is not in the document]
RonFed: Try and print the status code value in epbf using `bpf_printk`. maybe the int value is not getting read properly by the go code (alignment might be an issue)

internal/pkg/instrumentation/bpf/google.golang.org/grpc/client/bpf/probe.bpf.c
```diff
@@ -105,7 +109,46 @@ int uprobe_ClientConn_Invoke(struct pt_regs *ctx)
...
+    // the `error` interface concrete type here is a gRPC `internal.Error` struct
+    // type Error struct {
+    //   s *Status
+    // }
+    // The `Error` struct embeds a `Status` proto object
+    // type Status struct {
+    //   s *Status
+    // }
+    // The `Status` proto object contains a `Code` int32 field, which is what we want
+    void *resp_ptr = get_argument(ctx, 2);
```
MrAlias, line 137: How do we check if the returned error is `nil`?

internal/pkg/instrumentation/bpf/google.golang.org/grpc/client/probe.go
```diff
@@ -138,6 +154,7 @@ func convertEvent(e *event) []*probe.SpanEvent {
 			SpanContext:       &sc,
 			ParentSpanContext: pscPtr,
 			TracerSchema:      semconv.SchemaURL,
+			Status:            probe.Status{Code: codes.Code(int(e.StatusCode))},
```
MrAlias, line 157: This should only set the status for certain error status codes: https://github.com/open-telemetry/semantic-conventions/blob/5eb37f706894718493af7a668971c651acb80812/docs/rpc/grpc.md#grpc-status

And we will need to set it to `codes.Error`.

internal/test/e2e/grpc/traces.json
```diff
@@ -113,6 +119,47 @@
...
+                    "intValue": "14"
+                  }
+                }
+              ],
+              "flags": 768,
+              "kind": 3,
+              "name": "/helloworld.Greeter/SayHello",
+              "parentSpanId": "xxxxx",
+              "spanId": "xxxxx",
+              "status": {},
```
RonFed, line 161: Should this be an error status?

internal/pkg/instrumentation/bpf/google.golang.org/grpc/client/bpf/probe.bpf.c
```diff
@@ -18,6 +18,7 @@ struct grpc_request_t
     BASE_SPAN_PROPERTIES
     char method[MAX_SIZE];
     char target[MAX_SIZE];
+    int status_code;
```
RonFed, line 21: nit:  `s32` or `u32` as the type is clearer.

internal/pkg/instrumentation/bpf/google.golang.org/grpc/client/probe.go
damemi: Thanks, updated this. Looks like client spans are `codes.Error` for anything besides success which is easy here at least hah

damemi:
Updated the PR description and changelog to state that this is just for client spans. Looks like server spans will be a pretty different approach (reading [handleStream](https://github.com/grpc/grpc-go/blob/00d3ec8c71a7ba0f9c755881f0f7147eba5e814c/server.go#L1740)) so I'll do a separate PR for that

MrAlias (approved):

internal/pkg/instrumentation/bpf/google.golang.org/grpc/client/bpf/probe.bpf.c
```diff
@@ -105,7 +109,51 @@ int uprobe_ClientConn_Invoke(struct pt_regs *ctx)
...
+    bpf_probe_read_user(&status_ptr, sizeof(status_ptr), (void *)(resp_ptr+error_status_pos));
+    // get `s` field from Status object pointer
+    void *s_ptr = 0;
+    bpf_probe_read_user(&s_ptr, sizeof(s_ptr), (void *)(status_ptr + status_s_pos));
+    // Get status code from Status.s pointer
+    bpf_probe_read_user(&grpc_span->status_code, sizeof(grpc_span->status_code), (void *)(s_ptr + status_code_pos));
+
+done:
+    grpc_span->end_time = bpf_ktime_get_ns();
+    output_span_event(ctx, grpc_span, sizeof(*grpc_span), &grpc_span->sc);
```
RonFed, line 153: I think we should add a call to `stop_tracking_span` here. See https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/873#issuecomment-2178567698

grcevski (approved):
LGTM! Very cool!

internal/pkg/instrumentation/bpf/google.golang.org/grpc/client/bpf/probe.bpf.c
damemi: done

Issue #954, opened 2024-07-16
Custom SDK implementation


MrAlias:
Initial support for interoperability with manual instrumentation was added in https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/523. This approach though has issues when trying to implement methods like [`IsRecording`](https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/537), [`RecordError`](https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/540), and [`AddEvent`](https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/541).

### Provide our own SDK

An alternate solution for this interoperability was proposed in https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/352#issuecomment-1811166718:

> Provide a TracerProvider implementation from auto-instrumentation that manual instrumentation can explicitly use to say they want their data to go to the same pipeline. This means the auto-instrumentation will control the type that manual instrumentation uses to create, edit, and end spans if they want the data to be sent to the auto-instrumentation pipeline

Providing an SDK implementation like this would allow us to design minimal overhead in the Go implementation to resolve all the issues found with the current approach. Namely:

- For `IsRecording`, the span implementation could read its value from the span `struct` field that could be set in the eBPF space
- `RecordError` could call a method like `func (s span) recordError(errMsg, errName string, cfg trace.EventConfig). This method would have all the needed information for the probe already resolved.
- Similar to `RecordError`, `AddEvent` could call a `addEvent` method designed to have all the information for the probe already resolved in the call arguments.

### Interoperability

The remaining issues with providing a custom SDK, is how this would be integrated into users code?

The original idea was to offer this SDK as a package under `go.opentelemetry.io/auto`. This could still be provided, but if this is the only way this solution is integrated it means user code needs to be updated to inter-op with auto-instrumentation.

Requiring user code to be updated to work with auto-instrumentation is not desired. The whole point of auto-instrumentation is this kind of action would not be required.

Instead, we would like to have the default global `otel` implementation inter-op with this new SDK:

- A flag of some sort (i.e. a package level boolean `var`) would be added `go.opentelemetry.io/otel/internal/global`.
- This flag would be `false` by default
- When the Go auto-instrumentation loads, it will set this flag to `true`
[...]

PR #1045, opened 2024-08-27
PoC: Custom SDK

MrAlias:
PoC for #954 

This is a proof-of-concept for an SDK fully implemented by the auto-instrumentation. This supports all span functionality:

- Sampling (TODO: the `sample` method needs to be instrumented)
- Random correct ID generation
- All `Start` options
  - `WithLinks`
  - `WithNewRoot`
  - `WithSpanKind` (defaults to probe `SpanKind` if not set)
  - `WithTimestamp`
  - `WithAttributes`
- The `AddEvent` method, including all options
  - `WithStacktrace`
  - `WithAttributes`
  - `WithTimestamp`
- The `AddLink` method
- The `IsRecording` method (TODO: based on sampling support)
- The `SpanContext` method
- The `SetStatus` method
- The `SetAttribute` method
- The `TracerProvider` method
- All `End` options
  - `WithTimestamp`

#### `auto.GetTracerProvider`

There is only one function exported publicly. This is `GetTracerProvider` in `go.opentelemetry.io/auto`.

This function returns a singleton instance of an opentelemetry-go `trace.TracerProvider` that is held in the `internal/sdk` package.

#### `internal/sdk`

The `go.opentelemetry.io/auto/internal/sdk` package is added. This is a "full feature" OTel trace SDK from the perspective of the `Tracer` and `Span`.

All data about any `Span` created will be built in userspace. This is stored (mostly) in the collector's [`ptrace.Traces`](https://pkg.go.dev/go.opentelemetry.io/collector/pdata@v1.14.1/ptrace#Traces) type.

When the `Span` is ended the `ptrace.Traces` is marshaled into a proto binary encoding and passed as a buffer to the `ended` method of the `Span`. This method does nothing and is expecting a uprobe to be inserted at its call site.

#### `auto/sdk` probe

A simple probe is added to instrument the `go.opentelemetry.io/auto/internal/sdk` package. This probe does not rely on any offsets from the `sdk` types and simply routes the encoded span data from `ended` to the events eBPF map.

From there the `ptrace.Traces` data is unmarshaled and parsed into a `SpanEvent` that the `Controller` processes in the normal fashion.

#### Run Jaeger

```terminal
$ docker run --rm --name jaeger -e COLLECTOR_OTLP_ENABLED=true -p 16686:16686 -p 4318:4318 jaegertracing/all-in-one:latest
2024/08/28 20:38:34 maxprocs: Leaving GOMAXPROCS=8: CPU quota undefined
# ...
```

#### Run the example

```terminal
$ cd examples/auto-sdk && go build -o $GOPATH/bin/example && $GOPATH/bin/example
outter-0...done
outter-1...
```

#### Run the auto-instrumentation

```terminal
$ cd cli && go build
$ OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4318 \
  OTEL_GO_AUTO_TARGET_EXE=$GOPATH/bin/example \
  OTEL_SERVICE_NAME=example \
  sudo -E ./cli
{"level":"info","ts":1724885322.1967607,"logger":"go.opentelemetry.io/auto","caller":"cli/main.go:86","msg":"building OpenTelemetry Go instrumentation ...","globalImpl":false}
# ...
{"level":"info","ts":1724885324.8517134,"logger":"go.opentelemetry.io/auto","caller":"cli/main.go:115","msg":"instrumentation loaded successfully"}
```

You can also run with debug logging:

```terminal
$ OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4318 \
  OTEL_GO_AUTO_TARGET_EXE=$GOPATH/bin/example \
  OTEL_SERVICE_NAME=example \
  sudo -E ./cli -log-level=debug
```

Let this run for a bit and then stop the example. Stopping the example while there is a span active means you will get an error. E.g.

```terminal
go build -o $GOPATH/bin/example && $GOPATH/bin/example
outter-0...done
outter-1...^Cdone
```

(notice the `^C` is before the second `done`)

### Open Issues/Questions

- [ ] A maximum span serialization size of `412` is only supported
  - Ways to increase eBPF storage past the stack limit (512) need to be investigated
  - When we know the span is going to be too big, we need to drop attributes, links, and events in userspace
- [x] Sampling needs to be implemented.
- [ ] Fix call to `bpf_probe_read`: https://github.com/open-telemetry/opentelemetry-go-instrumentation/actions/runs/10605550069/job/29394558469?pr=1045
- [x] Currently the `SpanEvent` start and end times are relative offsets to the eBPF process time. This is changed in this PR, thereby breaking all other probes.
- [ ] Do we want to use `ptrace` from the collector as the serialization format? Do we want to build our own?
- [ ] This adds more uses of the `bpf_probe_write_user`. Can we use pinned eBPF maps to bypass this and communicate across processes?


internal/sdk/trace.go
```diff
@@ -0,0 +1,153 @@
...
+	newRoot     bool
+	spanKind    trace.SpanKind
+	stackTrace  bool
+}
+
+func (s Span) SpanContext() trace.SpanContext {
+	return s.spanContext
+}
+
+func (s Span) IsRecording() bool {
```
MrAlias, line 88: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#recording-sampled-reaction-table

This needs to return a value set during `embed` that the eBPF func will set a field. We need to return true here if sampled, and false otherwise. We do not plan to send the span to a processor if it is not sampled.

internal/sdk/trace.go
```diff
@@ -0,0 +1,153 @@
...
+
+func (Span) SetStatus(codes.Code, string) {}
+
+func (Span) SetAttributes(...attribute.KeyValue) {}
+
+func (s Span) End(opts ...trace.SpanEndOption) {
+	s.end(trace.NewSpanEndConfig(opts...))
+}
+
+func (Span) end(cfg trace.SpanConfig) {}
```
MrAlias, line 105:
```suggestion
func (s Span) End(opts ...trace.SpanEndOption) {
	buf := [bufLimit]byte{
		// TODO: serialize the span into a well defined byte format.
	}
	s.end(buf)
}

const bufLimit = 256  // TODO

func (Span) end(buf [bufLimit]byte) {}
```

internal/sdk/trace.go
```diff
@@ -0,0 +1,153 @@
...
+	spanKind    trace.SpanKind
+	stackTrace  bool
+}
+
+func (s Span) SpanContext() trace.SpanContext {
+	return s.spanContext
+}
+
+func (s Span) IsRecording() bool {
+	retVal := new(bool)
```
MrAlias, line 89: This should be a pointer to a field of the span. That means we will not need this offset.

internal/sdk/trace.go
MrAlias, line 105: We need to ensure this doesn't get inlined. Look for docs on comment needed to add here (i.e. `//go: noinline`).

RonFed:
> A maximum span serialization size of 412 is only supported

We can overcome this by using a `BPF_MAP_TYPE_PERCPU_ARRAY` which acts as a "heap". We are using this in multiple probes so you can have a look for example use cases.

Regarding my other comment, I think we should have this working example include an auto span in the same trace (HTTP).

internal/sdk/trace.go
```diff
@@ -0,0 +1,383 @@
...
+	ctx = trace.ContextWithSpan(ctx, span)
+	return ctx, span
+}
+
+// newSpan returns a new configured span.
+func (t *Tracer) newSpan(ctx context.Context, name string, cfg trace.SpanConfig) trace.Span {
+	// If told explicitly to make this a new root use a zero value SpanContext
+	// as a parent which contains an invalid trace ID and is not remote.
+	var psc trace.SpanContext
+	if cfg.NewRoot() {
```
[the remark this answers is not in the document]
MrAlias: Updated. I included the change and an example using both HTTP and DB instrumentation integrating with the manual span.

Issue #1026, opened 2024-08-19
Auto-instrumentation as a Collector receiver


damemi:
We could in theory expose a few parts of our internal code to make an auto-instrumentation receiver in the collector.

See https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/1025 for the code changes to our repo and the WIP collector changes in https://github.com/open-telemetry/opentelemetry-collector-contrib/compare/main...damemi:opentelemetry-collector-contrib:goautoreceiver

This would make it so you don't need to run an extra agent alongside a collector. It also makes it a lot easier to configure auto-instrumented telemetry with collector processors.

What do people think about this idea?

RonFed:
Is the use case for self tracing of the collector process?
If so, I think there will be some kind of positive feedback loop of exporting traces --> traces are automatically created for the exporter operation --> these traces are exported and this loop can go forever.
But maybe I did not understand the use case.

damemi:
The idea is to instrument user apps. Right now, the general approach with auto-instrumentation is to run your app+instrumentation as a sidecar. If you're running the collector as another sidecar, that adds complexity (and you have to coordinate startup/shutdown timing).

The Collector could just embed the auto-instrumentation libraries as a "receiver". Then the collector is an auto-instrumentation agent itself. I think this would only really work for Go, since that's what the collector is written in.

rogercoll:
+1 on this idea. Firstly, because it won't affect this repository (except the new exposed interfaces), but it will allow other applications to be built on top (as the collector receiver).

I think that having a receiver would ease the usage of this auto instrumentation as it will remove the side agent complexity in VM environments.  In Kubernetes, I think the Operator approach would still be the recommended as automatically injects the sidecar agent to the user's app Pod (required because it needs access to the user's app binary). Using the collector's receiver in K8s might be a bit more challenging and make the user apps deployment dependent to the collector.

Does the package require CGO? https://github.com/open-telemetry/opentelemetry-collector/blob/v0.109.0/CONTRIBUTING.md#using-cgo

PR #1050, opened 2024-09-04
Add the auto-instrumentation SDK wireframe

MrAlias:
Adds a new go.opentelemetry.io/auto/sdk module that holds the OpenTelemetry SDK implementation used by auto-instrumentation.

Part of #954
Split from #1045

See #1045 for the ultimate direction this wireframe will be built into.

This PR does not have a changelog entry because the module is ignored (it is not set to be released). A PR entry will be added when the wireframe is built out and this is ready to be released.


sdk/trace.go
```diff
@@ -0,0 +1,123 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package sdk
+
+import (
+	"context"
+
+	"go.opentelemetry.io/collector/pdata/ptrace"
```
damemi, line 9: why do we use the collector's objects here? Just curious

MrAlias: That's a good question. I'm not sure we do in the long term and it's something I had mentioned in the outstanding issues of https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/1045.

tl;dr: I have added it primary as a quick way to get this off the ground, and this is an already working binary encoding.

We need some encoding to transmit span data. As @RonFed  pointed out in a recent SIG meeting, trying to encode all the Go specific span data in C is challenging. Given we are going to have a processing chain of `Go -> eBPF (C) -> Go` if we have an encoding that is understood on both "Go" ends and can be transmitted though some generic `byte` structure in C we can avoid a lot of complexity.

The collector `ptrace` is an already highly-optimized binary encoding that provides a Go API. It is based off of OTLP and is designed to be highly performant. That means that we can start with this and get a working concept.

However, in the long run, there is still an open question as to whether we want to use this. The SDK is not included in the published module sets and it is something we should decide on before doing so.

Answering if we want to replace `ptrace` with our own binary encoding opens up some interesting questions:

- The collector is not stable, can we rely on `ptraces` if we want to go stable?
- Can we borrow tooling from the collector to make our own?
- The `Traces` type is over-sized in that it holds multiple spans. We only transmit one span at a time currently, do we need all the overhead of `Traces`? Should we update to utilize the full scope of `Traces` (see next).
- Should we fully pivot to using `pdata` as our data model? Instead of interfacing with an opentelemetry-go SDK should we interface with the collector receiver interface directly[^1]?
  - This means our public `Interface` type would be an instance of the collector instead of wrapping the opentelemetry-go APIs

[^1]: https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/1026

MrAlias: > Should we fully pivot to using `pdata` as our data model? Instead of interfacing with an opentelemetry-go SDK should we interface with the collector receiver interface directly[1](#user-content-fn-1-e6622de7b09db4912a65bf020341a161)?
> * This means our public `Instrumentation` type would be an instance of the collector instead of wrapping the opentelemetry-go APIs

This would mean, that instead of having something like this:

https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/1025/files#diff-f55965a613ff264fe2e580d1779471172a8c8a906c321256630aa0fdfed8eef3R13-R16

We would just use `ptrace.Traces` as the argument to `Controller.Trace`.

damemi (approved):

grcevski (approved):
LGTM! This looks like a reasonable change and if there are issues I think it will be a great first step we can build on top.

sdk/go.mod
```diff
@@ -0,0 +1,25 @@
+module go.opentelemetry.io/auto/sdk
+
+go 1.22.0
+
+toolchain go1.22.6
+
+require (
+	go.opentelemetry.io/collector/pdata v1.14.1
+	go.opentelemetry.io/otel v1.29.0
+	go.opentelemetry.io/otel/trace v1.29.0
```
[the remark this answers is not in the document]
MrAlias: It shouldn't, and in testing it looks to be the case that it doesn't.

Inter-module dependencies are allowed. They are resolved using the go mod strategy. And since this package can never import `go.openetelemetry.io/otel/internal/global` there is a strong guarantee that there will never be a cyclic import.

However, it verifying this, I did see indirect dependencies causing conflicts from `pdata`:

```terminal
go mod tidy in .
go mod tidy in ./bridge/opencensus
go: go.opentelemetry.io/otel/bridge/opencensus imports
	go.opentelemetry.io/otel imports
	go.opentelemetry.io/otel/internal/global imports
	go.opentelemetry.io/auto/sdk imports
	go.opentelemetry.io/collector/pdata/ptrace imports
	go.opentelemetry.io/collector/pdata/internal/data/protogen/collector/trace/v1 imports
	google.golang.org/grpc/status imports
	google.golang.org/genproto/googleapis/rpc/status: ambiguous import: found package google.golang.org/genproto/googleapis/rpc/status in multiple modules:
	google.golang.org/genproto v0.0.0-20200526211855-cb27e3aa2013 (/home/tyler/go/pkg/mod/google.golang.org/genproto@v0.0.0-20200526211855-cb27e3aa2013/googleapis/rpc/status)
	google.golang.org/genproto/googleapis/rpc v0.0.0-20240903143218-8af14fe29dc1 (/home/tyler/go/pkg/mod/google.golang.org/genproto/googleapis/rpc@v0.0.0-20240903143218-8af14fe29dc1/status)
go: go.opentelemetry.io/otel/bridge/opencensus imports
	go.opentelemetry.io/otel imports
	go.opentelemetry.io/otel/internal/global imports
	go.opentelemetry.io/auto/sdk imports
	go.opentelemetry.io/collector/pdata/ptrace imports
	go.opentelemetry.io/collector/pdata/internal/data/protogen/collector/trace/v1 imports
	google.golang.org/grpc/codes tested by
	google.golang.org/grpc/codes.test imports
	google.golang.org/genproto/googleapis/rpc/code: ambiguous import: found package google.golang.org/genproto/googleapis/rpc/code in multiple modules:
	google.golang.org/genproto v0.0.0-20200526211855-cb27e3aa2013 (/home/tyler/go/pkg/mod/google.golang.org/genproto@v0.0.0-20200526211855-cb27e3aa2013/googleapis/rpc/code)
	google.golang.org/genproto/googleapis/rpc v0.0.0-20240903143218-8af14fe29dc1 (/home/tyler/go/pkg/mod/google.golang.org/genproto/googleapis/rpc@v0.0.0-20240903143218-8af14fe29dc1/code)
go: go.opentelemetry.io/otel/bridge/opencensus imports
	go.opentelemetry.io/otel imports
	go.opentelemetry.io/otel/internal/global imports
	go.opentelemetry.io/auto/sdk imports
	go.opentelemetry.io/collector/pdata/ptrace imports
	go.opentelemetry.io/collector/pdata/internal/data/protogen/collector/trace/v1 imports
	google.golang.org/grpc/status tested by
	google.golang.org/grpc/status.test imports
	google.golang.org/genproto/googleapis/rpc/errdetails: ambiguous import: found package google.golang.org/genproto/googleapis/rpc/errdetails in multiple modules:
	google.golang.org/genproto v0.0.0-20200526211855-cb27e3aa2013 (/home/tyler/go/pkg/mod/google.golang.org/genproto@v0.0.0-20200526211855-cb27e3aa2013/googleapis/rpc/errdetails)
	google.golang.org/genproto/googleapis/rpc v0.0.0-20240903143218-8af14fe29dc1 (/home/tyler/go/pkg/mod/google.golang.org/genproto/googleapis/rpc@v0.0.0-20240903143218-8af14fe29dc1/errdetails)
```

This is likely only related to the way I was verifying:

- Update `go.opentelemetry.io/otel` to replace all references to `go.opentelemetry.io/auto/sdk` with the local copy (i.e. a go mod `replace` directive)
- Update `go.opentelemetry.io/otel/internal/global` with the dependency on this SDK
- Run `make go-mod-tidy` in opentelemetry-go

I'm still looking to see if I can resolve this conflict and have a clean build. But it does raise the question of if we want to use `pdata` in the long-term. Maybe we should look at using some custom data-structures encoded with `glob`(?).

sdk/trace.go
```diff
@@ -0,0 +1,123 @@
...
+	"go.opentelemetry.io/otel/trace/embedded"
+)
+
+// GetTracerProvider returns an auto-instrumentable [trace.TracerProvider].
+//
+// If an [go.opentelemetry.io/auto.Instrumentation] is configured to instrument
+// the process using the returned TracerProvider, all of the telemetry it
+// produces will be processed and handled by that Instrumentation. By default,
+// if no Instrumentation instruments the TracerProvider it will not generate
+// any trace telemetry.
```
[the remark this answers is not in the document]
MrAlias: I think adding package documentation about how the package performs makes sense. However, I'm not entirely sure the content of the documentation addition you are proposing. I don't plan to make the change in this PR.

Given this is supposed to be a wireframe from which we can build, I'll leave it to you to make a follow-up PR to add the desired documentation.

PR #1054, opened 2024-09-06
Flatten `auto/config` into `auto`

MrAlias:
Follow up to https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/982#discussion_r1735402520


RonFed (commented):
From a first glance it looks OK but I have a few questions.
The main advantage (and a very important one) is to not export internal functions.
However there are some downsides:
1. Adding one more level of conversion. 
2. Some duplication of the interfaces and structs related to config between the internal and the external ones. 
3. Added "complexity" to the code. 

Maybe this is the better approach, just wanted to hear the thoughts about the points I mentioned.
@MrAlias

damemi:
+1 to @RonFed's comments around the duplication+complexity. I would also prefer to organize the external-facing code into packages than having it all at the root.

I don't personally think the separation between internal/external is a high priority at this stage of the project, but I do think that addressing it is something we need to do. These changes make that line a little unclear to me too.

Wydt about taking the time to inventory our code and map out what makes sense to be public vs private in the ideal long term?

MrAlias:
I'm focused on ensuring that the user experience is prioritized here. Having an external API with unreachable types and functions or one that is bloated with unnecessary things is going to be a bad design with poor user experience. I'm also extremely hesitant to just add things to the API without justification and thought.

While we are at an early stage, with adoption, these APIs become de facto "stable".

I'm on-board for re-evaluating our approach. However, if we plan to do that we need to revert https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/982 and https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/1010 so we are doing so from a clean slate. Both of those PRs have added to the exported API and included elements mentioned above.

damemi:
@MrAlias I think we want the same things, but it's not clear to me what that looks like. So, I can't tell if this PR works toward that or not (I'm not saying it doesn't, I mean that I can't objectively say based on a specific goal)

I think we can reevaluate the approach without reverting everything. I would just like to know what the ideal end state of the user facing API is. So, just a step back that could include moving forward with this change too.

I am totally for treating the APIs as de facto stable, because even ignoring the user experience it builds good code practices to do so. That said, the user experience can still suffer if the public frontend is difficult to maintain. 

So I'm not disagreeing with you, and you've got a good example of a function that shouldn't be public. I'm saying let's look at every function and type, decide what they do, and design an API before we move forward trying to fix it. Do you think that's reasonable, or should we prioritize this patch right now?

MrAlias:
Yeah, I agree we are both striving for the same thing. I see this as an incremental positive change, and one we can build from. It addresses the two main issues mentioned above:

- Functionally related types and functions are unified into the same package.
  - The `Sampler` type used by the `Instrumentation` is in the same package
  - The `ConfigProvider` type used by the `Instrumentation` is in the same package
- It removes functions exposing internal types.

I'm happy to keep iterating here. I have thoughts about how these types could be refactored as well.

However, I do not think the current state of the project can just be left as is. It is not in a state that should be considered for release:

- https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/982 was merged with the expressed qualification that this type of refactor would be required afterwards: https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/982#discussion_r1742400937, https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/982#pullrequestreview-2283861134
- https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/1010 was merged with a single approval from the same company as the PR author

Both of these changes introduced modifications to our public API without the communities buy-in to the change. If we want to not move forward iterating on these issues we need to roll them back and address their inclusion from a clean slate.

damemi:
I think we agree that we need a longer term vision, but I also think we can unstick the current state to address the promises made in those PRs that we'd fix this.

So my two main concerns with this change are that 1) it seems somewhat over-flattened and 2) I don't understand why the duplication in `internal/pkg/instrumentation/config.go` and `provider.go`

For 1) I don't mind this being flattened, but if you're intending for that to be part of the long term design then I'd like to block and discuss it. Otherwise, that's totally fine with me to get things shaped up for now.

For 2), I could just be missing something so it would help if you can please explain that duplication

And just to get on the same page, do you mean for this PR to be part of the long term design or are you open to changes here as we work that out?

MrAlias:
> So my two main concerns with this change are that 1) it seems somewhat over-flattened [...]
> 
> For 1) I don't mind this being flattened, but if you're intending for that to be part of the long term design then I'd like to block and discuss it. Otherwise, that's totally fine with me to get things shaped up for now.

Can you expand on this? I don't see a case where a user would ever import `go.opentelemetry.io/auto/config` without importing `go.opentelemetry.io/auto` first. The `go.opentelemetry.io/auto/config` package is not a logically distinct collection of code. It is a part of `go.opentelemetry.io/auto` that seems to be broken off to support underlying developer complexity reduction.

FWIW, I am not opposed to adding additional package to `go.opentelemetry.io/auto/`. If there are logical distinct units and reusable code that will be used in multiple places it would make sense to add that code as its own package. This `config` package is demonstrably not that though.

MrAlias:
> 2) I don't understand why the duplication in internal/pkg/instrumentation/config.go and provider.go
> 
> [...]
>
> For 2), I could just be missing something so it would help if you can please explain that duplication

There is duplication because of the alternate representation of a `Sampler` externally vs internally. The internal provide uses the internal definition of a `sampling.Config`, where the external declaration uses the external `auto.Sampler`.

This is an area I think we could iterate on. Especially since we are going to need an way to allow users to pass custom `Sampler`s and support the remote samplers (i.e. Jaeger, AWS).

This forcing function will likely result in the `Sampler` definition to be updated into something used in both places and will likely result in a new `auto/sampling` package. But this is very speculative and I haven't done more than thought about it.

MrAlias:
I would like to merge this PR or revert the prior two. I see the current state of the repository in limbo without doing either.

I do not feel comfortable releasing what we have on `main` currently, which means we are blocked releasing all the existing changes we have merged.

I'm also not 100% sure I would want to release with this merged either. However, I see incremental changes that can be made on top of this that could get us to a release-able place.

damemi (approved):
Talked to @MrAlias offline to not clog up the thread with my own questions, because I was just pretty confused by this. But I think I understand now and it lgtm

As I see it right now, the goal is to fix up the code to a release-able state, but there is still some much broader design work to do. Ideally, that lets us reduce some of the off bits around these areas (like the duplication). But we can't get anywhere with the code in limbo

RonFed (approved):

Issue #1049, opened 2024-09-02
Got error when tracing kafka


NhatVuDuy:
### Describe the bug

I got a crash when using otel auto instrumentation docker container with the main container has kafka service.

### Environment

 - OS: [ubuntun k8s for go]
 - Go Version: [1.22]

### To Reproduce

Run otel auto instrumentation docker container with main docker container.
Got the error if main container has kafka service.

### Additional context

The log is:

```json
{"level":"info","ts":1724839308.755223,"logger":"[go.opentelemetry.io/auto](https://www.google.com/url?q=http://go.opentelemetry.io/auto&sa=D&source=editors&ust=1725277443124090&usg=AOvVaw2x1v3n3-SA5Mws4svnjIBu)","caller":"cli/main.go:86","msg":"building OpenTelemetry Go instrumentation ...","globalImpl":false}
{"level":"info","ts":1724839310.7561524,"logger":"Instrumentation.Analyzer","caller":"process/discover.go:67","msg":"found process","pid":8}
{"level":"error","ts":1724839310.7569983,"logger":"Instrumentation.Analyzer","caller":"process/analyze.go:83","msg":"error parsing module version","error":"Malformed version: (devel)","stacktrace":"[go.opentelemetry.io/auto/internal/pkg/process.(*Analyzer).Analyze\n\t/app/internal/pkg/process/analyze.go:83\ngo.opentelemetry.io/auto.NewInstrumentation\n\t/app/instrumentation.go:133\nmain.main\n\t/app/cli/main.go:104\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:271](https://www.google.com/url?q=http://go.opentelemetry.io/auto/internal/pkg/process.(*Analyzer).Analyze%5Cn%5Ct/app/internal/pkg/process/analyze.go:83%5Cngo.opentelemetry.io/auto.NewInstrumentation%5Cn%5Ct/app/instrumentation.go:133%5Cnmain.main%5Cn%5Ct/app/cli/main.go:104%5Cnruntime.main%5Cn%5Ct/usr/local/go/src/runtime/proc.go:271&sa=D&source=editors&ust=1725277443124181&usg=AOvVaw2Y3tc1TVDifYJWsrumCeiF)"}
{"level":"info","ts":1724839310.7942715,"logger":"Instrumentation.Allocate","caller":"process/allocate.go:73","msg":"Detaching from process","pid":8}
[...]

PR #1077, opened 2024-09-09
Log the CLI version

MrAlias:
To help debug issues[^1], include information about the CLI version when starting. Include the release version, a git hash, and information about the Go setup.

[^1]: https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/1049


damemi (approved):

cli/version.go
```diff
@@ -0,0 +1,66 @@
...
+
+type goInfo struct {
+	Version string
+	OS      string
+	Arch    string
+}
+
+func newGoInfo() goInfo {
+	return goInfo{
+		Version: runtime.Version(),
```
RonFed, line 62: This will print the Go version of the agent. IIRC we are printing the go version of the instrumented process but I'm not sure

RonFed (approved):

cli/version.go
MrAlias: Right, this will print the version of Go used to build the auto-instrumentation. That is the intention given we may need that to help debug issues that arise with the auto-instrumentation.

MrAlias: We also print the Go version of the target processes, but that is a later log message deeper in the code.

PR #1080, opened 2024-09-11
Use slog instead of logr

### LINKED ISSUE #1104: Logging Is Closed-Off and Unconfigurable
When I am building on top of the `Instrumentation` type, I am met with many short-falls with its logging.

There are many situations I would like to handled. It would be ideal if the logging allowed me to...

- Use the OpenTelemetry Logs Bridge to process and handle my logs.
- Use a different log format based on the environment (i.e. development, staging, production) the application is running in
- Use a different log processing pipeline based on the environment the application is running in
- Disable logging
- Use different environment variables than the OpenTelemetry defined schema to configure my logging
- Not add additional `zap` logging dependencies to my project when I don't use them

Ideally, the logging for this project would allow me to integrate with existing logging pipelines and better customization.

MrAlias:
- Use the built in `slog` package to log instead of using `logr`.
- Replace `WithLogLevel` with `WithLogger`
  - Allow users to provide a fully-customized logger with their handler of choice instead of imposing a choice on them.
- Removes dependency on `github.com/go-logr/zapr`, `go.uber.org/zap v1.27.0`, and `go.uber.org/multierr`

Resolves #1104 

### Example Output

Raw:

```terminal
{"time":"2024-09-11T17:29:35.700958753Z","level":"INFO","source":{"function":"go.opentelemetry.io/auto/internal/pkg/process.(*Analyzer).DiscoverProcessID","file":"/app/internal/pkg/process/discover.go","line":64},"msg":"found process","pid":754625}
{"time":"2024-09-11T17:29:35.723405405Z","level":"INFO","source":{"function":"go.opentelemetry.io/auto/internal/pkg/process.remoteAllocate.func1","file":"/app/internal/pkg/process/allocate.go","line":70},"msg":"Detaching from process","pid":754625}
{"time":"2024-09-11T17:29:35.723583269Z","level":"INFO","source":{"function":"go.opentelemetry.io/auto.NewInstrumentation","file":"/app/instrumentation.go","line":111},"msg":"target process analysis completed","pid":754625,"go_version":"1.23.1","dependencies":{"go.uber.org/multierr":"1.10.0","go.uber.org/zap":"1.27.0","std":"1.23.1"},"total_functions_found":2}
{"time":"2024-09-11T17:29:35.723973977Z","level":"INFO","source":{"function":"main.main","file":"/app/cli/main.go","line":134},"msg":"starting instrumentation..."}
{"time":"2024-09-11T17:29:35.731129749Z","level":"INFO","source":{"function":"go.opentelemetry.io/auto/internal/pkg/instrumentation.(*Manager).load","file":"/app/internal/pkg/instrumentation/manager.go","line":323},"msg":"loading probe","name":{"SpanKind":2,"InstrumentedPkg":"net/http"}}
{"time":"2024-09-11T17:29:36.079187915Z","level":"INFO","source":{"function":"go.opentelemetry.io/auto/internal/pkg/instrumentation.(*Manager).ConfigLoop","file":"/app/internal/pkg/instrumentation/manager.go","line":237},"msg":"Configuration provider closed, configuration updates will no longer be received"}
{"time":"2024-09-11T17:29:36.079275764Z","level":"INFO","source":{"function":"main.main.func3","file":"/app/cli/main.go","line":130},"msg":"instrumentation loaded successfully"}
```

Wrapped:

```json
{
  "time": "2024-09-11T17:29:35.723973977Z",
  "level": "INFO",
  "source": {
    "function": "main.main",
    "file": "/app/cli/main.go",
    "line": 134
  },
  "msg": "starting instrumentation..."
},
{
  "time": "2024-09-11T17:29:35.731129749Z",
  "level": "INFO",
  "source": {
    "function": "go.opentelemetry.io/auto/internal/pkg/instrumentation.(*Manager).load",
    "file": "/app/internal/pkg/instrumentation/manager.go",
    "line": 323
  },
  "msg": "loading probe",
  "name": {
    "SpanKind": 2,
    "InstrumentedPkg": "net/http"
  }
},
{
  "time": "2024-09-11T17:29:36.079187915Z",
  "level": "INFO",
  "source": {
    "function": "go.opentelemetry.io/auto/internal/pkg/instrumentation.(*Manager).ConfigLoop",
    "file": "/app/internal/pkg/instrumentation/manager.go",
    "line": 237
  },
  "msg": "Configuration provider closed, configuration updates will no longer be received"
},
{
  "time": "2024-09-11T17:29:36.079275764Z",
  "level": "INFO",
  "source": {
    "function": "main.main.func3",
    "file": "/app/cli/main.go",
    "line": 130
  },
  "msg": "instrumentation loaded successfully"
}
```


instrumentation_test.go
```diff
@@ -178,28 +163,6 @@ func TestWithResourceAttributes(t *testing.T) {
 	})
 }
 
-func TestWithLogLevel(t *testing.T) {
```
RonFed, line 181: Do we want to replace these with tests for the new `WithLogger` and also one for the env var `OTEL_LOG_LEVEL`?

instrumentation.go
```diff
@@ -441,16 +401,6 @@ func WithEnv() InstrumentationOption {
 				c.globalImpl = boolVal
 			}
 		}
-		if l, ok := lookupEnv(envLogLevelKey); ok {
```
RonFed, line 444: If I understand this correctly, this change implies that `OTEL_LOG_LEVEL` is only being inspected when the instrumentation is launched from the CLI. When the instrumentation is being called as a library it won't be inspected? If that is the case I think we should change this behavior.

MrAlias: That is correct. The `Instrumentation` no longer sets the minimum severity of the logger because it no longer constructs the logger. The `slog.Logger` is not something that is configured with this level it is something that the caller will need to setup.

Meaning the caller is given the opportunity to decide what ever way they want to configure this setting, or not.

instrumentation_test.go
MrAlias: I'll add a test for `WithLogger`. I'm not sure exactly how to test the CLI for `OTEL_LOG_LEVEL` though.

instrumentation.go
RonFed: If the `WithLogger` is not used, I'd expect `OTEL_LOG_LEVEL` to still have an impact

MrAlias: I've captured the many short-falls of our existing logging and examples of things I would like to achieve in https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/1104. This PR serves to address all of those.

I do not think we can force the `OTEL_LOG_LEVEL` on everyone and assume it will meet their needs. Accepting a `Logger` will still allow for this configuration, but will also allow for much more.

MrAlias: If the user doesn't pass the logger, we should still use `OTEL_LOG_LEVEL` to setup our default.

MrAlias: Updated. PTAL.

instrumentation_test.go
MrAlias: Updated. PTAL.

RonFed (approved):
looks great

PR #1107, opened 2024-09-18
Add the SDK probe

MrAlias:
**NOTE**: much of this change-set is boilerplate; there are **not** > 2.5k lines of code to review in detail

Instrument the SDK with an auto-instrumentation probe.

I have not added integration testing here as I want to keep this PR size "reasonable". It is planned as a follow-up task (see below).

Part of #954 

### Follow-up task

- [ ] Add integration (e.g. `e2e`) testing: https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/1108
- [ ] Support and test setting attribute setting on start
- [ ] Support and test setting span kind on start
- [ ] Support and test adding links on start
- [ ] Implement and test the rest of the `Span` methods
- [ ] Add example of usage


internal/pkg/instrumentation/bpf/go.opentelemetry.io/auto/sdk/bpf/probe.bpf.c
```diff
@@ -0,0 +1,219 @@
...
+		if (rc < 0) {
+			bpf_printk("failed to read encoded span data");
+			return -3;
+		}
+	} else {
+		bpf_printk("read too large: %d", event->size);
+		return -4;
+	}
+
+	bpf_perf_event_output(ctx, &events, BPF_F_CURRENT_CPU, event, sizeof(*event));
```
RonFed, line 216: We better use the common `output_span_event` function which all the other probes use. That function checks if the span is sampled or not and outputs it accordingly.
When including `span_output.h` the definition of the perf buffer is also included so the one in this file will need to be removed.
edit: if we avoid putting un-sampled spans in the `active_spans_by_span_ptr` map, we should not reach here for un-sampled spans, but it is still preferred to use that function for consistency.

RonFed, line 216: We can also save space on the perf buffer in this case, since we know the actual encoding size we don't have to output all the event buffer.

internal/pkg/instrumentation/bpf/go.opentelemetry.io/auto/sdk/bpf/probe.bpf.c
```diff
@@ -0,0 +1,219 @@
...
+	if (ret != 0) {
+		rc--;
+		bpf_printk("span_context_from_Go: failed to read trace flags", ret);
+	}
+
+	return rc;
+}
+
+SEC("uprobe/Tracer_start")
+int uprobe_Tracer_start(struct pt_regs *ctx) {
```
RonFed, line 117: please add a comment including the signature of the instrumented function

internal/pkg/instrumentation/bpf/go.opentelemetry.io/auto/sdk/bpf/probe.bpf.c
```diff
@@ -0,0 +1,219 @@
...
+		bpf_printk("write_span_context: NULL go_sc");
+		return -1;
+	}
+
+	long rc = 0;
+	void *tid = (void *)(go_sc + span_context_trace_id_pos);
+	long ret = bpf_probe_write_user(tid, &sc->TraceID, TRACE_ID_SIZE);
+	if (ret != 0) {
+		rc--;
+		bpf_printk("write_span_context: failed to write trace ID: %ld", ret);
```
RonFed, line 63: If we fail to write, do we want to return or keep trying for the other fields (as in the current code)?

internal/pkg/instrumentation/bpf/go.opentelemetry.io/auto/sdk/bpf/probe.bpf.c
```diff
@@ -0,0 +1,219 @@
...
+	}
+
+	void *span_context_ptr_val = get_argument(ctx, 7);
+	rc = write_span_context(span_context_ptr_val, &otel_span.sc);
+	if (rc != 0) {
+		bpf_printk("failed to write span context: %ld", rc);
+	}
+
+	void *span_ptr_val = get_argument(ctx, 4);
+    bpf_map_update_elem(&active_spans_by_span_ptr, &span_ptr_val, &otel_span, 0);
```
RonFed, line 165: I think we should not put un-sampled spans on this map - this will allow us to handle more traffic if sampling is used.

internal/tools/inspect/cmd/offsetgen/main.go
```diff
@@ -163,6 +163,24 @@ func manifests() ([]inspect.Manifest, error) {
...
+					"go.opentelemetry.io/otel/trace",
+					"SpanContext",
+					"spanID",
+				),
+				structfield.NewID(
+					"go.opentelemetry.io/otel",
+					"go.opentelemetry.io/otel/trace",
+					"SpanContext",
+					"traceFlags",
+				),
```
RonFed, line 183: nit: this different format is on purpose? do we want to keep it consistent?

internal/pkg/instrumentation/bpf/go.opentelemetry.io/auto/sdk/bpf/probe.bpf.c
MrAlias: Won't that mean that context will not be propagated for un-sampled spans? That would mean things like the parent based sampler wouldn't work correctly, right?

internal/pkg/instrumentation/bpf/go.opentelemetry.io/auto/sdk/bpf/probe.bpf.c
MrAlias: That function uses a span type definitions as the transport payload. This uses a payload of the encoded binary span data from Go. I don't think `output_span_event` is a compatible replacement, right?

MrAlias: How would we do that? My understanding is that the event buffer (i.e. `event.data`) is allocated to the stack? How do we dynamically send a smaller buffer?

Would the Go `event` definition also need to be updated? Would it still be an array?

RonFed: We need to pass the span context which we have here in `otel_span_t`, and that function accepts a pointer and a size which I think we can provide here as well. Does that make sense or am I missing something?

internal/tools/inspect/cmd/offsetgen/main.go
MrAlias: Updated.

internal/pkg/instrumentation/bpf/go.opentelemetry.io/auto/sdk/bpf/probe.bpf.c
MrAlias: Oh, my mistake. I was looking at the wrong function definition. Yeah, I think that makes sense, I'll update.

internal/pkg/instrumentation/bpf/go.opentelemetry.io/auto/sdk/bpf/probe.bpf.c
RonFed: If we call `start_tracking_span` it should make sure the context is propagated. That means we need to call `stop_tracking_span` when the span in ended. However, we use the `active_spans_by_span_ptr` to fetch the span in the end probe so not updating the map means we won't be able to fetch the span in the return probe, and won't clean the span context maps properly.

I'd prefer to save maps space for un-sampled spans but it seems more tricky than I initially thought.

internal/pkg/instrumentation/bpf/go.opentelemetry.io/auto/sdk/bpf/probe.bpf.c
MrAlias: How do we determine the size though? Is it as simple as `sizeof(event.size)` (i.e. 4) + `event.size`?

RonFed: I thinks so

MrAlias: Updated. PTAL.

MrAlias: Flattening `output_span_event` and checking size before `bpf_perf_event_output` got around the last error. However, now the base64 decoding is failing:

```terminal
{
  "time": "2024-09-20T10:22:43.749409627-07:00",
  "level": "ERROR",
  "source": {
    "function": "go.opentelemetry.io/auto/internal/pkg/instrumentation/probe.(*Base[...]).Run",
    "file": "/home/tyler/go/src/go.opentelemetry.io/auto/internal/pkg/instrumentation/probe/probe.go",
    "line": 216
  },
  "msg": "failed to process perf record",
  "error": "unexpected EOF"
}
```

I'm guessing this because it is trying to deserialize into an object larger than the serialized data now that we have shortened.  I'm going to verify with some ad hoc logging, but I expect this change might not be viable.

RonFed: There might be 4 bytes of padding between the size and data fields which causes the parsing error.

MrAlias: Hmm, not sure. Reading an extra 4 bytes after the `Size` didn't change the error. This involved writing a custom `perf.Record` parser for the SDK probe.

I was able to add more logging information and those error messages are coming from the `net/http` and `database/sql` probes. It looks like this is corrupting those perf events.

RonFed: Yes, doing this will require having a different parsing of perf events. The generic one assumes constant size events.

MrAlias: Gah, looks like I had a staged change that I didn't realize. I think I may have this working, disregard the last comment. :facepalm:

MrAlias: Updated. PTAL

Note, to support dynamically outputting the event, we cannot use `output_span_event`. The verifier complains that the size passed is not verified to not exceed the allowed memory. I have flattened it into the function.

MrAlias: This was reverted to accommodate https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/1107#discussion_r1768687578. See https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/1107#discussion_r1769102558.

internal/pkg/instrumentation/bpf/go.opentelemetry.io/auto/sdk/bpf/probe.bpf.c
MrAlias: @RonFed do you have thoughts on how to solve this? I'm not sure how to track span context without also tracking the span.

Should we just open an issue to look into this later?

internal/pkg/instrumentation/bpf/go.opentelemetry.io/auto/sdk/bpf/probe.bpf.c
```diff
@@ -0,0 +1,196 @@
...
+		return -4;
+	}
+
+	// Do not send the whole size.buf if it is not needed.
+	u64 size = sizeof(event->size) + event->size;
+	// Make the verifier happy, ensure no unbounded memory access.
+	if (size < sizeof(struct event_t)+1) {
+		return bpf_perf_event_output(ctx, &events, BPF_F_CURRENT_CPU, event, size);
+	}
+	bpf_printk("write too large: %d", event->size);
```
RonFed: Maybe this will make the verifier happy with the `output_span_event` function

```suggestion
         if (event->size > MAX_SIZE) {
             bpf_printk("write too large: %d", event->size);
             return -5;
         }
         // assuming MAX_SIZE is a power of 2
         u64 payload_size = size & (MAX_SIZE - 1)
         output_span_event(ctx, event, sizeof(event->size) + payload_size,  &span->sc)
```

internal/pkg/instrumentation/bpf/go.opentelemetry.io/auto/sdk/bpf/probe.bpf.c
RonFed: Yeah, worth opening an issue, it is relevant to all the probes.

internal/pkg/instrumentation/bpf/go.opentelemetry.io/auto/sdk/bpf/probe.bpf.c
MrAlias: I think a version of this could work. But, I'm wondering if we need to use `output_span_event`. We are checking the sampling decision at the top of the function and avoiding unnecessary work there. Given that `output_span_event` just wraps that check and `bpf_perf_event_output` it seems like it will just introduce redundant steps.

RonFed: Right, I missed the check for the is_sampled

internal/pkg/instrumentation/bpf/go.opentelemetry.io/auto/sdk/bpf/probe.bpf.c
```diff
@@ -0,0 +1,196 @@
...
+// https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/effdec9ac23e56e9e9655663d386600e62b10871/sdk/trace.go#L133-L136
+SEC("uprobe/Span_ended")
+int uprobe_Span_ended(struct pt_regs *ctx) {
+    void *span_ptr = get_argument(ctx, 1);
+    struct otel_span_t *span = bpf_map_lookup_elem(&active_spans_by_span_ptr, &span_ptr);
+    if (span == NULL) {
+        return 0;
+    }
+    stop_tracking_span(&span->sc, &span->psc);
+    bpf_map_delete_elem(&active_spans_by_span_ptr, &span_ptr);
```
RonFed, line 148: We should call delete only once we finish any handling with the underlying memory of the map value.

MrAlias: I think the only thing we do with the read `span` is check sampling. I'll update to store that in a local var prior to deleting.

MrAlias: Updated. PTAL.

internal/pkg/instrumentation/bpf/go.opentelemetry.io/auto/sdk/bpf/probe.bpf.c
MrAlias: Tracking here: https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/1117

MrAlias: I'm going to resolve this comment as this looks like a systemic issue we want to address and it doesn't seem like something that should block this initial probe for the SDK.

LMK if that is in error.

damemi (approved):

RonFed (approved):

PR #1108, opened 2024-09-19
Add end-to-end tests for the SDK

MrAlias:


internal/test/e2e/autosdk/build.sh
RonFed, line 1: What's the purpose of this script? Why we couldn't build the Dockerfile like we did in the other tests?

RonFed (approved):

internal/test/e2e/autosdk/build.sh
MrAlias: The build context needs to be at the root of the repository so it can copy the `sdk` package for testing.

The current builds all have build context local to the `Dockerfile`.

PR #1115, opened 2024-09-19
Bump Go supported versions

MrAlias:
Use Go 1.23 for building and testing.


damemi:
FYI someone just shared this issue with me, apparently there are problems with go 1.23 on ARM https://github.com/golang/go/issues/68976. Not sure if it's relevant but I did see the arm test failed here

MrAlias:
> FYI someone just shared this issue with me, apparently there are problems with go 1.23 on ARM [golang/go#68976](https://github.com/golang/go/issues/68976). Not sure if it's relevant but I did see the arm test failed here

The test failure seems to have been unrelated. Looks like the Go proxy issues from yesterday was the related problem.

RonFed (approved):

PR #1126, opened 2024-09-24
Add SDK data-model package

MrAlias:
This is the planned replacement for `pdata` within the `sdk` pacakge.


sdk/telemetry/traces.go
```diff
@@ -0,0 +1,189 @@
...
+
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// A collection of ScopeSpans from a Resource.
+type ResourceSpans struct {
```
MrAlias, line 75:
```suggestion
type ResourceSpan struct {
```

PR #1127, opened 2024-09-25
Instrument gRPC server span status codes

damemi:
Ref #172


internal/pkg/instrumentation/bpf/google.golang.org/grpc/server/bpf/probe.bpf.c
```diff
@@ -167,3 +170,48 @@ int uprobe_http2Server_operateHeader(struct pt_regs *ctx)
...
+    u32 stream_id = 0;
+    bpf_probe_read(&stream_id, sizeof(stream_id), (void *)(stream_ptr + stream_id_pos));
+    struct grpc_request_t *grpcReq = bpf_map_lookup_elem(&streamid_to_grpc_events, &stream_id);
+    if (grpcReq == NULL) {
+        // No parent span context, generate new span context
+        u32 zero = 0;
+        grpcReq = bpf_map_lookup_elem(&grpc_storage_map, &zero);
+        if (grpcReq == NULL) {
+            bpf_printk("failed to get grpcReq from storage map");
+            return 0;
```
MrAlias, line 190:
```suggestion
            return -1;
```

MrAlias (approved):

internal/pkg/instrumentation/bpf/google.golang.org/grpc/server/probe.go
```diff
@@ -72,6 +83,14 @@ func New(logger *slog.Logger) probe.Probe {
...
 				EntryProbe: "uprobe_http2Server_operateHeader",
 			},
+			{
+				Sym:        "google.golang.org/grpc/internal/transport.(*serverHandlerTransport).WriteStatus",
+				EntryProbe: "uprobe_serverHandlerTransport_WriteStatus",
+			},
+			{
+				Sym:        "google.golang.org/grpc/internal/transport.(*http2Server).WriteStatus",
+				EntryProbe: "uprobe_http2Server_WriteStatus",
+			},
```
RonFed, line 93: Are we instrumenting both of them because either of them might be used?

internal/pkg/instrumentation/bpf/google.golang.org/grpc/server/bpf/probe.bpf.c
```diff
@@ -167,3 +170,48 @@ int uprobe_http2Server_operateHeader(struct pt_regs *ctx)
...
+static __always_inline int get_status_code(struct pt_regs *ctx) {
+    struct go_iface go_context = {0};
+    get_Go_context(ctx, 2, stream_ctx_pos, true, &go_context);
+    void *key = get_consistent_key(ctx, go_context.data);
+
+    // Get parent context if exists
+    void *stream_ptr = get_argument(ctx, 2);
+    u32 stream_id = 0;
+    bpf_probe_read(&stream_id, sizeof(stream_id), (void *)(stream_ptr + stream_id_pos));
+    struct grpc_request_t *grpcReq = bpf_map_lookup_elem(&streamid_to_grpc_events, &stream_id);
```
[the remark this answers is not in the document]
damemi: That's a good question. If `handleStream` isn't called (say the server hits an error before that) then we might still want to record the status from `WriteStatus`. Maybe in that case, this probe should check if an event already exists in `grpc_events` and if not start a new span with the status?

It seems like `WriteStatus` might be used in places besides `handleStream` ([SendMsg](https://github.com/grpc/grpc-go/blob/ca4865d6dd6f3d8b77f1943ccfd6c9e78223912d/stream.go#L1655) and [RecvMsg](https://github.com/grpc/grpc-go/blob/ca4865d6dd6f3d8b77f1943ccfd6c9e78223912d/stream.go#L1743) are 2 instances I found). Do you think these are relevant? If not then I agree with your suggestion

internal/pkg/instrumentation/bpf/google.golang.org/grpc/server/probe.go
RonFed: It needs to be a struct.
The point I think worth verifying is whether: a) one of these implementations will be called per stream b) both of them will be called and each is a different layer.
The first case sounds more reasonable but worth making sure with a debugger. WDYT?

internal/pkg/instrumentation/bpf/google.golang.org/grpc/server/bpf/probe.bpf.c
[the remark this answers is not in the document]
RonFed: I think that depends on the [RPC life cycle](https://grpc.io/docs/what-is-grpc/core-concepts/#rpc-life-cycle) the application is using.
The creation of a new stream can be called on the initialization phase of the instrumented process, that is also a case where we'll miss the entry probe of `handleStream`.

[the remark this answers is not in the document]
damemi: > The creation of a new stream can be called on the initialization phase of the instrumented process, that is also a case where we'll miss the entry probe of handleStream

That sounds like a case where we would want to create a new trace in the WriteStatus probe, is that what you mean?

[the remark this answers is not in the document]
RonFed: I'm not sure. We won't have the RPC method in that case

internal/pkg/instrumentation/bpf/google.golang.org/grpc/server/probe.go
damemi: @RonFed I wasn't sure how I would trigger one or the other, so I don't think I could get that reliably from a debugger. But I did open https://github.com/grpc/grpc-go/issues/7730 to ask the grpc maintainers and got this reply that seems to answer it: https://github.com/grpc/grpc-go/issues/7730#issuecomment-2406553932

Sounds like http2server is the default gRPC implementation, and serverHandlerTransport is used for integrating gRPC into existing HTTP handlers. Given our goals (and the fact that we only probe http2server for `operateHeaders`), I think we can just go with that.

In the future, we might want to support serverHandlerTransport for the integration use case. But for OOTB gRPC support, http2 should be good if I understand. Will update the PR to just use that one

RonFed: Sounds good, thanks for checking

internal/pkg/instrumentation/bpf/google.golang.org/grpc/server/bpf/probe.bpf.c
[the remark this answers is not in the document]
damemi: WriteStatus does have the same `Stream` argument that is passed to `handleStream`, so we could try to get the RPC method from there. I just pushed an update that does that, ptal. I think this would be neat to cover the cases where we miss the original initialization like you mentioned. If not I'm all for just simplifying this down

internal/pkg/instrumentation/bpf/google.golang.org/grpc/server/bpf/probe.bpf.c
```diff
@@ -167,3 +170,65 @@ int uprobe_http2Server_operateHeader(struct pt_regs *ctx)
...
+    bpf_probe_read_user(&s_ptr, sizeof(s_ptr), (void *)(status_ptr + status_s_pos));
+    // Get status code from Status.s pointer
+    bpf_probe_read_user(&grpcReq->status_code, sizeof(grpcReq->status_code), (void *)(s_ptr + status_code_pos));
+
+    bpf_map_update_elem(&grpc_events, &key, grpcReq, 0);
+    bpf_map_delete_elem(&streamid_to_grpc_events, &stream_id);
+    return 0;
+}
+
+UPROBE_RETURN(http2Server_WriteStatus, struct grpc_request_t, grpc_events, events, 2, stream_ctx_pos, false)
```
RonFed, line 234: I am not sure about this return probe, since we already have `UPROBE_RETURN(server_handleStream...)`.
The `UPROBE_RETURN` macro sends the event to user-space so I don't think we need to have it twice.

damemi: my thinking here was to use the return probe in the case where the initialization (handleStream) was missed. Since the handleStream entry probe deletes the event from the map on success, I think [this return probe will be a no-op](https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/717ea4f1b7a8b60b0a4f0a99e1438adb9f1ad448/internal/include/uprobe.h#L33-L36) if we're not making a span for WriteStatus

internal/pkg/instrumentation/bpf/google.golang.org/grpc/server/bpf/probe.bpf.c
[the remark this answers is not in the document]
damemi: Update: based on what we talked about in the sig meeting I simplified this down to what you suggested @RonFed. This will now escape early if the handleStream event isn't found. Thanks for the back and forth! I think we learned a lot about this in case we decide to implement more handling here in the future.

damemi:
Updated based on feedback and sig discussion, ptal!

internal/pkg/instrumentation/bpf/google.golang.org/grpc/server/bpf/probe.bpf.c
```diff
@@ -167,3 +170,28 @@ int uprobe_http2Server_operateHeader(struct pt_regs *ctx)
...
+// func (ht *http2Server) WriteStatus(s *Stream, st *status.Status)
+// https://github.com/grpc/grpc-go/blob/bcf9171a20e44ed81a6eb152e3ca9e35b2c02c5d/internal/transport/http2_server.go#L1049
+SEC("uprobe/http2Server_WriteStatus")
+int uprobe_http2Server_WriteStatus(struct pt_regs *ctx) {
+    struct go_iface go_context = {0};
+    get_Go_context(ctx, 2, stream_ctx_pos, true, &go_context);
+    void *key = get_consistent_key(ctx, go_context.data);
+
+    struct grpc_request_t *grpcReq_event_ptr = bpf_map_lookup_elem(&grpc_events, &key);
+    // if grpcReq_event is null, then handleStream probe didn't run. Try starting a new span here
```
RonFed, line 183:
```suggestion
    // if we fail to get an active grpc span, nothing to do here
```

internal/pkg/instrumentation/bpf/google.golang.org/grpc/server/bpf/probe.bpf.c
```diff
@@ -167,3 +170,28 @@ int uprobe_http2Server_operateHeader(struct pt_regs *ctx)
...
+        bpf_printk("failed to get grpcReq_event from events map");
+        return -1;
+    }
+
+    void *status_ptr = get_argument(ctx, 3);
+    void *s_ptr = 0;
+    bpf_probe_read_user(&s_ptr, sizeof(s_ptr), (void *)(status_ptr + status_s_pos));
+    // Get status code from Status.s pointer
+    bpf_probe_read_user(&grpcReq_event_ptr->status_code, sizeof(grpcReq_event_ptr->status_code), (void *)(s_ptr + status_code_pos));
+    bpf_map_update_elem(&grpc_events, &key, grpcReq_event_ptr, 0);
```
RonFed, line 195: nit: I think this update call is redundant since we have a pointer to the map element and we are updating the underlying memory, but worth making sure.

RonFed (approved):
added 2 small nits. looks great

internal/pkg/instrumentation/bpf/google.golang.org/grpc/server/bpf/probe.bpf.c
damemi: Didn't know that, thought it was a copy. Removed this line and it works thanks

PR #1134, opened 2024-09-27
Add span kind support to SDK and SDK probe

MrAlias:
- Set the span kind based on span start options
- Parse and process the span kind in the SDK probe
- Use the probe defined span kind in the controller if set 
- Add integration tests

Part of #954


internal/pkg/instrumentation/bpf/go.opentelemetry.io/auto/sdk/probe.go
```diff
@@ -151,10 +151,28 @@ func (c *converter) convertEvent(e *event) []*probe.SpanEvent {
 		TracerName:        ss.Scope().Name(),
 		TracerVersion:     ss.Scope().Version(),
 		TracerSchema:      ss.SchemaUrl(),
+		Kind:              spanKind(span.Kind()),
 		// TODO: Status.
 		// TODO: Attributes.
 		// TODO: Events.
 		// TODO: Links.
 		// TODO: Span Kind.
```
MrAlias, line 159:
```suggestion
```

RonFed (approved):

PR #1140, opened 2024-09-30
[PoC] Run integration tests using docker-compose

MrAlias:
- Updates all e2e tests so they can be run with `docker-compose`
- Adds additional make targets to run e2e tests in `docker-compose`
- Replaces indeterminate timing steps for assumed completion with deterministic signal passing and handling
  - The sample app waits for a signal that the auto-instrumentation is ready
  - The collector is shutdown when it is verified the telemetry has been generated
- A Github workflow is added to run the e2e tests using docker compose

### Test runner

Adds a `runner` utility to run testing target applications with auto-instrumentation setup. This allows us to send run the sample application and auto-instrumentation from the same process and coordinate their execution by sending signals.

### Collector

Wraps the OTel collector in an HTTP server that listens for a "shutdown" signal. It will gracefully flush and stop the collector when this signal is received

### Executioner

Examples the collector's export metrics to determine when the expected number of spans have been processed. It then sends a "shutdown" signal to the collector

## Example

```terminal
$ make e2e-autosdk e2e-databasesql e2e-gin e2e-grpc e2e-kafka-go e2e-nethttp e2e-nethttp-custom e2e-otelglobal
# Test output ...
```


pellared:
I guess it makes also debugging easier.

RonFed:
looks great.
To run this on non-Linux machines we'll need to run it inside a container, right?

MrAlias:
> looks great. To run this on non-Linux machines we'll need to run it inside a container, right?

I'm guessing, based on what @damemi said in the SIG meeting, it'll need to be run in a VM given the kernel still needs to support eBPF outside of the container.

That said, we'll probably still want to run this in a container on Linux to get sandbox the permissions issues and not have to run this with sudo.

MrAlias:
Closing in favor of https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/2118

PR #1146, opened 2024-10-02
Separate docker image build step

MrAlias:


MrAlias:
Not going to work. You cannot split off steps from a job and preserve the existing environment without having a custom runner.

PR #971, opened 2024-07-28
Support parent-based and traceID ratio samplers in eBPF

RonFed:
What is being added:
* `alwaysOn`, `alwaysOff`, `parentBased`, and `traceIDRatio` samplers in eBPF. These are configured in user space by inspecting the `OTEL_TRACES_SAMPLER` and `OTEL_TRACES_SAMPLER_ARG` environment variables. The configuration is passed to the eBPF code using maps.
* `make test_ebpf` target aims to run all the tests which are (currently) loading maps and (in the future) loading programs. This target needs to be run as a privileged user.

What is removed:
* The `WithSampler` `InstrumentationOption` is removed since it is now misleading and doesn't really act as a sampler since the sampling decision was made after the spans have finished.

Issue #953, opened 2024-07-16
Sampling support.


RonFed:
This issue will track the development of sampling support.
As discussed in the SIG, the reason the sampling logic should be included in the eBPF code is that the span is created in eBPF,  and so is the span context which contains the trace flags. The trace flags should be propagated to the child spans in the same process. Hence, sampling in Go code might be too late and won't reflect the sampling correctly. This behavior is described in the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#sdk-span-creation).

- [ ] Add trace flags to span context: https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/868
- [ ] Refactor probe and map loading; https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/899.
- [ ] https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/918
- [ ] https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/949
- [ ] Support spec defined samplers
- [ ] Allow "custom" sapoling based on attributes.

MrAlias:
We should cut the scope to what we have already developed. We should try to integrate what we already have into OBI and iterate in the future on new sampling features once that is complete.

PR #1150, opened 2024-10-05
Implement traceID-ratio and parent-based samplers in eBPF

RonFed:
Add the eBPF side of basic samplers implementation: `traceIDRatio`, `ParentBased`, `AlwaysOn` and  `AlwasyOff`.

Basic testing I did with the `httpPlusDb` example:
I used a test script that calls the endpoint of the application 1000 times.

Using `OTEL_TRACES_SAMPLER=always_on`
always_on

Using `OTEL_TRACES_SAMPLER=parentbased_traceidratio, OTEL_TRACES_SAMPLER_ARG=0.1`
parentBased-traceIDRatio-0 1

Using `OTEL_TRACES_SAMPLER=parentbased_traceidratio, OTEL_TRACES_SAMPLER_ARG=0.01`
parentBased-traceIDRatio-0 01

Split from https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/971
Part of #953


Makefile
```diff
@@ -51,13 +51,21 @@ tools: $(GOLICENSES) $(MULTIMOD) $(GOLANGCI_LINT) $(DBOTCONF) $(OFFSETGEN)
 
 ALL_GO_MODS := $(shell find . -type f -name 'go.mod' ! -path '$(TOOLS_MOD_DIR)/*' ! -path './LICENSES/*' | sort)
 GO_MODS_TO_TEST := $(ALL_GO_MODS:%=test/%)
+GO_MODS_TO_EBPF_TEST := $(ALL_GO_MODS:%=test_ebpf/%)
```
MrAlias, line 54: Sorry for the possibly naive question, but where are these files? I'm not seeing any `test_ebpf` directories?

internal/pkg/instrumentation/probe/sampling/sampling.go
```diff
@@ -108,3 +188,76 @@ func DefaultParentBasedSampler() ParentBasedConfig {
...
+func (m *Manager) applyConfig(conf *Config) error {
+	samplerIDs := make([]SamplerID, len(conf.Samplers))
+	configs := make([]SamplerConfig, len(conf.Samplers))
+
+	i := 0
+	for id, samplerConfig := range conf.Samplers {
+		samplerIDs[i] = id
+		configs[i] = samplerConfig
+		i++
+	}
```
MrAlias, line 226: nit: This is a bit confusing about the meaning of `i` and if it shouldn't have been shadowed the following for loop. It may read better if we just append here:

```suggestion
	samplerIDs := make([]SamplerID, 0, len(conf.Samplers))
	configs := make([]SamplerConfig, 0, len(conf.Samplers))

	for id, samplerConfig := range conf.Samplers {
		samplerIDs = append(samplerIDs, id)
		configs = append(configs, samplerConfig)
	}
```

This isn't blocking.

internal/pkg/instrumentation/probe/sampling/sampling_linux_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+import (
+	"testing"
+
+	"github.com/cilium/ebpf"
+	"github.com/cilium/ebpf/rlimit"
+	"github.com/stretchr/testify/assert"
+)
+
+func mockEbpfCollectionForSampling() (*ebpf.Collection, error) {
```
MrAlias, line 15: nit: eBPF is an initialism:

```suggestion
func mockEBPFCollectionForSampling() (*ebpf.Collection, error) {
```

internal/pkg/instrumentation/probe/sampling/sampling_linux_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+	return &ebpf.Collection{
+		Maps: map[string]*ebpf.Map{
+			samplersConfigMapName:     samplersConfigMap,
+			probeActiveSamplerMapName: activeSamplerMap,
+		},
+	}, nil
+}
+
+func TestEbpfNewSamplingConfigDefault(t *testing.T) {
```
MrAlias, line 55:
```suggestion
func TestEBPFNewSamplingConfigDefault(t *testing.T) {
```

MrAlias (approved):
Outside of the `Makefile` question, this looks good to go.

Makefile
RonFed: My intention here was to create a separate make target for running unit tests that require sudo (loading maps and programs).
Examples of these tests are in `internal/pkg/instrumentation/probe/sampling/sampling_linux_test.go`.

Since we have targets for `test/<mod_name>`, I figured these would be the equivalent of that to avoid conflicting target names.

internal/pkg/instrumentation/probe/sampling/sampling_linux_test.go
```diff
@@ -0,0 +1,130 @@
+//go:build ebpf_test
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
```
MrAlias, line 4: Similar to the other files with build directives, we want to keep the copyright at the top.

```suggestion
// Copyright The OpenTelemetry Authors
// SPDX-License-Identifier: Apache-2.0

//go:build ebpf_test
```

PR #1177, opened 2024-10-08
[PoC] Switch to pdata as data-model

MrAlias:


instrumentation.go
```diff
@@ -472,16 +450,16 @@ func lookupResourceData() (string, []attribute.KeyValue, bool) {
...
-// telemetry.
+// WithTraceHandler returns an [InstrumentationOption] that will configure an
+// [Instrumentation] to use the provided handler to handle OpenTelemetry
+// tracing telemetry.
 //
 // If OTEL_TRACES_EXPORTER is defined, this option will conflict with
 // [WithEnv]. If both are used, the last one provided to an [Instrumentation]
 // will be used.
-func WithTraceExporter(exp trace.SpanExporter) InstrumentationOption {
+func WithTraceHandler(handler TraceHandler) InstrumentationOption {
```
RonFed, line 460: I think we should keep having a way of just configuring the exporter without having to configure a complete handler.

PR #1194, opened 2024-10-14
Add `telemetry/test` package

MrAlias:
This package is used to verify the encode/decode of the telemetry pkg is compliant with pdata (i.e. OTLP). It is added as a separate module to isolate the `pdata` dependencies from `telemetry` which is explicitly designed to not include these dependencies.


sdk/telemetry/test/conversion_test.go
```diff
@@ -0,0 +1,247 @@
...
+	}()
+)
+
+func TestDecode(t *testing.T) {
+	var enc ptrace.JSONMarshaler
+	b, err := enc.MarshalTraces(pTraces)
+	require.NoError(t, err)
+
+	t.Log(string(b))
+
```
pellared, line 226: nit: is this needed?

```suggestion
```

sdk/telemetry/test/conversion_test.go
```diff
@@ -0,0 +1,247 @@
...
+}
+
+func TestEncode(t *testing.T) {
+	var buf bytes.Buffer
+	enc := json.NewEncoder(&buf)
+	require.NoError(t, enc.Encode(traces))
+
+	data := buf.Bytes()
+	t.Log(string(data))
+
```
pellared, line 241: nit: is this needed?

```suggestion
	t.Log(string(data))

```

pellared (commented):
How about adding `TestJSONMarshalRoundtrip` test which does something like the following:

```go
func TestJSONMarshalRoundtrip(t *testing.T) {
  data, err := json.Marshal(traces)
  require.NoError(t, err)

  var got telemetry.Traces
  err := json.Unmarshal(data, &got)
  require.NoError(t, err)

   assert.Equal(t, traces, got)
}
```

It could added to tests in `telemetry` package. However, adding it here is easier as we already have "testdata" in form or `traces` field.
This comment can be addressed in a separate PR.

MrAlias:
> How about adding `TestJSONMarshalRoundtrip` test which does something like the following:
> 
> ```go
> func TestJSONMarshalRoundtrip(t *testing.T) {
>   data, err := json.Marshal(traces)
>   require.NoError(t, err)
> 
>   var got telemetry.Traces
>   err := json.Unmarshal(data, &got)
>   require.NoError(t, err)
> 
>   assert.Equal(t, traces, got)
> }
> ```
> 
> It could added to tests in `telemetry` package. However, adding it here is easier as we already have "testdata" in form or `traces` field. This comment can be addressed in a separate PR.

I'm looking at adding this in a separate PR presently.

.github/dependabot.yml
```diff
@@ -253,6 +253,15 @@ updates:
     schedule:
       interval: weekly
       day: sunday
+  - package-ecosystem: gomod
+    directory: /sdk/telemetry/test
```
pellared, line 257: You need to also update `versions.yaml`.

sdk/telemetry/test/conversion_test.go
MrAlias: When this fails, the decode error does not always provide a clear understanding of what failed to decode. This only will be logged with verbose testing or when the failure occurs. I've found it useful when debugging the tests so far and think leaving it in may help in the future. It can be removed if you really think it shouldn't be added.

pellared: I think it is fine. I thought that could be reason. You can leave a comment that this log line is helpful when test fails.

pellared (approved):
👍

sdk/telemetry/test/conversion_test.go
pellared:
```suggestion
	t.Log(string(b)) // This helps when test fails to understand what being decoded.

```

sdk/telemetry/test/conversion_test.go
pellared:
```suggestion
	t.Log(string(data)) // This helps when test fails to understand what how the data has been encoded.

```

damemi (approved):

MrAlias:
> > How about adding `TestJSONMarshalRoundtrip` test which does something like the following:
> > ```go
> > func TestJSONMarshalRoundtrip(t *testing.T) {
> >   data, err := json.Marshal(traces)
> >   require.NoError(t, err)
> > 
> >   var got telemetry.Traces
> >   err := json.Unmarshal(data, &got)
> >   require.NoError(t, err)
> > 
> >   assert.Equal(t, traces, got)
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It could added to tests in `telemetry` package. However, adding it here is easier as we already have "testdata" in form or `traces` field. This comment can be addressed in a separate PR.
> 
> I'm looking at adding this in a separate PR presently.

https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/1196

PR #1195, opened 2024-10-14
Use `sdk/telemetry` instead of pdata for the `autosdk` probe

MrAlias:


Dockerfile
```diff
@@ -1,8 +1,10 @@
 FROM golang:1.23.2-bullseye AS base
 
+RUN apt-get update && apt-get install -y curl clang gcc llvm make libbpf-dev
+
 WORKDIR /app
 
-RUN apt-get update && apt-get install -y curl clang gcc llvm make libbpf-dev
+COPY sdk/ /app/sdk/
```
RonFed, line 7: Why is this needed?

RonFed (approved):

Dockerfile
MrAlias: Yes, correct.

PR #1196, opened 2024-10-15
Add encoding tests to the `telemetry` package

MrAlias:
- Test all types can be marshaled into expected JSON
- Test all types can be unmarshaled with CamelCase field names
- Test all types can be unmarshaled with snake_case field names

### Fixes

- Handles flags in span links
- Decodes snake_case for scopes
- Does not drop the required name of a span if it is empty
- Handles zero time values
- Omits parent span IDs if empty


pellared (approved):
🎖️

RonFed (approved):

PR #1198, opened 2024-10-15
Remove build-context from make fixtures

damemi:
This flag was added in https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/1108 and I am not sure if that was intentional or not. Running it as it is, I get the following error:

```
make[1]: Leaving directory '/usr/local/google/home/mikedame/go/src/github.com/open-telemetry/opentelemetry-go-instrumentation'                    
if [ -f ./internal/test/e2e/grpc/build.sh ]; then \
	./internal/test/e2e/grpc/build.sh; \
else \
	cd internal/test/e2e/grpc && docker build --build-context ../../../../ -t sample-app . ;\
fi
unknown flag: --build-context
See 'docker build --help'.
make: *** [Makefile:167: fixtures/grpc] Error 125
```

Changing it to `docker buildx build` showed that it's actually supposed to be a key=value pair, so it doesn't seem to be used anyway:

```
ERROR: invalid context value: ../../../../, expected key=value
make: *** [Makefile:167: fixtures/grpc] Error 1
```

I don't think this was caught because our e2es don't actually call `make fixtures`, instead they re-build essentially the same steps without the need for docker


damemi:
cc @MrAlias please correct me if I'm wrong and this has a purpose I'm missing

MrAlias (approved):
Yeah, this was a mistake to have been merged. It was left over from other testing.

PR #1207, opened 2024-10-21
Replace `probe.Event` with `ptrace.ScopeSpans`

### LINKED ISSUE #1125: [Spike] Alternate data model
We currently use the internal `Event` type as the underlying data-model of the telemetry transport.

There are many performance and convenience reasons to possibly switch to using the collector's `pdata` module as the underlying data-model.

This should be investigated.

MrAlias:
This implements part of #1177. It removes the `probe.Event` type and replaces its use with `ptrace.ScopeSpans`. It does not replace the `Controller` with a collector receiver and new `Handler` type (like #1177). The goal here is to lay the foundation from which we can build out the collector receiver and generalize the probe handler.

Resolve #1125


grcevski (approved):
LGTM! This looks good to me, it took a bit of time to review, I like the code design changes. I wasn't able to check every detail, but the tests pass and there are new unit tests for the new code.

go.mod
```diff
@@ -22,7 +22,7 @@ require (
 	github.com/hashicorp/go-version v1.7.0
 	github.com/pkg/errors v0.9.1
 	github.com/stretchr/testify v1.9.0
-	go.opentelemetry.io/auto/sdk v0.1.0-alpha
+	go.opentelemetry.io/collector/pdata v1.17.0
```
RonFed, line 25: Do you think this will have a large impact on the binary/image size?

internal/pkg/instrumentation/bpf/database/sql/probe.go
```diff
@@ -72,40 +73,33 @@ type event struct {
...
-	} else {
-		pscPtr = nil
-	}
+func processFn(pkg, ver, schemaURL string) func(*event) ptrace.ScopeSpans {
+	scopeName := "go.opentelemetry.io/auto/" + pkg
+	return func(e *event) ptrace.ScopeSpans {
+		ss := ptrace.NewScopeSpans()
+
+		scope := ss.Scope()
+		scope.SetName(scopeName)
```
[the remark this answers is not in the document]
MrAlias: Updated PTAL.

go.mod
```diff
@@ -66,12 +70,11 @@ require (
...
+	go.uber.org/multierr v1.11.0 // indirect
 	golang.org/x/exp v0.0.0-20230224173230-c95f2b4c22f2 // indirect
 	golang.org/x/net v0.30.0 // indirect
 	golang.org/x/text v0.19.0 // indirect
 	google.golang.org/genproto/googleapis/api v0.0.0-20241007155032-5fefd90f89a9 // indirect
 	google.golang.org/genproto/googleapis/rpc v0.0.0-20241007155032-5fefd90f89a9 // indirect
 	google.golang.org/protobuf v1.35.1 // indirect
 )
-
-replace go.opentelemetry.io/auto/sdk => ./sdk
```
RonFed, line 77: We can also remove this line from the Dockerfile, right?
https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/7889fea1dc4aedc8d1a8f9ce30d8c0d19b05436a/Dockerfile#L7

internal/pkg/instrumentation/probe/probe.go
```diff
@@ -215,42 +215,21 @@ func (i *Base[BPFObj, BPFEvent]) buildEBPFCollection(td *process.TargetDetails,
 }
 
 // Run runs the events processing loop.
```
RonFed, line 217: this comment needs to be updated for the `read` function

internal/pkg/instrumentation/probe/probe.go
```diff
@@ -280,6 +258,58 @@ func (i *Base[BPFObj, BPFEvent]) Close() error {
 	return err
 }
 
+type SpanProducer[BPFObj any, BPFEvent any] struct {
+	Base[BPFObj, BPFEvent]
+
+	Version   string
```
RonFed, line 264: Currently, this version is constant across all the probes that use `SpanProducer` - Do we still want to pass it through all the `New` functions?

internal/pkg/opentelemetry/controller.go
```diff
@@ -110,3 +112,153 @@ func (c *Controller) Shutdown(ctx context.Context) error {
...
+}
+
+func attr(k string, v pcommon.Value) attribute.KeyValue {
+	return attribute.KeyValue{
+		Key:   attribute.Key(k),
+		Value: val(v),
+	}
+}
+
+func val(val pcommon.Value) (out attribute.Value) {
```
RonFed, line 131: All these conversion functions will be removed once we move to a collector receiver, right?

RonFed (approved):
Added a few non-blocking comments. This looks awesome.

internal/pkg/opentelemetry/controller.go
MrAlias: Yes! That is the plan.

internal/pkg/instrumentation/probe/probe.go
MrAlias: I think we do. The end goal is to have custom probes, each of which will need to report its own version. If we design our probes to match this end goal we will avoid having to remove the coupling added when liking the version information here.

MrAlias: I'm going to resolve this. We can re-address in a follow-up PR if the need arises.

PR #1210, opened 2024-10-22
Release v0.16.0-alpha

MrAlias:
### Added

- Support `golang.org/x/net` `v0.30.0`. ([#1149](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/1149))
- Support `google.golang.org/grpc` `1.65.1`. ([#1174](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/1174))
- Support `go.opentelemetry.io/otel@v1.31.0`. ([#1178](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/1178))
- Support `google.golang.org/grpc` `1.69.0-dev`. ([#1203](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/1203))
- Implement traceID ratio and parent-based samplers. ([#1150](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/1150))
- The `go.opentelemetry.io/auto/sdk` module. This module is used directly when you want to explicilty use auto-instrumentation to process OTel API telemetry. It is also provided so the default OTel global API will use this when auto-instrumentation is loaded (WIP). ([#1210](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/1210))

### Fixed

- The `"golang.org/x/net/http2".FrameHeader.StreamID` offset for version `0.8.0` is corrected. ([#1208](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/1208))
- The `"golang.org/x/net/http2".MetaHeadersFrame.Fields` offset for version `0.8.0` is corrected. ([#1208](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/1208))

### DIFF
--- a/sdk/telemetry/test/go.mod
+++ b/sdk/telemetry/test/go.mod
@@ -4,7 +4,7 @@ go 1.22.0
 
 require (
 	github.com/stretchr/testify v1.9.0
-	go.opentelemetry.io/auto/sdk v0.16.0-alpha
+	go.opentelemetry.io/auto/sdk v0.1.0-alpha
 	go.opentelemetry.io/collector/pdata v1.17.0
 )
 
--- a/sdk/telemetry/test/go.mod
+++ b/sdk/telemetry/test/go.mod
@@ -4,7 +4,7 @@ go 1.22.0
 
 require (
 	github.com/stretchr/testify v1.9.0
-	go.opentelemetry.io/auto/sdk v0.0.0-00010101000000-000000000000
+	go.opentelemetry.io/auto/sdk v0.16.0-alpha
 	go.opentelemetry.io/collector/pdata v1.17.0
 )
 
--- a/versions.yaml
+++ b/versions.yaml
@@ -7,7 +7,6 @@ module-sets:
     modules:
       - go.opentelemetry.io/auto
       - go.opentelemetry.io/auto/sdk
-      - go.opentelemetry.io/auto/sdk/telemetry/test
 excluded-modules:
   - github.com/hashicorp/go-version
   - go.opentelemetry.io/auto/examples
@@ -23,3 +22,4 @@ excluded-modules:
   - go.opentelemetry.io/auto/internal/test/e2e/nethttp
   - go.opentelemetry.io/auto/internal/test/e2e/otelglobal
   - go.opentelemetry.io/auto/internal/tools
+  - go.opentelemetry.io/auto/sdk/telemetry/test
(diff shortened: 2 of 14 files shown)


versions.yaml
```diff
@@ -3,9 +3,11 @@
 
 module-sets:
   alpha:
-    version: v0.15.0-alpha
+    version: v0.16.0-alpha
     modules:
       - go.opentelemetry.io/auto
+      - go.opentelemetry.io/auto/sdk
+      - go.opentelemetry.io/auto/sdk/telemetry/test
```
RonFed, line 10: Do we have to publish the test module?

sdk/telemetry/test/go.mod
```diff
@@ -4,7 +4,7 @@ go 1.22.0
 
 require (
 	github.com/stretchr/testify v1.9.0
-	go.opentelemetry.io/auto/sdk v0.0.0-00010101000000-000000000000
+	go.opentelemetry.io/auto/sdk v0.16.0-alpha
```
RonFed, line 7: Just want to make sure I understood this correctly:
We have the `replace` for local builds/tests and this entry here will be used once the API is using this module?

RonFed (approved):

sdk/telemetry/test/go.mod
MrAlias: Correct. The replace only applies to this test module when using it directly. If this were ever imported the importer would not have this replace statement and would use `go.opentelemetry.io/auto/sdk v0.16.0-alpha` as an indirect dependency.

versions.yaml
MrAlias: Updated. PTAL.

MrAlias:
@RonFed please take another look. I removed the `test` module, but I also split the auto and sdk modules to be different versions. I think this split will help prevent unneeded releases and independent versioning for the future.

isZumpo:
The docker build appears to have [failed](https://github.com/open-telemetry/opentelemetry-go-instrumentation/actions/runs/11484100804/job/31961389082). I'm guessing that it got stuck somehow since it ran for 8 hours.

RonFed:
@isZumpo I think it succeeded now (I re-ran it).

isZumpo:
> @isZumpo I think it succeeded now (I re-ran it).

Thanks! I can confirm that the docker image is available now 👍

PR #1213, opened 2024-10-24
Update Makefile to only generate what is needed 

MrAlias:
Use depinfo files and make targets ensure that generate is only run when needed.

Additionally, clean up the generate commands by using the environment variables for configuration.


Makefile
```diff
@@ -59,11 +63,35 @@ test: generate $(ALL_GO_MODS:%=test/%)
...
+%/bpf_no_tp_arm64_bpfel.o: generate/%
+	$(info building $@)
+
 .PHONY: generate
-generate: export CFLAGS := $(BPF_INCLUDE)
-generate: go-mod-tidy
-generate:
-	$(GOCMD) generate ./...
+generate: go-mod-tidy $(PROBE_GEN)
+generate/%:
```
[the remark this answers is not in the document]
MrAlias: Updated.

Makefile
```diff
@@ -57,10 +61,36 @@ test: generate $(ALL_GO_MODS:%=test/%)
...
+	$(info building $@)
+
+%/bpf_arm64_bpfel.o: generate/%
+	$(info building $@)
+
+%/bpf_no_tp_x86_bpfel.o: generate/%
+	$(info building $@)
+
+%/bpf_no_tp_arm64_bpfel.o: generate/%
+	$(info building $@)
```
RonFed, line 87: If in the future we'll add another compilation variant we will need to add another target here. Is there a way to capture all the relevant files in one target (using regex)?

Makefile
```diff
@@ -57,10 +61,36 @@ test: generate $(ALL_GO_MODS:%=test/%)
...
+
+.PHONY: generate generate/all
+generate: go-mod-tidy $(PROBE_GEN)
+generate/%:
+	$(GOCMD) generate ./$*...
+generate/all:
```
RonFed, line 93: Do we want to call `generate/all` in CI and Dockerfile? If I understand this correctly, as long as there is no caching it should not make a difference. But maybe in the CI we want to make sure we generate all the files?

Makefile
```diff
@@ -57,10 +61,36 @@ test: generate $(ALL_GO_MODS:%=test/%)
...
-generate: export CFLAGS := $(BPF_INCLUDE)
-generate: go-mod-tidy
-generate:
+PROBE_ROOT = internal/pkg/instrumentation/bpf/
+PROBE_DIRS = $(shell find $(PROBE_ROOT) -type d -exec sh -c '(ls -p "{}"|grep />/dev/null)||dirname "{}"' \;)
+PROBE_GEN := $(addsuffix /bpf_x86_bpfel.go,$(PROBE_DIRS))
+PROBE_GEN += $(addsuffix /bpf_arm64_bpfel.go,$(PROBE_DIRS))
+PROBE_GEN := $(wildcard $(PROBE_GEN))
+PROBE_GEN += internal/pkg/instrumentation/bpf/net/http/client/bpf_no_tp_arm64_bpfel.go
+PROBE_GEN += internal/pkg/instrumentation/bpf/net/http/client/bpf_no_tp_x86_bpfel.go
```
RonFed, line 70: This is similar to other comments: If we add more compilation variants this will need to be updated.

Makefile
MrAlias: I think if we checked these objects in to git that would be a good idea, but without that, I don't see a need.

FWIW, I've been reconsidering if we should be checking these into git. But that is a topic we can discuss outside of this PR.

Makefile
MrAlias: I mean, sure, we can match all object files. Just, that will possibly also cause conflict with targets we don't want to match in the future. It seems like an issue that will need to be addressed in the future either way and being explicit seemed the more conservative approach.

I can change to the permissive case if that is desired though.

grcevski (approved):
LGTM! Nice!

Makefile
```diff
@@ -57,10 +61,21 @@ test: generate $(ALL_GO_MODS:%=test/%)
...
-generate:
+PROBE_ROOT = internal/pkg/instrumentation/bpf/
+PROBE_GEN_GO := $(shell find $(PROBE_ROOT) -type f -name 'bpf_*_bpfe[lb].go')
+PROBE_GEN_OBJ := $(PROBE_GEN_GO:.go=.o)
+
+# Include all depinfo files to ensure we only re-generate when needed.
+-include $(shell find $(PROBE_ROOT) -type f -name 'bpf_*_bpfel.go.d')
+
+.PHONY: generate generate/all
+generate: go-mod-tidy $(PROBE_GEN_GO)
```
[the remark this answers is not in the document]
MrAlias: Updated, PTAL.

RonFed (approved):

PR #1214, opened 2024-10-24
Improve multi-platform builds

RonFed:
Use cross-compilation instead of emulation for multi-platform builds.

reference: https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.23.2-bookworm AS base
+FROM  --platform=$BUILDPLATFORM golang:1.23.2-bookworm AS base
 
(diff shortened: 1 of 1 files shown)


damemi (approved):
Thanks for the helpful context link!

Dockerfile
```diff
@@ -12,11 +12,13 @@ COPY go.mod go.sum ./
 RUN --mount=type=cache,target=/go/pkg \
     go mod download && go mod verify
 
-FROM base as builder
+FROM --platform=$BUILDPLATFORM base as builder
```
MrAlias, line 15:
```suggestion
FROM --platform=$BUILDPLATFORM base AS builder
```

PR #1215, opened 2024-10-24
[chore] Remove unused variables in Makefile

MrAlias:


Makefile
```diff
@@ -13,8 +13,6 @@ BPF_INCLUDE += -I${REPODIR}/internal/include
 
 # Go default variables
 GOCMD?= go
-GOOS=linux
-CGO_ENABLED=0
```
RonFed, line 17: I think CGO_ENABLED default value is 1, so this might have an impact. We used to have a CGO dependency in the past, but I remember removing it made cross-compiling easier.

MrAlias: Yeah, without `export`ing it it is only defined within the make scope. And it is never used within this file. Same for GOOS for what its worth.

damemi (approved):

PR #1232, opened 2024-10-30
Use CGO_ENBABLED=0 for build

RonFed:
We used to have a dependency that required CGO_ENABLED - this was removed in https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/624.
According to https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/1215, it seems we are building with CGO_ENABLED (even though we don't have to).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -93,7 +93,7 @@ golangci-lint/%: | $(GOLANGCI_LINT)
 
 .PHONY: build
 build: generate
-	$(GOCMD) build -o otel-go-instrumentation ./cli/...
+	CGO_ENABLED=0 $(GOCMD) build -o otel-go-instrumentation ./cli/...
 
 .PHONY: docker-build
 docker-build:
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,7 @@ BPF_INCLUDE += -I${REPODIR}/internal/include
 
 # Go default variables
 GOCMD?= go
+CGO_ENABLED?=0
 
 .DEFAULT_GOAL := precommit
 
@@ -93,7 +94,7 @@ golangci-lint/%: | $(GOLANGCI_LINT)
 
 .PHONY: build
 build: generate
-	CGO_ENABLED=0 $(GOCMD) build -o otel-go-instrumentation ./cli/...
+	CGO_ENABLED=$(CGO_ENABLED) $(GOCMD) build -o otel-go-instrumentation ./cli/...
 
 .PHONY: docker-build
 docker-build:


MrAlias:
Why do we want to change this now?

RonFed:
> Why do we want to change this now?

Updated the description with some context.

Makefile
```diff
@@ -93,7 +93,7 @@ golangci-lint/%: | $(GOLANGCI_LINT)
 
 .PHONY: build
 build: generate
-	$(GOCMD) build -o otel-go-instrumentation ./cli/...
+	CGO_ENABLED=0 $(GOCMD) build -o otel-go-instrumentation ./cli/...
```
MrAlias, line 96: Can we change this to use a conditionally set var? E.g.

```make
CGO_ENABLED?=0

.PHONY: echo
echo:
	CGO_ENABLED=$(CGO_ENABLED) go env CGO_ENABLED
```

That way it can still be set from the command line if wanted, just it would have a different default:

```terminal
$ make echo
CGO_ENABLED=0 go env CGO_ENABLED
0
$ CGO_ENABLED=1 make echo
CGO_ENABLED=1 go env CGO_ENABLED
1
```

pellared:
> Why do we want to change this now?

Do we want to change

https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/77e8de9e0f4e33afb00bacb77917bad4f3056d1f/Dockerfile#L23

to `FROM scratch` (or something else which is lighter than the current "base" image?

RonFed:
> > Why do we want to change this now?
> 
> Do we want to change
> 
> https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/77e8de9e0f4e33afb00bacb77917bad4f3056d1f/Dockerfile#L23
> 
> to `FROM scratch` (or something else which is lighter than the current "base" image?


> > Why do we want to change this now?
> 
> Do we want to change
> 
> https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/77e8de9e0f4e33afb00bacb77917bad4f3056d1f/Dockerfile#L23
> 
> to `FROM scratch` (or something else which is lighter than the current "base" image?

Sure. I think that image is there because we used to have some glibc requirements in the past.
Should we do this in a separate PR?

pellared:
> Sure. I think that image is there because we used to have some glibc requirements in the past.
> Should we do this in a separate PR?

I think it would be cleaner as a separate PR.

Makefile
RonFed: Updated, PTAL

pellared (approved):

grcevski (approved):
LGTM!

MrAlias (approved):

PR #1235, opened 2024-10-31
Support gRPC status for versions >=1.40

### LINKED ISSUE #1225: It needs to support retrieving status codes in gRPC client versions prior to v1.40.
In the file `/bpf/google.golang.org/grpc/client/bpf/probe.bpf.c`, the `uprobe_ClientConn_Invoke_Returns` method reads the return value from the Error.s.s.Code field as the status code.

However, the structure of google.golang.org/grpc/internal/status.Error varies across different versions:

For v1.30.0 - v1.39.1 ([link](https://github.com/grpc/grpc-go/blob/v1.31.x/internal/status/status.go)):

```
type Error struct {
    e *spb.Status
}
```

For v1.29.0 - v1.29.1 ([link](https://github.com/grpc/grpc-go/blob/v1.29.x/internal/status/status.go)):

```
type Error spb.Status
```

The structure spb.Status is defined in google.golang.org/genproto/googleapis/rpc/status.Status.

For v1.8.0 - v1.28.1, the returned Error is actually the struct google.golang.org/grpc/status.statusError same as v1.29.0 - v1.29.1 ([link](https://github.com/grpc/grpc-go/blob/v1.16.x/status/status.go)):

```
type statusError spb.Status
```
Therefore, the logic to read the status code should be implemented according to the different structures.

In offset_results.json, add offset definitions for google.golang.org/grpc/internal/status.Error:

google.golang.org/grpc/internal/status.Error.e
google.golang.org/grpc/internal/status.Error.Code

And for google.golang.org/grpc/status.statusError:

google.golang.org/grpc/status.statusError.Code

In the uprobe_ClientConn_Invoke_Returns method, retrieve the corresponding status code by checking for non-null addresses in Error.s, Error.e, and Error.Code.

damemi:
The `Status` fields for gRPC have changed several times. Since we only recently added support for parsing the gRPC status into auto-instrumentation, I think it is reasonable to add a gate to only support the Status for recent versions (that have the same field format), rather than adding various checks for every version of gRPC.

Working on testing this manually

Fixes #1225

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ OpenTelemetry Go Automatic Instrumentation adheres to [Semantic Versioning](http
 ### Fixed
 
 - Sporadic shutdown deadlock. ([#1220](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/1220))
+- Only support gRPC status codes for gRPC >= 1.40. ([#1235](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/1235))
 
 ## [v0.16.0-alpha] - 2024-10-22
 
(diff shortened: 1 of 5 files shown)


MrAlias (approved):

MrAlias:
Can you add a changelog mentioning the fix and what support restrictions this introduces?

RonFed (approved):

PR #1239, opened 2024-11-04
offsetgen: support Docker username and password

### LINKED ISSUE #435: Support docker credentials in `offsetgen`
**Is your feature request related to a problem? Please describe.**

When a developer runs a full offset update, or tries to add a new offset, on a new system without a cache of all the docker images of Go Dockerhub will rate-limit their requests for containers. This is something like 100 pulls a day, but more requests are needed for a completely new setup.

Support user authentication in `offsetgen`. If it is set in the CLI or via the environment use that auth to pull images so the rate limit is not a blocker.

**Describe alternatives you've considered**

Currently you can work around this by authenticating with the docker CLI and just manually pulling all the images you need.

**Additional context**

https://docs.docker.com/docker-hub/download-rate-limit/

damemi:
Fixes #435

Allows you to set `DOCKER_USERNAME` and `DOCKER_PASSWORD` to authenticate with docker hub on pull. Note that a PAT can also be used for password


RonFed (approved):
Looks great.
Do we need to update the Makefile for this to work with something like:
`DOCKER_USERNAME=<username> DOCKER_PASSWORD=<password> make docker-offsets`?

MrAlias (approved):

PR #1242, opened 2024-11-05
Add gRPC server address and port

### LINKED ISSUE #172: Audit and update `google/golang/org/grpc` instrumentation to comply with RPC semantic conventions

damemi:
Fixes #172


damemi:
@MrAlias @RonFed ready for review

internal/pkg/instrumentation/bpf/google.golang.org/grpc/server/probe.go
```diff
@@ -184,6 +233,11 @@ func processFn(e *event) ptrace.SpanSlice {
 		}
 	}
 
+	if serverAddr {
+		attrs = append(attrs, semconv.ServerAddress(net.IP(e.LocalAddr.IP[:]).String()))
+		attrs = append(attrs, semconv.ServerPort(int(e.LocalAddr.Port)))
```
MrAlias, line 238: This `int64`->`int` has the potential for a truncation error. Given ports should always fit in an `int32` can we use an `int32` for the `Port` field instead?

internal/pkg/instrumentation/bpf/google.golang.org/grpc/server/probe.go
```diff
@@ -141,11 +165,36 @@ func (w writeStatusConst) InjectOption(td *process.TargetDetails) (inject.Option
...
 	context.BaseSpanProperties
 	Method     [100]byte
 	StatusCode int32
+	LocalAddr  NetAddr
+}
+
+type NetAddr struct {
+	IP   [16]uint8
+	Port uint64
+	IsV6 uint64
```
MrAlias, line 197: This doesn't ever seem to be used. Should it be removed?

internal/pkg/inject/offset_results.json
```diff
@@ -3798,25 +3798,808 @@
...
         ]
-      }
-    ]
-  },
-  {
-    "module": "std",
-    "packages": [
+      },
       {
-        "package": "bufio",
```
MrAlias, line 3810: I might be miss-reading the diff, but it seems like the `"bufio".Writer.buf` offset is being removed here. Is that intentional?

damemi: That threw me off too, but I think it was just moved above (if you search for it it's still there, please double check me)

MrAlias: Ah, yeah, you are correct. All good. :+1:

internal/pkg/instrumentation/bpf/google.golang.org/grpc/server/probe.go
damemi: Seems to work in my latest commit

internal/pkg/instrumentation/bpf/google.golang.org/grpc/server/probe.go
damemi: Yup think we can remove it

internal/test/e2e/grpc/verify.bats
```diff
@@ -57,8 +67,8 @@ SCOPE="go.opentelemetry.io/auto/google.golang.org/grpc"
   server_trace_id=$(server_spans_from_scope_named ${SCOPE} | jq ".traceId" | jq -Rn '[inputs]' | jq -r .[0])
   assert_equal "$server_trace_id" "$client_trace_id"
 
-  client_trace_id=$(client_spans_from_scope_named ${SCOPE} | jq ".traceId" | jq -Rn '[inputs]' | jq -r .[0])
-  server_trace_id=$(server_spans_from_scope_named ${SCOPE} | jq ".traceId" | jq -Rn '[inputs]' | jq -r .[0])
+  client_trace_id=$(client_spans_from_scope_named ${SCOPE} | jq ".traceId" | jq -Rn '[inputs]' | jq -r .[1])
+  server_trace_id=$(server_spans_from_scope_named ${SCOPE} | jq ".traceId" | jq -Rn '[inputs]' | jq -r .[1])
```
MrAlias, line 71: Why did this change?

MrAlias (approved):

internal/test/e2e/grpc/verify.bats
damemi: I noticed this test was wrong while I was adding the new ones. This is supposed to check the first and second spans. The lines directly above this are already checking `[0]`, so this was supposed to be `[1]`. Figured I'd fix it here

PR #1245, opened 2024-11-06
Split Instrumentation.Run to Load and Run

### LINKED ISSUE #1240: Improve Instrumentation.Run API
Currently, our Run function is:
1. Blocking
2. Has 2 main parts - 
         a) Initializing the probes - loading and attaching them.
         b) Running the main events loop.
After the first part is done we close the loaded indicator.

The main disadvantage to this design is the error reporting to the user.
A typical use case might be:

```Go
	go func() {
		err := inst.Run(ctx)
		if err != nil {
			Logger.Error(err, "failed to run instrumentation")
		}
		runError <- err
	}()

       // we want to make sure the instrumentation is loaded before we continue

	select {
	case <-loadedIndicator:
	     Logger.Info("instrumentation loaded")
	case <-runError:
            // error returned from run - this can be either a loading error or a closing error
	}
```

One possible solution for this is to split the `Run` function into `Load` and `Run` - this way we can remove the loaded indicator. 
`Load` won't be blocking and will do the first part of what `Run` is doing toady.
In addition, the errors returned from Load and Run are well-defined and won't have possible double meaning.

After this change the above code can be changed to something like:
```Go
        err := inst.Load(ctx)
        if err != nil {
              return err
        }

         // the instrumentation is loaded and ready to send data

	go func() {
		err := inst.Run(ctx)
		if err != nil {
			Logger.Error(err, "failed to run instrumentation")
		}

               // the instrumentation is closed, report possible closing errors
                if err != nil {
		         closingErrors <- err
               }
	}()

     //Continue doing other stuff, possibly listening in the background on the closingErrors channel for closing errors

```

RonFed:
Resolves #1240

* Split Instrumentation.Run to 2 parts - Load and Run.
* Remove loadedIndicator.
* Add `Manager.Stop` which should be called if `Instrumentation.Close` is called in between Load and Run.
* Improve state handling inside the manager and protect it with a mutex.


instrumentation.go
```diff
@@ -534,17 +540,6 @@ func WithResourceAttributes(attrs ...attribute.KeyValue) InstrumentationOption {
...
-// [Instrumentation] to close the provided indicator channel when the target
-// process has been instrumented (i.e. all probes have been loaded).
-// The provided indicator channel needs to be initialized by the caller.
-func WithLoadedIndicator(indicator chan struct{}) InstrumentationOption {
-	return fnOpt(func(_ context.Context, c instConfig) (instConfig, error) {
-		c.loadIndicator = indicator
-		return c, nil
-	})
-}
-
```
MrAlias, line 547: Please include this removal in its own changelog entry.

internal/pkg/instrumentation/manager.go
```diff
@@ -306,6 +308,26 @@ func (m *Manager) Run(ctx context.Context, target *process.TargetDetails) error
 	return <-done
 }
 
+// Stop stops all probes and cleans up all the resources associated with them.
+func (m *Manager) Stop() error {
```
MrAlias, line 312: Calling this will not stop `Run`. `Run` is waiting for the context to be done. This will clear probes without having that process end.

internal/pkg/instrumentation/manager.go
```diff
@@ -306,6 +308,26 @@ func (m *Manager) Run(ctx context.Context, target *process.TargetDetails) error
...
+
+	// Wait for all probes to stop before closing the chan they send on.
+	m.runningProbesWG.Wait()
+	close(m.telemetryCh)
+
+	m.state = managerStateStopped
+	return err
+}
+
 func (m *Manager) load(target *process.TargetDetails) error {
```
MrAlias, line 331: It seems a bit odd to have both `Load` and `load`. Should these methods be unified?

internal/pkg/instrumentation/manager.go
MrAlias: I would expect most users would want the `Manager` to stop if they call this, meaning the `Run` call should return. Given you have state checking added here, I think we could wrap the passed context to `Run` and sync the stop methods.

For example, we can add a `stop` field to the `Manager`.

```diff
--- a/internal/pkg/instrumentation/manager.go
+++ b/internal/pkg/instrumentation/manager.go
@@ -58,6 +58,7 @@ type Manager struct {
        cp              ConfigProvider
        exe             *link.Executable
        td              *process.TargetDetails
+       stop            context.CancelCauseFunc
        runningProbesWG sync.WaitGroup
        telemetryCh     chan ptrace.ScopeSpans
        currentConfig   Config
```

And then in `Run`:

```diff
@@ -283,6 +284,8 @@ func (m *Manager) Run(ctx context.Context) error {
                return errors.New("manager is not loaded, call Load before Run")
        }

+       ctx, m.stop = context.WithCancelCause(ctx)
+
        for id, p := range m.probes {
                if isProbeEnabled(id, m.currentConfig) {
                        m.runProbe(p)
@@ -299,7 +302,10 @@ func (m *Manager) Run(ctx context.Context) error {
                <-ctx.Done()

                err := m.Stop()
-               done <- errors.Join(err, ctx.Err())
+               if e := context.Cause(ctx); !errors.Is(e, errStop) {
+                       err = errors.Join(err, e)
+               }
+               done <- err
        }()

        for e := range m.telemetryCh {
```

Then we can close this child context when `Stop` is called:

```diff
@@ -308,12 +314,16 @@ func (m *Manager) Run(ctx context.Context) error {
        return <-done
 }

+var errStop = errors.New("stopped called")
+
 // Stop stops all probes and cleans up all the resources associated with them.
 func (m *Manager) Stop() error {
        if m.state == managerStateUninitialized || m.state == managerStateStopped {
                return nil
        }

+       m.stop(errStop)
```

internal/pkg/instrumentation/manager.go
RonFed: renamed the internal function to `loadProbes`

internal/pkg/instrumentation/bpffs/bpfsfs_other.go
```diff
@@ -14,9 +18,15 @@ func PathForTargetApplication(target *process.TargetDetails) string {
...
+	if target == nil {
+		return errors.New("target is nil in Mount")
+	}
 	return nil
 }
 
 func Cleanup(target *process.TargetDetails) error {
+	if target == nil {
+		return errors.New("target is nil in Cleanup")
+	}
```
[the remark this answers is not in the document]
RonFed: Right, removed

CHANGELOG.md
```diff
@@ -8,6 +8,12 @@ OpenTelemetry Go Automatic Instrumentation adheres to [Semantic Versioning](http
 
 ## [Unreleased]
 
+### Changed
+
+- Split the functionality of `Instrumentation.Run` to `Instrumentation.Load` and `Instrumentation.Run`.
+  `Load` will report any errors in the loading and attaching phase of the probes. ([#1245](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/1245))
+- Remove the `WithLoadedIndicator` `InstrumentationOption` since the `Instrumentation.Load` will indicate whether the probes are loaded in a synchronous way.  ([#1245](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/1245))
```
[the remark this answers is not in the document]
RonFed: Fixed

instrumentation.go
```diff
@@ -126,7 +126,12 @@ func NewInstrumentation(ctx context.Context, opts ...InstrumentationOption) (*In
 	}, nil
 }
 
-// Run starts the instrumentation.
+// Load loads and attaches the relevant probes to the target process.
+func (i *Instrumentation) Load(ctx context.Context) error {
+	return i.manager.Load(ctx, i.target)
+}
+
+// Run starts the instrumentation. It must be called after [Load].
```
MrAlias, line 134:
```suggestion
// Run starts the instrumentation. It must be called after [Instrumentation.Load].
```

MrAlias (approved):

PR #1246, opened 2024-11-06
Don't include db.query if it is empty

RonFed:
If the `OTEL_GO_AUTO_INCLUDE_DB_STATEMENT` environment variable is not set, we won't collect the SQL query, hence in that case there is no point in adding an empty attribute.

Related to https://github.com/odigos-io/odigos/issues/1695


damemi (approved):
+1 for bugfix in changelog

Issue #743, opened 2024-04-01
Audit and update `database/sql` instrumentation to comply with SQL semantic conventions


damemi:
https://github.com/open-telemetry/semantic-conventions/blob/a04a1f4ba413d0a1f137c223fc243ea14fdcd89b/docs/database/sql.md

damemi:
Latest semconv: https://github.com/open-telemetry/semantic-conventions/blob/21b8f4620d2029fe8abb98a4a06655da3232d11d/docs/database/sql.md

Required attributes:
* `db.collection.name` (“if readily available. The collection name MAY be parsed from the query text, in which case it SHOULD be the first collection name found in the query.”)
* `db.namespace` (“if available”)
* `db.operation.name` (“if readily available. The operation name MAY be parsed from the query text, in which case it SHOULD be the first operation name found in the query.)
* `error.type` (“if and only if the operation failed”)
* `server.port` (If using a port other than the default port for this DBMS and if `server.address` is set.

Client calls semconv: https://github.com/open-telemetry/semantic-conventions/blob/21b8f4620d2029fe8abb98a4a06655da3232d11d/docs/database/database-spans.md

> ## Name
>
> Database spans MUST follow the overall [guidelines for span names](https://github.com/open-telemetry/opentelemetry-specification/tree/v1.33.0/specification/trace/api.md#span).
>
> The **span name** SHOULD be `{db.operation.name} {target}` if there is a
(low-cardinality) `{db.operation.name}` available (see below for the exact definition of the [`{target}`](#target-placeholder) placeholder).
>
> If there is no (low-cardinality) `db.operation.name` available, database span names
SHOULD be [`{target}`](#target-placeholder).
>
> If neither `{db.operation.name}` nor `{target}` are available, span name SHOULD be `{db.system}`.
>
> Semantic conventions for individual database systems MAY specify different span name format.
> 
> The <span id="target-placeholder">`{target}`</span> SHOULD describe the entity that the operation is performed against
and SHOULD adhere to one of the following values, provided they are accessible:
>
> - `db.collection.name` SHOULD be used for data manipulation operations or operations on database collections.
> - `db.namespace` SHOULD be used for operations on a specific database namespace.
> - `server.address:server.port` SHOULD be used for other operations not targeting any specific database(s) or collection(s)
>
>If a corresponding `{target}` value is not available for a specific operation, the instrumentation SHOULD omit the `{target}`.
[...]

PR #1253, opened 2024-11-07
Support SELECT/INSERT/UPDATE/DELETE db operation + span names

damemi:
Ref #743 

From the [semconv for db spans](https://github.com/open-telemetry/semantic-conventions/blob/e0e93bad394a15161af7942d65d197defac7b88b/docs/database/database-spans.md#name):

```
Database spans MUST follow the overall [guidelines for span names](https://github.com/open-telemetry/opentelemetry-specification/tree/v1.37.0/specification/trace/api.md#span).

The span name SHOULD be {db.query.summary} if a summary is available.

If no summary is available, the span name SHOULD be {db.operation.name} {target} provided that a (low-cardinality) db.operation.name is available (see below for the exact definition of the [{target}](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/database-spans.md#target-placeholder) placeholder).

If a (low-cardinality) db.operation.name is not available, database span names SHOULD default to the [{target}](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/database-spans.md#target-placeholder).

If neither {db.operation.name} nor {target} are available, span name SHOULD be {db.system}.

Semantic conventions for individual database systems MAY specify different span name format.

The {target} SHOULD describe the entity that the operation is performed against and SHOULD adhere to one of the following values, provided they are accessible:

db.collection.name SHOULD be used for data manipulation operations or operations on a database collection.
db.namespace SHOULD be used for operations on a specific database namespace.
server.address:server.port SHOULD be used for other operations not targeting any specific database(s) or collection(s)
If a corresponding {target} value is not available for a specific operation, the instrumentation SHOULD omit the {target}. For example, for an operation describing SQL query on an anonymous table like SELECT * FROM (SELECT * FROM table) t, span name should be SELECT.
```

We don't have `collection` (table), `namespace`, or `server.address` readily available (right now), but we can at least parse certain operations. We can also support more over time. So this gets us at least a couple useful span names besides `DB` (fallback for unsupported operations)

This means we can also set [`db.operation.name`](https://github.com/open-telemetry/semantic-conventions/blob/e0e93bad394a15161af7942d65d197defac7b88b/docs/database/sql.md#attributes) with that same value when it is available.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -27,6 +27,7 @@ Alternatively, you can add support for additional or different configurations by
 | Environment variable                | Description                                            | Default value |
 |-------------------------------------|--------------------------------------------------------|---------------|
 | `OTEL_GO_AUTO_INCLUDE_DB_STATEMENT` | Sets whether to include SQL queries in the trace data. |               |
+| `OTEL_GO_AUTO_INCLUDE_DB_OPERATION` | Sets whether to include SQL operation in the trace data. |               |
 
 ## Traces exporter
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -27,7 +27,7 @@ Alternatively, you can add support for additional or different configurations by
 | Environment variable                | Description                                            | Default value |
 |-------------------------------------|--------------------------------------------------------|---------------|
 | `OTEL_GO_AUTO_INCLUDE_DB_STATEMENT` | Sets whether to include SQL queries in the trace data. |               |
-| `OTEL_GO_AUTO_INCLUDE_DB_OPERATION` | Sets whether to include SQL operation in the trace data. |               |
+| `OTEL_GO_AUTO_PARSE_DB_STATEMENT` | Sets whether to parse the SQL statement for trace data, setting `db.operation.name`. Only valid if `OTEL_GO_AUTO_INCLUDE_DB_STATEMENT` is also set. |               |
 
 ## Traces exporter
 
--- a/go.mod
+++ b/go.mod
@@ -23,6 +23,8 @@ require (
 	github.com/pkg/errors v0.9.1
 	github.com/stretchr/testify v1.10.0
 	go.opentelemetry.io/collector/pdata v1.21.0
+	github.com/stretchr/testify v1.9.0
+	github.com/xwb1989/sqlparser v0.0.0-20180606152119-120387863bf2
 	go.opentelemetry.io/contrib/exporters/autoexport v0.57.0
 	go.opentelemetry.io/otel v1.32.0
 	go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracehttp v1.32.0
--- a/go.mod
+++ b/go.mod
@@ -22,9 +22,8 @@ require (
 	github.com/hashicorp/go-version v1.7.0
 	github.com/pkg/errors v0.9.1
 	github.com/stretchr/testify v1.10.0
-	go.opentelemetry.io/collector/pdata v1.21.0
-	github.com/stretchr/testify v1.9.0
 	github.com/xwb1989/sqlparser v0.0.0-20180606152119-120387863bf2
+	go.opentelemetry.io/collector/pdata v1.21.0
 	go.opentelemetry.io/contrib/exporters/autoexport v0.57.0
 	go.opentelemetry.io/otel v1.32.0
 	go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracehttp v1.32.0
--- a/internal/pkg/instrumentation/bpf/database/sql/probe.go
+++ b/internal/pkg/instrumentation/bpf/database/sql/probe.go
@@ -29,6 +29,9 @@ const (
 
 	// IncludeDBStatementEnvVar is the environment variable to opt-in for sql query inclusion in the trace.
 	IncludeDBStatementEnvVar = "OTEL_GO_AUTO_INCLUDE_DB_STATEMENT"
+
+	// IncludeDBOperationEnvVar is the environment variable to opt-in for sql query operation in the trace.
+	IncludeDBOperationEnvVar = "OTEL_GO_AUTO_INCLUDE_DB_OPERATION"
 )
 
 // New returns a new [probe.Probe].
@@ -97,24 +100,30 @@ func processFn(e *event) ptrace.SpanSlice {
 	query := unix.ByteSliceToString(e.Query[:])
 	if query != "" {
 		span.Attributes().PutStr(string(semconv.DBQueryTextKey), query)
+	}
 
-		q, err := sql.Parse(query)
-		if err == nil {
-			operation := ""
-			switch q.(type) {
-			case *sql.Select:
-				operation = "SELECT"
-			case *sql.Update:
-				operation = "UPDATE"
-			case *sql.Insert:
-				operation = "INSERT"
-			case *sql.Delete:
-				operation = "DELETE"
-			}
+	includeOperationVal := os.Getenv(IncludeDBOperationEnvVar)
+	if includeOperationVal != "" {
+		include, err := strconv.ParseBool(includeOperationVal)
+		if err == nil && include {
+			q, err := sql.Parse(query)
+			if err == nil {
+				operation := ""
+				switch q.(type) {
+				case *sql.Select:
+					operation = "SELECT"
+				case *sql.Update:
+					operation = "UPDATE"
+				case *sql.Insert:
+					operation = "INSERT"
+				case *sql.Delete:
+					operation = "DELETE"
+				}
 
-			if operation != "" {
-				span.Attributes().PutStr(string(semconv.DBOperationNameKey), operation)
-				span.SetName(operation)
+				if operation != "" {
+					span.Attributes().PutStr(string(semconv.DBOperationNameKey), operation)
+					span.SetName(operation)
+				}
 			}
 		}
 	}
--- a/internal/pkg/instrumentation/bpf/database/sql/probe.go
+++ b/internal/pkg/instrumentation/bpf/database/sql/probe.go
@@ -8,6 +8,8 @@ import (
 	"os"
 	"strconv"
 
+	sql "github.com/xwb1989/sqlparser"
+
 	"go.opentelemetry.io/collector/pdata/pcommon"
 	"go.opentelemetry.io/collector/pdata/ptrace"
 	semconv "go.opentelemetry.io/otel/semconv/v1.26.0"
@@ -95,6 +97,26 @@ func processFn(e *event) ptrace.SpanSlice {
 	query := unix.ByteSliceToString(e.Query[:])
 	if query != "" {
 		span.Attributes().PutStr(string(semconv.DBQueryTextKey), query)
+
+		q, err := sql.Parse(query)
+		if err == nil {
+			operation := ""
+			switch q.(type) {
+			case *sql.Select:
+				operation = "SELECT"
+			case *sql.Update:
+				operation = "UPDATE"
+			case *sql.Insert:
+				operation = "INSERT"
+			case *sql.Delete:
+				operation = "DELETE"
+			}
+
+			if operation != "" {
+				span.Attributes().PutStr(string(semconv.DBOperationNameKey), operation)
+				span.SetName(operation)
+			}
+		}
 	}
 
 	return spans
--- a/internal/pkg/instrumentation/bpf/database/sql/probe.go
+++ b/internal/pkg/instrumentation/bpf/database/sql/probe.go
@@ -8,7 +8,6 @@ import (
 	"log/slog"
 	"os"
 	"strconv"
-	"strings"
 
 	"github.com/xwb1989/sqlparser"
 
@@ -32,8 +31,8 @@ const (
 	// IncludeDBStatementEnvVar is the environment variable to opt-in for sql query inclusion in the trace.
 	IncludeDBStatementEnvVar = "OTEL_GO_AUTO_INCLUDE_DB_STATEMENT"
 
-	// IncludeDBOperationEnvVar is the environment variable to opt-in for sql query operation in the trace.
-	IncludeDBOperationEnvVar = "OTEL_GO_AUTO_PARSE_DB_STATEMENT"
+	// ParseDBStatementEnvVar is the environment variable to opt-in for sql query operation in the trace.
+	ParseDBStatementEnvVar = "OTEL_GO_AUTO_PARSE_DB_STATEMENT"
 )
 
 // New returns a new [probe.Probe].
@@ -104,18 +103,26 @@ func processFn(e *event) ptrace.SpanSlice {
 		span.Attributes().PutStr(string(semconv.DBQueryTextKey), query)
 	}
 
-	includeOperationVal := os.Getenv(IncludeDBOperationEnvVar)
+	includeOperationVal := os.Getenv(ParseDBStatementEnvVar)
 	if includeOperationVal != "" {
 		include, err := strconv.ParseBool(includeOperationVal)
 		if err == nil && include {
 			operation, target, err := Parse(query)
 			if err == nil {
-				span.SetName(strings.Join([]string{operation, target}, " "))
+				name := ""
 				if operation != "" {
 					span.Attributes().PutStr(string(semconv.DBOperationNameKey), operation)
+					name = operation
 				}
 				if target != "" {
 					span.Attributes().PutStr(string(semconv.DBCollectionNameKey), target)
+					if name != "" {
+						// if operation is in the name and target is available, set name to {operation} {target}
+						name += " " + target
+					}
+				}
+				if name != "" {
+					span.SetName(name)
 				}
 			}
 		}
--- a/internal/pkg/instrumentation/bpf/database/sql/probe.go
+++ b/internal/pkg/instrumentation/bpf/database/sql/probe.go
@@ -4,11 +4,13 @@
 package sql
 
 import (
+	"fmt"
 	"log/slog"
 	"os"
 	"strconv"
+	"strings"
 
-	sql "github.com/xwb1989/sqlparser"
+	"github.com/xwb1989/sqlparser"
 
 	"go.opentelemetry.io/collector/pdata/pcommon"
 	"go.opentelemetry.io/collector/pdata/ptrace"
@@ -31,7 +33,7 @@ const (
 	IncludeDBStatementEnvVar = "OTEL_GO_AUTO_INCLUDE_DB_STATEMENT"
 
 	// IncludeDBOperationEnvVar is the environment variable to opt-in for sql query operation in the trace.
-	IncludeDBOperationEnvVar = "OTEL_GO_AUTO_INCLUDE_DB_OPERATION"
+	IncludeDBOperationEnvVar = "OTEL_GO_AUTO_PARSE_DB_STATEMENT"
 )
 
 // New returns a new [probe.Probe].
@@ -106,23 +108,14 @@ func processFn(e *event) ptrace.SpanSlice {
 	if includeOperationVal != "" {
 		include, err := strconv.ParseBool(includeOperationVal)
 		if err == nil && include {
-			q, err := sql.Parse(query)
+			operation, target, err := Parse(query)
 			if err == nil {
-				operation := ""
-				switch q.(type) {
-				case *sql.Select:
-					operation = "SELECT"
-				case *sql.Update:
-					operation = "UPDATE"
-				case *sql.Insert:
-					operation = "INSERT"
-				case *sql.Delete:
-					operation = "DELETE"
-				}
-
+				span.SetName(strings.Join([]string{operation, target}, " "))
 				if operation != "" {
 					span.Attributes().PutStr(string(semconv.DBOperationNameKey), operation)
-					span.SetName(operation)
+				}
+				if target != "" {
+					span.Attributes().PutStr(string(semconv.DBCollectionNameKey), target)
 				}
 			}
 		}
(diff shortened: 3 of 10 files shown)


internal/pkg/instrumentation/bpf/database/sql/probe.go
```diff
@@ -8,6 +8,8 @@ import (
 	"os"
 	"strconv"
 
+	sql "github.com/xwb1989/sqlparser"
```
damemi, line 11: I looked through a couple sql parsing libraries, this isn't the most maintained... https://github.com/krasun/gosqlparser seems good but for some reason I don't think it supports `*` in queries (??). For what we're doing now I think it's fine but this could use another look some day

internal/pkg/instrumentation/bpf/database/sql/probe.go
```diff
@@ -95,6 +97,26 @@ func processFn(e *event) ptrace.SpanSlice {
 	query := unix.ByteSliceToString(e.Query[:])
 	if query != "" {
 		span.Attributes().PutStr(string(semconv.DBQueryTextKey), query)
+
+		q, err := sql.Parse(query)
```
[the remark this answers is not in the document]
damemi: I pushed a benchmark test, and it does look like this takes a decent performance hit to parse the queries:
```
$ go test -bench=.
goos: darwin
goarch: arm64
pkg: go.opentelemetry.io/auto/internal/pkg/instrumentation/bpf/database/sql
cpu: Apple M3 Pro
BenchmarkProcessFn/no_query_(baseline)-12         	10643458	       104.1 ns/op
BenchmarkProcessFn/simple_query-12                	  242698	      4691 ns/op
BenchmarkProcessFn/medium_query-12                	   99649	     11987 ns/op
BenchmarkProcessFn/hard_query-12                  	  551522	      2373 ns/op
PASS
ok  	go.opentelemetry.io/auto/internal/pkg/instrumentation/bpf/database/sql	5.339s
```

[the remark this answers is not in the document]
damemi: @MrAlias do you think we could make the [call to `processFunc`](https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/5007dc331c40ee253dac4edcec98d0e6a7a18530/internal/pkg/instrumentation/probe/probe.go#L286) async to keep the pipeline from getting backed up?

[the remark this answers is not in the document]
MrAlias: SIG meeting notes:

- Add a configuration option to enable this parsing. Default to not adding this attribute.

internal/pkg/instrumentation/bpf/database/sql/probe.go
```diff
@@ -27,6 +29,9 @@ const (
 
 	// IncludeDBStatementEnvVar is the environment variable to opt-in for sql query inclusion in the trace.
 	IncludeDBStatementEnvVar = "OTEL_GO_AUTO_INCLUDE_DB_STATEMENT"
+
+	// IncludeDBOperationEnvVar is the environment variable to opt-in for sql query operation in the trace.
+	IncludeDBOperationEnvVar = "OTEL_GO_AUTO_INCLUDE_DB_OPERATION"
```
MrAlias, line 34: This will need to be added here: https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/main/docs/configuration.md#instrumentation-options

internal/pkg/instrumentation/bpf/database/sql/probe.go
```diff
@@ -97,6 +102,32 @@ func processFn(e *event) ptrace.SpanSlice {
...
+					operation = "UPDATE"
+				case *sql.Insert:
+					operation = "INSERT"
+				case *sql.Delete:
+					operation = "DELETE"
+				}
+
+				if operation != "" {
+					span.Attributes().PutStr(string(semconv.DBOperationNameKey), operation)
+					span.SetName(operation)
```
MrAlias, line 125: Should we try to build the [`{db.query.summary}`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/database-spans.md#common-attributes) or `{db.operation.name} {target}` here given we have already parsed the query?

internal/pkg/instrumentation/bpf/database/sql/probe.go
MrAlias, line 34: Can we call this `OTEL_GO_AUTO_PARSE_DB_STATEMENT` instead? Give we are going to use the parsed semantics for more than just the operation it might help better convey the intention.

go.mod
```diff
@@ -22,6 +22,7 @@ require (
 	github.com/hashicorp/go-version v1.7.0
 	github.com/pkg/errors v0.9.1
 	github.com/stretchr/testify v1.9.0
+	github.com/xwb1989/sqlparser v0.0.0-20180606152119-120387863bf2
```
MrAlias, line 25: Could we instead use the code this was based on: https://github.com/vitessio/vitess/tree/main/go/vt/sqlparser

It seems like this was a fork of that repository[^1], but it hasn't been maintained or synced since that fork.

Would it be too large of a dependency to just rely on vitess directly?

[^1]: https://github.com/xwb1989/sqlparser?tab=readme-ov-file#notice

internal/pkg/instrumentation/bpf/database/sql/probe.go
damemi: updated

internal/pkg/instrumentation/bpf/database/sql/probe.go
damemi: I was originally doing `db.query.summary`, but we don't have any of the values for [`{target}`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/database-spans.md#target-placeholder) (db.collection, db.namespace, or server address). So, it recommends just using the operation:

> If a corresponding `{target}` value is not available for a specific operation, the instrumentation SHOULD omit the `{target}`. For example, for an operation describing SQL query on an anonymous table like `SELECT * FROM (SELECT * FROM table) t`, span name should be `SELECT`.

go.mod
damemi: I did think about using vitess directly, but it seemed like a lot more than what we needed. Since this is just an implementation detail, we could refactor it later if we want but I think this is the best option for right now

MrAlias: I'm concerned when looking through all the changes that have been applied to the upstream code over the past 6 years, there's a considerable amount of changes. Notably:

- There are many additional statement types that can be parsed: https://pkg.go.dev/vitess.io/vitess/go/vt/sqlparser#StatementType
- The statement type itself can print its name meaning we will not need to add an additional switch statement: https://pkg.go.dev/vitess.io/vitess/go/vt/sqlparser#StatementType.String
- The code is not owned by Google, it is licensed by a CNCF project
- https://github.com/vitessio/vitess/issues/4174
- https://github.com/vitessio/vitess/pull/2395 which addressed:
  - https://github.com/vitessio/vitess/issues/1953
  - https://github.com/vitessio/vitess/issues/2184
- https://github.com/vitessio/vitess/pull/8307

Should we make our own fork of this pacakge?

MrAlias: Can we maybe get a better understanding of dependency size by looking at what the binary size is based on the possible dependencies?

internal/pkg/instrumentation/bpf/database/sql/probe.go
damemi: Not that I could find, unfortunately :\ The first library I tried (https://github.com/krasun/gosqlparser) does have that, which is how I was building it at first. But that library breaks when it tries to parse a `*` character (like `SELECT * FROM`). Which makes no sense to me, but that's my understanding from trying to debug it and read that library's code. I'm open to add a TODO here for this though because maybe I'm just missing something

go.mod
damemi: I'll admit it looks like you've dug into it deeper than I did, hah. These are some great points so maybe that is the better approach. I'll try switching it to that

internal/pkg/instrumentation/bpf/database/sql/probe.go
MrAlias: I think this should work, right?

```go
package main

import (
	"fmt"

	"github.com/xwb1989/sqlparser"
)

// Parse takes a SQL query string and returns the parsed query statement type
// and table name, or an error if parsing failed.
func Parse(query string) (string, string, error) {
	stmt, err := sqlparser.Parse(query)
	if err != nil {
		return "", "", fmt.Errorf("failed to parse query: %w", err)
	}

	switch stmt := stmt.(type) {
	case *sqlparser.Select:
		return "SELECT", getTableName(stmt.From), nil
	case *sqlparser.Update:
		return "UPDATE", getTableName(stmt.TableExprs), nil
	case *sqlparser.Insert:
		return "INSERT", stmt.Table.Name.String(), nil
	case *sqlparser.Delete:
		return "DELETE", getTableName(stmt.TableExprs), nil
	default:
		return "UNKNOWN", "", fmt.Errorf("unsupported statement type")
	}
}

// getTableName extracts the table name from a SQL node.
func getTableName(node sqlparser.SQLNode) string {
	switch tableExpr := node.(type) {
	case sqlparser.TableName:
		return tableExpr.Name.String()
	case sqlparser.TableExprs:
		for _, expr := range tableExpr {
			if tableName, ok := expr.(*sqlparser.AliasedTableExpr); ok {
				if name, ok := tableName.Expr.(sqlparser.TableName); ok {
					return name.Name.String()
				}
			}
		}
	}
	return ""
}

func main() {
	queries := []string{
		"SELECT * FROM users WHERE id = 1",
		"SELECT id, name FROM users WHERE id = 1",
		"INSERT INTO orders (id, amount) VALUES (1, 100)",
		"UPDATE products SET price = 19.99 WHERE id = 10",
		"DELETE FROM sessions WHERE expired = true",
		"CREATE TABLE logs (id INT, message TEXT)",
	}
	for _, query := range queries {
		fmt.Println("Query: ", query)
		statement, table, err := Parse(query)
		if err != nil {
			fmt.Println("Error:", err)
			continue
		}
		fmt.Printf("Statement: %s, Table: %s\n", statement, table)
	}
}
```

FWIW, looking at vitess, it seems like it should be able to support more statement types in the table lookup (i.e. table create).

MrAlias: Using `"vitess.io/vitess/go/vt/sqlparser"`:

```go
package main

import (
	"fmt"

	"vitess.io/vitess/go/vt/sqlparser"
)

// Parse takes a SQL query string and returns the parsed query statement type
// and table name(s), or an error if parsing failed.
func Parse(query string) (string, []string, error) {
	p, err := sqlparser.New(sqlparser.Options{})
	if err != nil {
		return "", nil, fmt.Errorf("failed to create parser: %w", err)
	}

	stmt, err := p.Parse(query)
	if err != nil {
		return "", nil, fmt.Errorf("failed to parse query: %w", err)
	}

	var statementType string
	var tables []string

	switch stmt := stmt.(type) {
	case *sqlparser.Select:
		statementType = "SELECT"
		tables = extractTables(stmt.From)
	case *sqlparser.Update:
		statementType = "UPDATE"
		tables = extractTables(stmt.TableExprs)
	case *sqlparser.Insert:
		statementType = "INSERT"
		tables = []string{stmt.Table.TableNameString()}
	case *sqlparser.Delete:
		statementType = "DELETE"
		tables = extractTables(stmt.TableExprs)
	case *sqlparser.CreateTable:
		statementType = "CREATE TABLE"
		tables = []string{stmt.Table.Name.String()}
	case *sqlparser.AlterTable:
		statementType = "ALTER TABLE"
		tables = []string{stmt.Table.Name.String()}
	case *sqlparser.DropTable:
		statementType = "DROP TABLE"
		for _, table := range stmt.FromTables {
			tables = append(tables, table.Name.String())
		}
	case *sqlparser.CreateDatabase:
		statementType = "CREATE DATABASE"
		tables = []string{stmt.DBName.String()}
	case *sqlparser.DropDatabase:
		statementType = "DROP DATABASE"
		tables = []string{stmt.DBName.String()}
	case *sqlparser.TruncateTable:
		statementType = "TRUNCATE TABLE"
		tables = []string{stmt.Table.Name.String()}
	default:
		return "UNKNOWN", nil, fmt.Errorf("unsupported statement type")
	}

	return statementType, tables, nil
}

// extractTables extracts table names from a list of SQL nodes.
func extractTables(exprs sqlparser.TableExprs) []string {
	var tables []string
	for _, expr := range exprs {
		switch tableExpr := expr.(type) {
		case *sqlparser.AliasedTableExpr:
			if name, ok := tableExpr.Expr.(sqlparser.TableName); ok {
				tables = append(tables, name.Name.String())
			}
		}
	}
	return tables
}

func main() {
	queries := []string{
		"SELECT * FROM users WHERE id = 1",
		"SELECT id, name FROM users WHERE id = 1",
		"INSERT INTO users (id, name) VALUES (1, 'Alice')",
		"UPDATE users SET name = 'Bob' WHERE id = 1",
		"DELETE FROM users WHERE id = 1",
		"CREATE TABLE users (id INT, name VARCHAR(100))",
		"ALTER TABLE users ADD COLUMN age INT",
		"DROP TABLE users",
		"CREATE DATABASE test_db",
		"DROP DATABASE test_db",
		"TRUNCATE TABLE users",
	}

	for _, query := range queries {
		fmt.Println("Query: ", query)
		statement, tables, err := Parse(query)
		if err != nil {
			fmt.Printf("Error parsing query: %s\nQuery: %s\n\n", err, query)
			continue
		}
		fmt.Printf("Statement: %s, Tables/DBs: %v\n", statement, tables)
	}
}
```

```terminal
$ go run .
Query:  SELECT * FROM users WHERE id = 1
Statement: SELECT, Tables/DBs: [users]
Query:  SELECT id, name FROM users WHERE id = 1
Statement: SELECT, Tables/DBs: [users]
Query:  INSERT INTO users (id, name) VALUES (1, 'Alice')
Statement: INSERT, Tables/DBs: [users]
Query:  UPDATE users SET name = 'Bob' WHERE id = 1
Statement: UPDATE, Tables/DBs: [users]
Query:  DELETE FROM users WHERE id = 1
Statement: DELETE, Tables/DBs: [users]
Query:  CREATE TABLE users (id INT, name VARCHAR(100))
Statement: CREATE TABLE, Tables/DBs: [users]
Query:  ALTER TABLE users ADD COLUMN age INT
Statement: ALTER TABLE, Tables/DBs: [users]
Query:  DROP TABLE users
Statement: DROP TABLE, Tables/DBs: [users]
Query:  CREATE DATABASE test_db
Statement: CREATE DATABASE, Tables/DBs: [test_db]
Query:  DROP DATABASE test_db
Statement: DROP DATABASE, Tables/DBs: [test_db]
Query:  TRUNCATE TABLE users
Statement: TRUNCATE TABLE, Tables/DBs: [users]
```

damemi: Updated to use vitess with your suggestion, looks good to me, ptal!

go.mod
damemi: The binary size went from 30MB->37MB with vitess included, that's not small but I'm not personally concerned about it and it could be something we look to refactor later. wdyt?

internal/pkg/instrumentation/bpf/database/sql/probe.go
damemi: done

internal/test/e2e/databasesql/traces.json
```diff
@@ -63,6 +63,226 @@
...
+              "spanId": "xxxxx",
+              "status": {},
+              "traceId": "xxxxx"
+            },
+            {
+              "attributes": [
+                {
+                  "key": "db.query.text",
+                  "value": {
+                    "stringValue": "syntax error"
```
damemi, line 267: will follow up to parse error response codes from databases separately, and mark them as error spans appropriately. Don't want to grow this PR too much and this is equivalent to what we do now

go.mod
```diff
@@ -1,6 +1,6 @@
 module go.opentelemetry.io/auto
 
-go 1.22.7
+go 1.23.2
```
MrAlias, line 3: Hmm, this might be problematic. This would violate our compatibility policy: https://github.com/open-telemetry/opentelemetry-go-instrumentation?tab=readme-ov-file#compatibility

internal/pkg/instrumentation/bpf/database/sql/probe_test.go
```diff
@@ -18,7 +19,60 @@ import (
...
+					Query: byteQuery,
+				})
+			}
+		})
+	}
+}
+
 func TestProbeConvertEvent(t *testing.T) {
+	err := os.Setenv(IncludeDBOperationEnvVar, "true")
+	assert.NoError(t, err)
```
MrAlias, line 75:
```suggestion
	t.Setenv(IncludeDBOperationEnvVar, "true")
```

internal/test/e2e/databasesql/main.go
```diff
@@ -72,20 +71,19 @@ func NewServer() *Server {
...
 
-	rows, err := conn.QueryContext(req.Context(), sqlQuery)
+	rows, err := conn.QueryContext(req.Context(), query)
 	if err != nil {
-		panic(err)
+		logger.Error(err.Error())
 	}
 
-	logger.Info("queryDb called")
+	logger.Info(fmt.Sprintf("Query: %s", query))
```
MrAlias, line 86: nit: use the structured elements of the logger to handle this.

```suggestion
	logger.Info("queryDB called", "query", query))
```

MrAlias (commented):
This looks like an exciting change. Mainly small comments outside of the compatibility policy issue.

Do we need to support older versions of Go? Would that break any downstream users of `go.opentelemetry.io/auto`?

LICENSES/libbpf/bpf.h
RonFed, line 1: What's the reason for these changes? can you pleas add it to the PR description

docs/configuration.md
```diff
@@ -27,6 +27,7 @@ Alternatively, you can add support for additional or different configurations by
 | Environment variable                | Description                                            | Default value |
 |-------------------------------------|--------------------------------------------------------|---------------|
 | `OTEL_GO_AUTO_INCLUDE_DB_STATEMENT` | Sets whether to include SQL queries in the trace data. |               |
+| `OTEL_GO_AUTO_INCLUDE_DB_OPERATION` | Sets whether to include SQL operation in the trace data. |               |
```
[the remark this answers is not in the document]
RonFed: It is not valid to have `OTEL_GO_AUTO_INCLUDE_DB_OPERATION` without `OTEL_GO_AUTO_INCLUDE_DB_STATEMENT`, right? If so we should probably point that out.
Also, should we state the exact attributes we'll add? i.e `db.collection.name`  `db.operation.name` etc`

go.mod
RonFed: I agree. this looks quite aggressive. 
For reference, when I wanted to update the dependency of this repo I had to bump the go.mod to 1.22.7 which is what we require today (https://github.com/odigos-io/odigos/pull/1776/commits/fb167f82c1f76947ee2f1d2966c5bc363a279fd1), and we have quite a lot of other dependencies and none of them required this relatively new version.

LICENSES/libbpf/bpf.h
damemi: this all came in when I ran `make update-licenses` (CI license check failed saying it needed updates). I agree it's weird that libbpf was also updated

MrAlias: It looks like the libbpf script wasn't run.

MrAlias:
I agree with the revert to the other sql parsing package. I think we should still be able to support table parsing though based on https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/1253#discussion_r1854195790

docs/configuration.md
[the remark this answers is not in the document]
damemi: Updated

go.mod
damemi: Reverting back to the other library, since it doesn't seem like vitess will work for us right now :(

damemi:
@MrAlias @RonFed this is ready for another round of review, I reverted back to the other library and we'll just support basic statements + collection when available

internal/pkg/instrumentation/bpf/database/sql/probe.go
```diff
@@ -97,6 +104,23 @@ func processFn(e *event) ptrace.SpanSlice {
 		span.Attributes().PutStr(string(semconv.DBQueryTextKey), query)
 	}
 
+	includeOperationVal := os.Getenv(IncludeDBOperationEnvVar)
+	if includeOperationVal != "" {
+		include, err := strconv.ParseBool(includeOperationVal)
```
RonFed, line 109: We could read the env var once and cache the result ( not sure how expensive it is).

MrAlias: `Getenv` ultimately calls `syscall.Getenv`: https://cs.opensource.google/go/go/+/refs/tags/go1.23.4:src/syscall/env_unix.go;l=69-89

This is pretty well cached and handles updates concurrently and efficiently. We should be fine making this call each time.

internal/pkg/instrumentation/bpf/database/sql/probe_test.go
```diff
@@ -18,7 +18,59 @@ import (
...
+	endOffset := utils.TimeToBootOffset(end)
+
+	traceID := trace.TraceID{1}
+	spanID := trace.SpanID{1}
+
+	for _, t := range tests {
+		b.Run(t.name, func(b *testing.B) {
+			var byteQuery [256]byte
+			copy(byteQuery[:], []byte(t.query))
+			b.ResetTimer()
```
MrAlias, line 57: So we can see memory benchmark.

```suggestion
			b.ReportAllocs()
			b.ResetTimer()
```

MrAlias (approved):
:medal_sports:

PR #1292, opened 2024-11-13
Handle nil sampling config

RonFed:
An implementation of the `ConfigProvider` interface might provide configurations that don't specify a sampler.

* For the initial configuration, the ` NewSamplingManager` function is used - where a default config will be used if non is present.
* For an update in configuration the `applyConfig` function will be used (this is not fully implemented yet in the main manager). In that case, the `ConfigLoop` handler will avoid calling the sampling `applyConfig` if no config was provided - this will make that update a no-op for sampling. For this reason, the `applyConfig` function is changed to return an error if a nil config is provided.


CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ OpenTelemetry Go Automatic Instrumentation adheres to [Semantic Versioning](http
 ### Fixed
 
 - Don't include `db.query.text` attribute in `database/sql` if the query string is empty or not collected. ([#1246](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/1246))
+- Handle a `ConfigProvider` which doesn't provide a sampling config in the initial configuration be applying the default sampler. ([#1292](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/1292))
```
MrAlias, line 21:
```suggestion
- Handle a `ConfigProvider` which doesn't provide a sampling config in the initial configuration by applying the default sampler. ([#1292](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/1292))
```

MrAlias (approved):

Issue #1297, opened 2024-11-18
SDK needs to be concurrent safe


MrAlias:
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#concurrency:

> ## Concurrency
> 
> For languages which support concurrent execution the Tracing APIs provide
> specific guarantees and safeties. Not all of API functions are safe to
> be called concurrently.
> 
> **TracerProvider** - all methods are safe to be called concurrently.
> 
> **Tracer** - all methods are safe to be called concurrently.
> 
> **Span** - All methods of Span are safe to be called concurrently.
> 
> **Event** - Events are immutable and safe to be used concurrently.
> 
> **Link** - Links are immutable and safe to be used concurrently.

PR #1298, opened 2024-11-18
Test SDK TracerProvider

MrAlias:
- Move the tracerProvider type and related functionality to its own file to help discoverability.
- Use a singleton instance of the tracerProvider for tracerProviderInstance instead of having a new value created each call to TracerProvider.
- Test the singleton nature of TracerProvider.
- Test the concurrent safe behavior of the provided TracerProvider (part of #1297).

### DIFF
--- a/sdk/trace.go
+++ b/sdk/trace.go
@@ -20,30 +20,6 @@ import (
 	"go.opentelemetry.io/auto/sdk/internal/telemetry"
 )
 
-// TracerProvider returns an auto-instrumentable [trace.TracerProvider].
-//
-// If an [go.opentelemetry.io/auto.Instrumentation] is configured to instrument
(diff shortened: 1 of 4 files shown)


RonFed:
Which part here addresses concurrency? If I understood this correctly, the old code also used a singleton?

MrAlias:
> Which part here addresses concurrency? If I understood this correctly, the old code also used a singleton?

Correct. The old code is already concurrent safe. This just adds a test to ensure that it remains that way.

The old code uses a default value as the file scope tracer provider. Since Go will always copy-by-value, each value returned from `TracerProvider` was being copied to new memory. This also changes that to always reference the same instance of a `*tracerProvider` that is allocated to the heap.

RonFed (approved):

PR #1300, opened 2024-11-18
Make SDK span concurrent safe

MrAlias:
Part of #1297 

Update the span implementation to ensure concurrent safe methods. Include a test to verify concurrent safety.

### DIFF
--- a/sdk/trace.go
+++ b/sdk/trace.go
@@ -9,6 +9,8 @@ import (
 	"fmt"
 	"reflect"
 	"runtime"
+	"sync"
+	"sync/atomic"
 	"time"
 
 	"go.opentelemetry.io/otel/attribute"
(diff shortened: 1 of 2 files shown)


sdk/trace.go
```diff
@@ -54,14 +56,17 @@ var _ trace.Tracer = tracer{}
...
 	// Ask eBPF for sampling decision and span context info.
-	t.start(ctx, span, &psc, &span.sampled, &span.spanContext)
+	t.start(ctx, span, &psc, &sampled, &span.spanContext)
+
+	span.sampled.Store(sampled)
 
 	ctx = trace.ContextWithSpan(ctx, span)
 
-	if span.sampled {
+	if sampled {
```
damemi, line 69: will `sampled` always be true here? I didn't see a test for `span.traces` and `span.span` that are set in this block but maybe I missed it

MrAlias: If the eBPF program sets it to false based on its sampler `sampled` will be `false`.

damemi: ahh right, this is the magic variable

damemi (approved):

RonFed (approved):

PR #1301, opened 2024-11-19
[PoC] Span limits

MrAlias:
Add functionality to `auto/sdk` so it can interpret configuration from OTel defined environment variables for tracing limits.


damemi:
Closing , poc not needed anymore

Issue #1314, opened 2024-11-25
Support span limits in `auto/sdk`


MrAlias:
- [x] #1315 
- [ ] #1325 
- [x] #1320 
- [x] #1324

PR #1315, opened 2024-11-25
Add span attribute limit to `auto/sdk`

MrAlias:
Part of #1314

### DIFF
--- a/sdk/limit.go
+++ b/sdk/limit.go
@@ -1,0 +1,87 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package sdk
+
+import (
+	"os"
+	"strconv"
+)
+
+// maxSpan are the span limits resolved during startup.
+var maxSpan = newSpanLimits()
+
+type spanLimits struct {
+	// Attrs is the number of allowed attributes for a span.
+	//
+	// This is resolved from the environment variable value for the
+	// OTEL_SPAN_ATTRIBUTE_COUNT_LIMIT key if it exists. Otherwise, the
+	// environment variable value for OTEL_ATTRIBUTE_COUNT_LIMIT, or 128 if
+	// that is not set, is used.
+	Attrs int
+	// AttrValueLen is the maximum attribute value length allowed for a span.
+	//
+	// This is resolved from the environment variable value for the
+	// OTEL_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT key if it exists. Otherwise, the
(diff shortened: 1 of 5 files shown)


sdk/limit.go
```diff
@@ -0,0 +1,87 @@
...
+			continue
+		}
+
+		v, err := strconv.Atoi(strV)
+		if err == nil {
+			return v
+		}
+	}
+
+	return defaultVal
```
damemi, line 86: is it ok for the sdk to silently use the default value in case of an error?

sdk/limit_test.go
```diff
@@ -0,0 +1,81 @@
...
+
+package sdk
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestSpanAttrValLenLimit(t *testing.T) {
```
damemi, line 12: could these be table tests instead of individual tests? Also, a test for invalid env values falling back to the default would be good

damemi (approved):
small nits, but lgtm

sdk/limit.go
MrAlias: [From the specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md#numeric-value):

> For variables accepting a numeric value, if the user provides a value the implementation cannot parse, or which is outside the valid range for the configuration item, the implementation SHOULD generate a warning and gracefully ignore the setting, i.e., treat them as not set.

This is a non-standard SDK, so we are not trying to implement the specification here per se, but I do think we want to drop the invalid value and use the default as the specification prescribes. I think it will be the best we can do in this error situation. It also follows what the default SDK does which will match initiated OTel users expectations.

As for the warning, I'm hesitant to generically log something here. We do not have any user configured logger, meaning we will only be able to generically log something using the global logger. That and this will be in the default global API. If a user doesn't even use the OTel API because the auto-instrumentation is not attached this would still log a warning for invalid env var.

That said, I can change this to log using the global `slog` logger if you think it makes sense.

damemi: sgtm, it would be nice to log the error but I agree that's a little messy without a user-defined logger (phantom log spam you can't control). So I'm fine without it, I'll leave it up to you if you think the benefit outweighs the control

MrAlias: I'll update to add logging with the default `slog` logger. It will only log in an error scenario where a user sets in invalid value and they can override the behavior (in theory) by replacing the default logger with something that drops the message.

MrAlias: Updated. PTAL

RonFed (approved):
🥇

PR #1319, opened 2024-11-27
fix: parse of tracers map for otel-go 1.32

RonFed:
Following #1302, this PR fixes a bug of parsing the tracers map for version 1.32 of otel-go.
Added unused tracers for the e2e test to test the parsing of non-trivial tracers map.


internal/test/e2e/otelglobal/main.go
```diff
@@ -19,6 +20,12 @@ var tracer = otel.Tracer(
 	trace.WithSchemaURL("https://some_schema"),
 )
 
+func setUnUsedTracers() {
```
MrAlias, line 23:
```suggestion
func setUnusedTracers() {
```

Unused is a single word.

MrAlias (approved):

internal/test/e2e/otelglobal/main.go
RonFed: Fixed

PR #780, opened 2024-04-12
Remove gin specific probes

grcevski:
Since we switched to putting the server HTTP probes on `net/http.serverHandler.ServeHTTP` the Gin specific probe is not needed. The handler for Gin is called from the default HTTP server handler and effectively we get duplicate probes. The e2e tests are purposefully left in place so that we can still test that Gin instrumentation is working. I modified the gin tests to do the same level of validation as the net/http tests and I removed the double server span.

This is the backtrace from Delve attached to our Gin example:

```sh
   569:	// ServeHTTP conforms to the http.Handler interface.
=> 570:	func (engine *Engine) ServeHTTP(w http.ResponseWriter, req *http.Request) {
   571:		c := engine.pool.Get().(*Context)
   572:		c.writermem.reset(w)
   573:		c.Request = req
   574:		c.reset()
   575:	
(dlv) bt
0  0x0000000000770a4e in github.com/gin-gonic/gin.(*Engine).ServeHTTP
   at /home/nino/go/pkg/mod/github.com/gin-gonic/gin@v1.9.1/gin.go:570
1  0x000000000065242e in net/http.serverHandler.ServeHTTP <--- /** the default HTTP server handler **/
   at /usr/local/go/src/net/http/server.go:3137
2  0x000000000064ea08 in net/http.(*conn).serve
   at /usr/local/go/src/net/http/server.go:2039
3  0x0000000000652c48 in net/http.(*Server).Serve.gowrap3
   at /usr/local/go/src/net/http/server.go:3285
4  0x000000000046ece1 in runtime.goexit
   at /usr/local/go/src/runtime/asm_amd64.s:1695
```

Closes #171 as not needed

PR #1359, opened 2024-12-02
feat: Support `http.route` attribute for `gin` and `chi`

iBluemind:
The name of the HTTP span must be low-cardinality.
To achieve this, the `http.route` attribute can be included in the span's name.

[Semantic conventions for HTTP spans
](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-spans.md#semantic-conventions-for-http-spans)
> HTTP span names SHOULD be {method} {target} if there is a (low-cardinality) target available.
>
> The {target} SHOULD be one of the following:
> 
>  - [http.route](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/attributes-registry/http.md) for HTTP Server spans

However, most Go-based web frameworks use different methods for parsing routing paths.
It's not feasible to handle all cases with just an eBPF probe for `net/http`.

Therefore I submit a patch that adds the eBPF probe support for [gin](https://github.com/gin-gonic/gin) and [chi](https://github.com/go-chi/chi) frameworks.


MrAlias (changes requested):
I am not in favor of accepting these changes:

- This change is > 2k lines
- It does not address the reasons we removed this instrumentation in the first place: https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/780
- It is adding excessive probe code to the codebase prior to resolution of https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/1105

iBluemind:
@MrAlias Thank you for your prompt feedback. I'm aware of the #780 in question. 
However, most frameworks, including Gin, do not use the ServeMux pattern. So I think there are limitations to using only net/http probes.

iBluemind:
@MrAlias The e2e test is written like [this line](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/1359/files#diff-e6cfa0b82feb81465ee5bd1e366f81bf404872f475c5bf9ed17346cfdd3293dfR21). It passed by [this line](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/1359/files#diff-54cff84b62cc0fa8a411f4c7e4bdbbc0b05312046fa98811bd529b8711a9b36bR24). I ran the e2e test by [this workflow](https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/main/.github/workflows/kind.yml).

Below is the backtrace from Delve attached to our nethttp example.
```
> net/http.(*conn).serve() /opt/homebrew/Cellar/go/1.23.1/libexec/src/net/http/server.go:2093 (PC: 0x10077c16c)
Warning: debugging optimized function
Values returned:

  2088:			// in parallel even if their responses need to be serialized.
  2089:			// But we're not going to implement HTTP pipelining because it
  2090:			// was never deployed in the wild and the answer is HTTP/2.
  2091:			inFlightResponse = w
  2092:			serverHandler{c.server}.ServeHTTP(w, w.req)
=>2093:			inFlightResponse = nil
  2094:			w.cancelCtx()
  2095:			if c.hijacked() {
  2096:				return
  2097:			}
  2098:			w.finishRequest()
(dlv) p req
("*net/http.Request")(0x140000b4000)
*net/http.Request {
	Method: "GET",
	URL: *net/url.URL {
		Scheme: "",
		Opaque: "",
		User: *net/url.Userinfo nil,
		Host: "",
		Path: "/hello/42",
		RawPath: "",
		OmitHost: false,
		ForceQuery: false,
		RawQuery: "query=true",
		Fragment: "",
		RawFragment: "",},
	Proto: "HTTP/1.1",
	ProtoMajor: 1,
	ProtoMinor: 1,
	Header: net/http.Header [
		"User-Agent": [
			"Go-http-client/1.1",
		],
		"Authorization": [
			"Basic dXNlcjo=",
		],
		"Accept-Encoding": ["gzip"],
	],
	Body: io.ReadCloser(net/http.noBody) {},
	GetBody: nil,
	ContentLength: 0,
	TransferEncoding: []string len: 0, cap: 0, nil,
	Close: false,
	Host: "localhost:8080",
	Form: net/url.Values nil,
	PostForm: net/url.Values nil,
	MultipartForm: *mime/multipart.Form nil,
	Trailer: net/http.Header nil,
	RemoteAddr: "[::1]:59968",
	RequestURI: "/hello/42?query=true",
	TLS: *crypto/tls.ConnectionState nil,
	Cancel: <-chan struct {} {},
	Response: *net/http.Response nil,
	Pattern: "/hello/{id}",
	ctx: context.Context(*context.cancelCtx) *{
		Context: context.Context(*context.cancelCtx) ...,
		mu: (*sync.Mutex)(0x1400019c2e0),
		done: (*"sync/atomic.Value")(0x1400019c2e8),
		children: map[context.canceler]struct {} nil,
		err: error nil,
		cause: error nil,},
	pat: *net/http.pattern {                  <--------------------------- /** `pat` is filled **/
		str: "/hello/{id}",
		method: "",
		host: "",
		segments: []net/http.segment len: 2, cap: 2, [
			(*"net/http.segment")(0x1400007ec60),
			(*"net/http.segment")(0x1400007ec78),
		],
		loc: "/Users/manjong/Documents/sources/opentelemetry-go-instrumentatio...+38 more",},
	matches: []string len: 1, cap: 1, ["42"],
	otherValues: map[string]string nil,}
```

And this is the one for our Gin example.
```
> net/http.(*conn).serve() /opt/homebrew/Cellar/go/1.23.1/libexec/src/net/http/server.go:2093 (PC: 0x10477abac)
Warning: debugging optimized function
Values returned:

  2088:			// in parallel even if their responses need to be serialized.
  2089:			// But we're not going to implement HTTP pipelining because it
  2090:			// was never deployed in the wild and the answer is HTTP/2.
  2091:			inFlightResponse = w
  2092:			serverHandler{c.server}.ServeHTTP(w, w.req)
=>2093:			inFlightResponse = nil
  2094:			w.cancelCtx()
  2095:			if c.hijacked() {
  2096:				return
  2097:			}
  2098:			w.finishRequest()
(dlv) p req
("*net/http.Request")(0x14000489180)
*net/http.Request {
	Method: "GET",
	URL: *net/url.URL {
		Scheme: "",
		Opaque: "",
		User: *net/url.Userinfo nil,
		Host: "",
		Path: "/hello-gin/1",
		RawPath: "",
		OmitHost: false,
		ForceQuery: false,
		RawQuery: "",
		Fragment: "",
		RawFragment: "",},
	Proto: "HTTP/1.1",
	ProtoMajor: 1,
	ProtoMinor: 1,
	Header: net/http.Header [
		"User-Agent": [
			"Go-http-client/1.1",
		],
		"Accept-Encoding": ["gzip"],
	],
	Body: io.ReadCloser(net/http.noBody) {},
	GetBody: nil,
	ContentLength: 0,
	TransferEncoding: []string len: 0, cap: 0, nil,
	Close: false,
	Host: "localhost:8080",
	Form: net/url.Values nil,
	PostForm: net/url.Values nil,
	MultipartForm: *mime/multipart.Form nil,
	Trailer: net/http.Header nil,
	RemoteAddr: "[::1]:60307",
	RequestURI: "/hello-gin/1",
	TLS: *crypto/tls.ConnectionState nil,
	Cancel: <-chan struct {} {},
	Response: *net/http.Response nil,
	Pattern: "",
	ctx: context.Context(*context.cancelCtx) *{
		Context: context.Context(*context.cancelCtx) ...,
		mu: (*sync.Mutex)(0x1400019a010),
		done: (*"sync/atomic.Value")(0x1400019a018),
		children: map[context.canceler]struct {} nil,
		err: error nil,
		cause: error nil,},
	pat: *net/http.pattern nil,            <--------------------------- /** `pat` is nil **/
	matches: []string len: 0, cap: 0, nil,
	otherValues: map[string]string nil,}
```

RonFed:
@iBluemind Thank you for the contribution!

I didn't go in-depth to the code yet, just addressing @MrAlias points:

* I think the original reason we removed the gin probes is that we didn't collect any gin-specific data and those spans didn't add any extra data. This PR adds the `http.route` attribute for those frameworks which is something super useful and I saw a few users who would greatly benefit from these changes. So in general, if we can add attributes specific to gin or chi (which goes together with the semconv) - I think this has great value.
* large PR - I agree - but a lot of it is offsets and test JSON data.
* #1105 is a good point, however, as I said I do see a lot of value in adding more attributes to gin users.

MrAlias:
I understand the name is different. But I'm not seeing that as justification in adding redundant telemetry as identified (and fixed by) #780.

MrAlias:
> I agree that having 3 probes(nethttp, gin, chi) seems like a bit of redundancy. Is there a better way to do this?

Can we update the `nethttp` instrumentation to detect what name parsing should be performed and signal the eBPF code with that information? I.e. detect if gin, chi, or net/http routing patterns should be used and send an enum as a flag.

iBluemind:
> Can we update the `nethttp` instrumentation to detect what name parsing should be performed and signal the eBPF code with that information? I.e. detect if gin, chi, or net/http routing patterns should be used and send an enum as a flag.

That would be better, thank you

PR #1405, opened 2024-12-10
Enable auto/sdk in otel global

### LINKED ISSUE #974: Add instrumentation for the "go.opentelemetry.io/otel/trace".nonRecordingSpan
A common instrumentation pattern for tracing is to return the the `TracerProvider` from the span held in a context:

```go
import (
	"net/http"

	"go.opentelemetry.io/otel/trace"
)

func handler(w http.ResponseWriter, r *http.Request) {
	ctx := r.Context()
	tracer := trace.SpanFromContext(ctx).TracerProvider().Tracer("tracer")
	ctx, span := tracer.Start(ctx, "span")
	defer span.End()
	/* ... */
}
```

When using the instrumentation of the global trace API, this span will not be recorded.

The span created in this above code should be captured by auto-instrumentation. Meaning the [`nonRecordingSpan`](https://github.com/open-telemetry/opentelemetry-go/blob/1b5834f96c6eaaac5ffc21250edc296dea2525df/trace/nonrecording.go#L9) should be instrumented.

### LINKED ISSUE #1399: Write the flag to enable the autosdk for the otel global
Our instrumentation for the OTel global API needs to set this bool when loaded:

https://github.com/open-telemetry/opentelemetry-go/blob/dd83cac3f8f4412bc0c40e739eb33780bd9a3aba/internal/global/trace.go#L152-L159

### LINKED ISSUE #1400: Add and e2e test for otel global
Similar to the `autosdk` e2e test except it calls `otel.GetTracerProvider` instead of `auto.TracerProvider` from `go.opentelemetry.io/otel`.

MrAlias:
Resolve #1399 
Resolve #1400 
Resolve #974 

This includes a `probe.Uprobe` refactor. Instead of passing back `links` when the uprobe loads, the uprobe itself now hold these values. When its `Close` method is called it will, itself, call each of these close methods.

This redesign is used so that the `newSpan` uprobe can be closed prior to the others (which are closed by the `Manager` ).


MrAlias:
A release of OTel Go is required to not depend on a commit hash version.

MrAlias:
Waiting on https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/1418

internal/pkg/instrumentation/bpf/go.opentelemetry.io/otel/traceglobal/bpf/probe.bpf.c
```diff
@@ -388,6 +391,33 @@ static __always_inline long fill_tracer_id(tracer_id_t *tracer_id, go_tracer_ptr
     return 0;
 }
 
+// This instrumentation attaches uprobe to the following function:
+// func (t *tracer) newSpan(ctx context.Context, autoSpan *bool, name string, opts []trace.SpanStartOption) (context.Context, trace.Span) {
+// https://github.com/open-telemetry/opentelemetry-go/blob/ac386f383cdfc14f546b4e55e8726a0a45e8a409/internal/global/trace.go#L161
+SEC("uprobe/newSpan")
+int uprobe_newStart(struct pt_regs *ctx) {
```
MrAlias, line 398: Ideally this would be unloaded once this sets the flag for a process.

It was suggested by @RonFed that we could send a special event to the output perf buffer and in the probe read the event as a special case and unload the probe. Investigation is needed.

MrAlias: I've updated locally to accomplish this but having issues. The `parent` span in the e2e test fails to get tracing IDs. The debug tracing logs contain:

```
             app-2959921 [006] ...11 527565.795722: bpf_trace_printk: write_span_context: failed to write trace ID: -14
             app-2959921 [006] ...11 527565.795730: bpf_trace_printk: failed to write span context: -2
```

Looking at the kernel errors `-14` indicates this is likely `EFAULT`: https://elixir.bootlin.com/linux/v6.12.5/source/mm/maccess.c#L143

I'm not 100% sure that is where the error is coming from, and I'm very much less sure why the error is being trown.

MrAlias: I'm considering just skipping this unload step and tracking in an issue for follow-up.

RonFed: @MrAlias could you push a commit with the attempt to unload the probe and I can have a look at it?

MrAlias: @RonFed sure thing: https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/1405/commits/9a4811713b8e3385dd0fa0633c2178aa6622b716

MrAlias: Hmm, things seem to be "working" now. I think the [upgrade of github.com/cilium/ebpf to v0.17.0](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/1500) may have changed things.

Still need to look into the span kind failure.

MrAlias: It does looks to be working locally as well. I'm not exactly sure what changed to make this work, but happy to see it is working as expected.

MrAlias: Resolving. This looks resolved and ready for review.

internal/pkg/instrumentation/bpf/go.opentelemetry.io/otel/traceglobal/probe.go
```diff
@@ -41,6 +44,17 @@ func New(logger *slog.Logger) probe.Probe {
 		SpanKind:        trace.SpanKindClient,
 		InstrumentedPkg: pkg,
 	}
+
+	uprobeNewStart := &probe.Uprobe{
+		Sym:         "go.opentelemetry.io/otel/internal/global.(*tracer).newSpan",
+		EntryProbe:  "uprobe_newStart",
+		FailureMode: probe.FailureModeWarn,
```
RonFed, line 51: Should we have `PackageConstrainsts` here for >= 1.33?

internal/pkg/instrumentation/probe/probe.go
```diff
@@ -365,50 +371,76 @@ type Uprobe struct {
 	// function specified by Sym. If ReturnProbe is empty, no eBPF program will be attached to the return of the function.
 	ReturnProbe string
 	DependsOn   []string
+
+	closers atomic.Pointer[[]io.Closer]
```
RonFed, line 375: Why not `[]io.Closer`?
What are the concurrent scenarios we are trying to protect against?

internal/pkg/instrumentation/bpf/go.opentelemetry.io/otel/traceglobal/probe.go
```diff
@@ -133,12 +148,55 @@ func New(logger *slog.Logger) probe.Probe {
...
+	var kind recordKind
+	err := binary.Read(reader, binary.LittleEndian, &kind)
+	if err != nil {
+		return nil, err
+	}
+
+	var e *event
+	switch kind {
+	case recordKindTelemetry:
+		e := new(event)
```
RonFed, line 184: This might be an ignorant question. Why is this required if we have `var e *event` declared?

RonFed (commented):
It's very cool to see it all come together,
Left some comments but it looks great.

internal/pkg/instrumentation/probe/probe.go
MrAlias: The `Uprobe.Close` is called by the `Manager` which can shutdown asynchronously. For example, this can occur when the context passed to the new `Instrumentation` is canceled asynchronously[^1]. This is to prevent that data-race.

[^1]: https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/aeb95422568a2e2901a02620527e526161936321/cli/main.go#L87-L90

internal/pkg/instrumentation/bpf/go.opentelemetry.io/otel/traceglobal/probe.go
MrAlias: The event pointer is declared above, but the memory has not been allocated. Without the allocation there will be a panic: https://go.dev/play/p/AADS3FvylTR

That said, this is shadowing `e` instead of assigning to it.

MrAlias, line 184:
```suggestion
		e = new(event)
```

internal/pkg/instrumentation/bpf/go.opentelemetry.io/otel/traceglobal/probe.go
MrAlias: Updated. PTAL.

internal/pkg/instrumentation/bpf/go.opentelemetry.io/auto/sdk/probe.go
RonFed, line 1: Should we also add package constraints on the auto sdk probes?

internal/pkg/instrumentation/bpf/go.opentelemetry.io/otel/traceglobal/probe.go
```diff
@@ -107,38 +144,97 @@ func New(logger *slog.Logger) probe.Probe {
...
+	}
+
+	var e *event
+	switch kind {
+	case recordKindTelemetry:
+		e = new(event)
+		reader.Reset(record.RawSample)
+		err = binary.Read(reader, binary.LittleEndian, e)
+	case recordKindConrol:
+		c.uprobeNewStartMu.Lock()
```
RonFed, line 226: The `decodeEvent`  function of a probe instance can't be called from multiple goroutines , right? If so, maybe should remove the mutex

internal/pkg/instrumentation/probe/probe.go
RonFed: I see. Do you think we can use `sync.Once` instead? I think it might be more readable

MrAlias: Using `sync.Once` would mean we need two fields here. One for the `sync.Once` and the other to hold the closer slice. I think having a single field is cleaner. It does not require a user to understand the implicit coupling between two fields and avoids the error case when they are not correctly handled.

Alternatively, we could use a [`sync.OnceFunc`](https://pkg.go.dev/sync#OnceFunc) as a single field. However, this would leave a data-race when trying to replace the "swap" functionality. We would want to check if the "close" field is non-nil before assigning to it in `load`. This would need a semaphore to ensure no data-race is needed, and again mean we require two fields.

It may also be tempting to not implement the "swap" functionality. That way we would just assign blindly on load and hope for the best on close (meaning there would be a data-race there). However, data-race asside, that would mean we no longer support instrumentation shutdown and then restarting, or probe unloading and reloading.

We could also use `atomic.Value` here. That was explored, but decided against due to the lack of type safety and the inability to set a value of `nil` (representing a state of stopped) without a panic.

internal/pkg/instrumentation/bpf/go.opentelemetry.io/auto/sdk/probe.go
MrAlias: I think we want to always add uprobes for `auto/sdk`. If that package is loaded, it may be because the user has done so explicitly and we don't want to not instrument it because of the version of `otel`. It also doesn't add runtime overhead if it is not used given the uprobe break-point is never triggered.

RonFed (approved):

Issue #1209, opened 2024-10-22
How to get the value of go arguments `interface{}` in bpf?


minimAluminiumalism:
I'm trying to a new instrument probe for a package, the target function attached the probe is:

```
func (c cmdable) Set(ctx context.Context, key string, value interface{}, expiration time.Duration) *StatusCmd
```
I successfully obtained the arg `key` with following code:

```
u64 key_ptr_pos = 4;
u64 key_len_ptr_pos = 5;
void *set_key_ptr = get_argument(ctx, key_ptr_pos);
u64 key_str_len = (u64)get_argument(ctx, key_len_ptr_pos);
u64 query_size = MAX_QUERY_SIZE < key_str_len ? MAX_QUERY_SIZE : key_str_len;
bpf_probe_read(key_str, query_size, set_key_ptr);
```
But I I encountered difficulties in obtaining another arg `value` which is `interface{}` type. I searched through all the code in your project but not found any related functions like the above function. Could you provide some guidance?

damemi:
Does https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/1176 help?

minimAluminiumalism:
@damemi The difficulty is how can I get the type of the `interface` param?

minimAluminiumalism:
@MrAlias Can you give some guidance about obtaining the `interface{}` type args? 
[Here is the function](https://github.com/redis/go-redis/blob/1ed936eb097d10f3aaf40a44d282778f879eac32/string_commands.go#L166) I want to add a probe to.
I can get the forth arg `value` when I ensure the data type is string, but the actual type is interface{}.
So I need the get the arg type first and then get the data? I have tried but failed to obtain the type.

Or do I need to first obtain the address of the data in kernel mode, then send it(`bpf_perf_event_output`) to user mode for parsing?

MrAlias:
@minimAluminiumalism I am not sure how to achieve this.

You will need to parse the `itab._type` and dynamically reconstruct the type pointed to by the interface's `data` field. Essentially, you will need to recreate the Go runtime implementation in C.

This is something that has blocked prior tasks and caused us to search alternate solutions.

For what its worth, https://cmc.gitbook.io/go-internals/chapter-ii-interfaces provides a good overview of how the runtime package produces assembly code that will do the thing you need to do. Hopefully it is a decent jumping off point. Good luck.

minimAluminiumalism:
[...]

Issue #1053, opened 2024-09-06
Redis support


minimAluminiumalism:
I check the library support list and there is still no support for redis. I also tested and ensured that. Do you have a schedule?
In addition, how can i instrument the redis lib with this project? Can I use original otel sdk like `go.opentelemetry.io/otel/trace`.

MrAlias:
We have updated our contributing guidelines and added issues templates for new instrumentation requests. We are now tracking this in https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/1775

PR #1525, opened 2024-12-27
Add new package instrument: go-redis

minimAluminiumalism:
Ref https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/1053

This project currently does not support Redis, and this PR aims to support it.

Instrumentation for the library `github.com/go-redis/redis/v8` is quite difficult, primarily because some of the available instrument functions include input params of type `interface{}` (this issue is also mentioned in https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/1209) like this:

```
func (c *baseClient) process(ctx context.Context, cmd Cmder)
```
## A possible solution (found to be unfeasible after attempting)

I want to get the **type** of these `interface{}` type input params, then try to retrieving the corresponding value based on that type. The `type` field's position of `eface` is:

```
type eface struct {
    _type *_type
    data  unsafe.Pointer
}
type _type struct {
    size       uintptr
    ptrdata    uintptr
    hash       uint32
    tflag      uint8
    align      uint8
    fieldalign uint8
    kind       uint8
    equal      func(unsafe.Pointer, unsafe.Pointer) bool
    gcdata     *byte
    str        nameOff
    ptrToThis  typeOff
}
```

The biggest issue with this solution is its high level of complexity, which will exceed the instruction limit of eBPF(100W) with error `program uprobe_process: load program: argument list too long: BPF program is too large. Processed 1000001 insn (4760 line(s) omitted)`

The implementation is the closed PR https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/1522

## Final selected solution

So I suppose we cannot directly start from the `interface{}` input parameters of instrument function. I decided to start with the **RESP** protocol. Redis cmds will be encoded to bytes stream(RESP), so I chose the instrument func:

```
func (cn *Conn) WithWriter(ctx context.Context, timeout time.Duration, fn func(wr *proto.Writer)
```

But the `bufio` uses write-back buffer strategy. It makes me cannot determine the boundaries of a valid RESP byte stream which is significant for redis pipelining mode with multiple cmds. So I pre-fetched the segments of command through func:

```
func (c *baseClient) generalProcessPipeline(ctx context.Context, cmds []Cmder, p pipelineProcessor)
```
I also added a example `httpRedis` to validate the new instrumentation.


minimAluminiumalism:
CR @damemi

MrAlias:
Thank you for your interest in the project. We are [currently not accepting new instrumentation](https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/591d260072381e5b0072a3e841785b3991d0d4f8/CONTRIBUTING.md#pull-requests). I have added this as [an issue](https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/1775). Please add any additional information about your use-case and the environment you run this package in in that issue.

PR #1528, opened 2024-12-29
Add gRPC error message for client spans

RonFed:


internal/pkg/instrumentation/bpf/google.golang.org/grpc/client/bpf/probe.bpf.c
```diff
@@ -139,6 +142,11 @@ int uprobe_ClientConn_Invoke_Returns(struct pt_regs *ctx) {
     //   s *Status
     // }
     // The `Status` proto object contains a `Code` int32 field, which is what we want
+    // type Status struct {
+	//   Code int32
+	//   Message string
+	//   Details []*anypb.Any
+    // }
```
MrAlias, line 149:
```suggestion
    // type Status struct {
    //     Code int32
    //     Message string
    //     Details []*anypb.Any
    // }
```

internal/test/e2e/grpc/verify.bats
```diff
@@ -112,11 +112,6 @@ SCOPE="go.opentelemetry.io/auto/google.golang.org/grpc"
   assert_equal "$client_span_id" "$server_parent_span_id"
 }
 
-@test "server :: expected (redacted) trace output" {
-  redact_json
-  assert_equal "$(git --no-pager diff ${BATS_TEST_DIRNAME}/traces.json)" ""
-}
-
```
MrAlias, line 119: Should this be removed?

internal/pkg/instrumentation/bpf/google.golang.org/grpc/server/bpf/probe.bpf.c
```diff
@@ -273,9 +273,12 @@ SEC("uprobe/server_handleStream2")
 int uprobe_server_handleStream2_Returns(struct pt_regs *ctx) {
     u64 server_stream_pos = 4;
     void *server_stream_ptr = get_argument(ctx, server_stream_pos);
+    struct go_iface go_context = {0};
+    void *key = NULL;
     if (server_stream_ptr == NULL) {
-        bpf_printk("grpc:server:uprobe/server_handleStream2Return: failed to get ServerStream arg");
-        return -1;
+        // we might fail get the pointer for versions of go which use register ABI, as this function does not return anything
+        // this is not an error in that case so we can just go to the lookup which will happen by goroutine
```
MrAlias, line 280:
```suggestion
        // We might fail to get the pointer for versions of Go which use register ABI, as this function does not return anything.
        // This is not an error in that case so we can just go to the lookup which will happen by goroutine.
```

internal/pkg/instrumentation/bpf/google.golang.org/grpc/server/bpf/probe.bpf.c
```diff
@@ -273,9 +273,12 @@ SEC("uprobe/server_handleStream2")
 int uprobe_server_handleStream2_Returns(struct pt_regs *ctx) {
     u64 server_stream_pos = 4;
     void *server_stream_ptr = get_argument(ctx, server_stream_pos);
+    struct go_iface go_context = {0};
+    void *key = NULL;
```
MrAlias, line 277: Why move the key declaration here?

RonFed: A matter of style I guess (declaring all the variables at the start),
I can move it back if you prefer

internal/test/e2e/grpc/verify.bats
RonFed: I moved it to the bottom of the file

internal/pkg/instrumentation/bpf/google.golang.org/grpc/server/bpf/probe.bpf.c
MrAlias: Nah, sounds good to me. I was wondering if I was missing something.

MrAlias (approved):

Issue #236, opened 2023-07-21
Add blog post on opentelemetry.io after beta release


pellared:
After release add a blog post in https://opentelemetry.io/  (e.g. https://opentelemetry.io/blog/2022/dotnet-instrumentation-first-beta/ )

MrAlias:
Having an update to our examples to highlight the integration with `otel` would be ideal for the blog post.

Maybe we could integrate https://github.com/MrAlias/otel-auto-demo here.

PR #1566, opened 2025-01-08
Update the `rolldice` example

MrAlias:
Part of #236 

This adapts the example found in https://github.com/MrAlias/otel-auto-demo into our current `rolldice` example.

This change highlights ...

- support for context propagation by splitting the application into two services
- multiple instrumentation probes by including a database to manage users
- the integration with `otel` global API and hybrid instrumentation


examples/rolldice/README.md
```diff
@@ -1,19 +1,105 @@
...
+
+### Context Propagation
+
+Looking at one of the traces within Jaeger we can see spans from both the `frontend` and `user` services.
+
+![Jaeger trace](./images/jaeger_trace.png)
+
+The `frontend` service has been manually instrumented.
+It does not use the OpenTelemetry Go auto-instrumentation to generate spans.
+It does so by directly using the OpenTelemetry API, SDK, and pre-built instrumentation packages.
```
RonFed, line 53: Maybe we should also auto-instrument the `frontend` service instead of manually instrumenting it?
I guess it has a value in showing the differences but since we want to emphasize the automatic aspects and not confuse maybe worth to consider.

MrAlias: Updated, PTAL>

RonFed (approved):

PR #1594, opened 2025-01-13
Makefile: docker-build dep of fixtures

MrAlias:
Instead of calling make for docker-build within fixtures/%, have that target be a dependency run prior to fixtures.


damemi (approved):
lgtm but just curious @MrAlias does this have some sort of benefit or is it just a style thing?

MrAlias:
> lgtm but just curious @MrAlias does this have some sort of benefit or is it just a style thing?

A little bit of both. Right now it is just a style thing, but I'm looking at splitting many of these dependencies out and then they become parallelized using the built in make tooling.

PR #1602, opened 2025-01-14
Split sample build to own target

MrAlias:
Allow the sample App to be built in parallel with the instrumentation container.


Makefile
```diff
@@ -169,12 +178,7 @@ fixture-otelglobal: fixtures/otelglobal
 fixture-autosdk: fixtures/autosdk
 fixture-kafka-go: fixtures/kafka-go
 fixtures/%: LIBRARY=$*
-fixtures/%: docker-build
-	if [ -f ./internal/test/e2e/$(LIBRARY)/build.sh ]; then \
-		./internal/test/e2e/$(LIBRARY)/build.sh; \
-	else \
-		cd internal/test/e2e/$(LIBRARY) && docker build -t sample-app . ;\
-	fi
+fixtures/%: docker-build sample-app/%
```
RonFed, line 181: I'm not sure if this will run these in parallel.
Maybe we need to add the `-j` flag for that: https://www.gnu.org/software/make/manual/html_node/Parallel.html

MrAlias: Yeah, adding the `j` flag is the intention. I should have made that more clear.

MrAlias: This should run exactly the same without the `j` flag, just it gives the opportunity to those who want faster builds to do so.

RonFed (approved):

Issue #1318, opened 2024-11-27
Support HTTP Traceparent header field reading in Go 1.24


grcevski:
Swiss maps have been enabled by default in Go main, which means they will probably become the default in Go 1.24 and break the reading of the HTTP request maps. gRPC is not affected, since we read an array there.

https://go-review.googlesource.com/c/go/+/574516

Go 1.24 is scheduled for release in February (2nd week), we need a fix by then: https://docs.google.com/presentation/d/1jvc0RzrshOOjkeEnsyea2jwjs_zVOcyETG558pqPOK0/mobilepresent?slide=id.g840eaeb4b4_0_0

One possible fix for this is to write a version of the map lookup code which is meant for Swiss maps. Based on some quick code reading I did, this looks more complex now than it used to be. 

Another possible solution we discussed at the Go Auto SIG today is potentially reading the Traceparent field while the headers are being parsed, although this will be slower since we will run a uprobe on every header field. The advantage of this approach is that this works for all Go versions.

grcevski:
I also realized that maybe we simply need to figure out how to work with the swiss maps. To support manual spans we also read a Go map in `fill_partial_tracer_id_from_tracers_map`, which also reads a Go map.

MrAlias:
We need to fully audit the existing probes to determine where we instrument using maps.

We should see if we can find short-term fixes for all these locations if we cannot get support for swiss maps working before their release.

MrAlias:
@grcevski plans to contribute the code from Beyla they used to get a stop-gap solution working for HTTP

MrAlias:
Doing a quick [search for `MAP_BUCKET_TYPE`](https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-go-instrumentation%20MAP_BUCKET_TYPE&type=code) shows only two locations we are using that definition:

- https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/d5869d513063a463ebdfa7dcc50eb2f2c4e255ad/internal/pkg/instrumentation/bpf/net/http/server/bpf/probe.bpf.c#L57
- https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/d5869d513063a463ebdfa7dcc50eb2f2c4e255ad/internal/pkg/instrumentation/bpf/go.opentelemetry.io/otel/traceglobal/bpf/probe.bpf.c#L120

[...]

PR #1636, opened 2025-01-16
Switch to using per header field parsing for Go versions >= 1.24

grcevski:
This PR is changing how we parse handle the Go HTTP request headers to switch from parsing the Go map to using per header uprobe.

The existing tests should be sufficient to prove this initial version works, since we are already testing for context propagation.

In this first version of the PR I've simply switched how the parsing is done to use the new approach. This approach works regardless of the Go version used, e.g. it will work for Go < 1.24 and Go >= 1.24.

I'm looking for feedback on if we should add a load time constant for the BPF program for the Go version. If we test for the Go version in the BPF code we can choose to use the old approach and not mount the new probe.

There are few pros and cons to each approach:

Proposal 1: We don't read any Go maps until we have swiss maps support

Pros:
- Simpler to test, since all of our tests will be exercising the same code regardless of the Go version used.
- Code is much simpler to reason about
- Go iterates quickly on releases and old versions are quickly dropped out of support

Cons:
- All go versions instrumentations are slower even for those that we know to read the maps for.

Proposal 2: We use the Go version to turn this on and off

Pros:
- Older Go version instrumentations will be faster
- If we ever implement support for swiss maps most of the code will already be there to handle < 1.24 and >= 1.24

Cons:
- We need to add Go 1.24 release candidate tests to prove this new code works
- Potential bugs might be version dependent
- Diminishing returns for the complexity with new Go releases, unless we implement swiss map support quickly.

Consensus:

We chose Proposal 2, where we check for Go version 1.24 to apply the new parsing based approach. I tested manually by modifying the min version in the file to say 1.23.0 and watched the kernel pipe for a specific message to appear that we are using the new approach. The debug message was removed from the final code.

We discussed in the Go Auto SIG the testing story, it seemed acceptable to wait for 1.24 to build the tests and if something doesn't work we'll address it then. I'm going to try to make a test with the pre-release 1.24 and follow up with another PR.

Relates to https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/1318


internal/include/utils.h
```diff
@@ -20,6 +20,18 @@ static __always_inline bool bpf_memcmp(char *s1, char *s2, s32 size)
     return true;
 }
 
+// assumes s2 is all lowercase
+static __always_inline int bpf_memicmp(const char *s1, const char *s2, s32 size) {
```
grcevski, line 24: I added a helper we have used before in Beyla which can do case insensitive compares for the header fields. e.g. Traceparent, TRACEPARENT, TraceParent, traceparent all match.

internal/pkg/instrumentation/bpf/net/http/server/bpf/probe.bpf.c
```diff
@@ -50,6 +50,14 @@ struct
     __uint(max_entries, MAX_CONCURRENT);
 } http_server_uprobes SEC(".maps");
 
+struct
+{
+    __uint(type, BPF_MAP_TYPE_HASH);
+    __type(key, void *);
+    __type(value, struct span_context);
+    __uint(max_entries, MAX_CONCURRENT);
+} http_server_context_headers SEC(".maps");
```
RonFed, line 59: We have maps used for context tracking in:
https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/5f77fdb4433aeb1ecb64395f4afd50970043df02/internal/include/go_context.h#L14-L21

To use these we need:
* pass NULL to `get_parent_span_context_fn` which will default to use `get_parent_span_context`.
* Call `start_tracking_span` to update the maps.
* Call  `stop_tracking_span` to delete the entries.

internal/pkg/instrumentation/bpf/net/http/server/probe.go
```diff
@@ -109,6 +109,10 @@ func New(logger *slog.Logger, version string) probe.Probe {
 					EntryProbe:  "uprobe_serverHandler_ServeHTTP",
 					ReturnProbe: "uprobe_serverHandler_ServeHTTP_Returns",
 				},
+				{
+					Sym:         "net/textproto.(*Reader).readContinuedLineSlice",
+					ReturnProbe: "uprobe_textproto_Reader_readContinuedLineSlice_Returns",
+				},
```
RonFed, line 115: We can use the package constraint we have for the go versions as defined in:

https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/5f77fdb4433aeb1ecb64395f4afd50970043df02/internal/pkg/instrumentation/bpf/go.opentelemetry.io/otel/traceglobal/probe.go#L60

This will only load the probe for version which have Swiss maps.
It also means will have to inject a constant specifying whether to extract the headers the old or the new way - I think that's a good solution until we'll have full support for Swiss maps.

internal/pkg/instrumentation/bpf/net/http/server/bpf/probe.bpf.c
grcevski: I looked into more detail here and I don't think I can use the existing span tracking functions. The `readContinuedLineSlice` function is called from other locations in Go and it would potentially pollute the maps with information we never clean up. I also realized that we need to make this extra storage LRU_HASH to avoid exhaustion if other uses of `readContinuedLineSlice` or `readMimeHeader` outside of http store data in there.

It's very unlikely that we'll store data from these other places because we look for `traceparent`, but it could happen.

LMK what you think.

internal/pkg/instrumentation/bpf/net/http/server/probe.go
grcevski: Done, thanks for the suggestion.

internal/pkg/instrumentation/bpf/net/http/server/bpf/probe.bpf.c
```diff
@@ -280,5 +299,29 @@ int uprobe_serverHandler_ServeHTTP_Returns(struct pt_regs *ctx) {
 
     stop_tracking_span(&http_server_span->sc, &http_server_span->psc);
     bpf_map_delete_elem(&http_server_uprobes, &key);
+    bpf_map_delete_elem(&http_server_context_headers, &key);
+    return 0;
+}
+
+SEC("uprobe/textproto_Reader_readContinuedLineSlice")
```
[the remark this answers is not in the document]
grcevski: Done.

grcevski:
@RonFed @MrAlias, do you know why I get this error in my k8s grpc test run?
```
{"time":"2025-01-26T22:57:39.277971911Z","level":"ERROR","source":{"function":"main.main","file":"/app/cli/main.go","line":125},"msg":"failed to load instrumentation","error":"could not find offset for function net/http.serverHandler.ServeHTTP"}
```

RonFed (approved):
Looks awesome, thanks!

CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ OpenTelemetry Go Automatic Instrumentation adheres to [Semantic Versioning](http
 - Support Go `1.22.11`. ([#1638](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/1638))
 - Support Go `1.23.5`. ([#1638](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/1638))
 - Support `google.golang.org/grpc` `1.70.0`. ([#1682](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/1682))
+- Support Go `1.24` swiss maps. ([#1636](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/1636))
```
MrAlias, line 31: This needs to be moved to the "Unreleased" section.

MrAlias, line 31:
```suggestion
- Preemptively update instrumentation for `net/http` to support Go `1.24` when swiss maps are going to be added. ([#1636](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/1636))
```

internal/pkg/instrumentation/bpf/net/http/server/probe.go
```diff
@@ -28,9 +28,19 @@ import (
...
+	pkg            = "net/http"
+	minGoSwissMaps = "1.24.0"
 )
 
+var goWithSwissMaps = probe.PackageConstrainst{
+	Package: "std",
+	Constraints: version.MustConstraints(
+		version.NewConstraint(">= " + minGoSwissMaps),
+	),
+	// Don't warn, we have a backup path
```
MrAlias, line 40:
```suggestion
	// Don't warn, we have a backup path.
```

internal/pkg/instrumentation/probe/probe.go
```diff
@@ -188,7 +188,7 @@ func (i *Base[BPFObj, BPFEvent]) loadUprobes(exec *link.Executable, td *process.
 				logFn = i.Logger.Warn
 			default:
 				// Unknown and FailureModeError.
-				return fmt.Errorf("uprobe %s package constraint (%s) not meet", up.Sym, pc.Constraints.String())
+				return fmt.Errorf("uprobe %s package constraint (%s) not met, version %v", up.Sym, pc.Constraints.String(), td.Modules[pc.Package])
```
MrAlias, line 191:
```suggestion
				return fmt.Errorf("uprobe %s package constraint (%s) not meet, version %v", up.Sym, pc.Constraints.String(), td.Modules[pc.Package])
```

MrAlias (commented):
Overall looks good, just comment and doc cleanup. :+1:

CHANGELOG.md
grcevski: Done! Thanks!

internal/pkg/instrumentation/probe/probe.go
grcevski: Hey @MrAlias, I'm happy to change it back from 'met' to 'meet', I thought it was more correct form. The package constraint is not met, I think, or maybe I misunderstand the wording.

CHANGELOG.md
grcevski: I think have under:
```
## [Unreleased]
### Added
```

Or maybe I need to merge main to get the beta released section. I'll do that now.

grcevski: I merged main and moved this fix to the unreleased section.

MrAlias (approved):

PR #1641, opened 2025-01-18
support stripped binaries built wit CGO_ENABLED

### LINKED ISSUE #1637: "error":"Malformed version: (devel)","dependency"
{"time":"2025-01-17T02:29:30.830390503Z","level":"ERROR","source":{"function":"go.opentelemetry.io/auto/internal/pkg/process.(*Analyzer).Analyze","file":"/app/internal/pkg/process/analyze.go","line":87},"msg":"parsing dependency version","error":"Malformed version: (devel)","dependency":{"Path":"msgcenter","Version":"(devel)","Sum":"","Replace":null}}
{"time":"2025-01-17T02:29:30.849378599Z","level":"ERROR","source":{"function":"main.main","file":"/app/cli/main.go","line":119},"msg":"failed to create instrumentation","error":".gosymtab section not found in target binary, make sure this is a Go application"}

RonFed:
Resolves #1637 and improves our support for stripped binaries.

* The `.gosymtab` section is not required to parse the pcln table as stated in https://pkg.go.dev/debug/gosym#NewTable
* When calling [NewLineTable](https://pkg.go.dev/debug/gosym#NewLineTable), we need to pass the value of `runtime.text` symbol from the instrumented binary. For binaries not including C code, this is the same as the start of the `.text` section. However for those that use C code, this value will differ. This caused incorrect offsets to get probed when instrumenting a stripped binary that uses a CGO lib. An example of this is our DB example and test which uses the [go-sqlite3 library](https://github.com/mattn/go-sqlite3).


Makefile
```diff
@@ -120,7 +120,7 @@ docker-build:
 docker-build-base:
 	docker buildx build -t $(IMG_NAME_BASE) --target base .
 
-.PHONY: sample-app/nethttp sample-app/gin sample-app/databasesql sample-app/nethttp-custom sample-app/otelglobal sample-app/autosdk sample-app/kafka-go
+.PHONY: sample-app/%
```
MrAlias, line 123: This is not expanded in a `.PHONY` target. It removes the functionality and completion of this target. It should not be changed.

RonFed: The current version of the Makefile is not working for me for the `fixture` targets. The sample-app targets are not executed.
```Bash
make sample-app/nethttp
make: Nothing to be done for `sample-app/nethttp'.
```

This change fixed that, I'm not sure why.

grcevski (approved):
LGTM! Looks great!

Makefile
MrAlias: Can we move changes to the Makefile to their own PR so we can decouple this issue from the intent of this PR?

RonFed: Removed from this PR

PR #1681, opened 2025-01-23
Fix issues with span tests in SDK

MrAlias:
Found in building https://github.com/open-telemetry/opentelemetry-go/pull/6203

- Run the function returned from assertTracer.
- Call the span passed to the goroutine.

Planning to follow-up with a PR ensuring golangci-lint catches these in this repository.


MrAlias:
Superseded by #1688

PR #1725, opened 2025-01-31
Add defensive checks for CGO stripped binaries

grcevski:
This is a follow-up PR to https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/1641. I think we should add these defensive checks in case of corrupted .gopclntab sections. It's unlikely to happen, but I thought I should propose a PR since I recently reviewed the code.

### DIFF
--- a/internal/pkg/process/binary/funcs_stripped.go
+++ b/internal/pkg/process/binary/funcs_stripped.go
@@ -22,6 +22,10 @@ func FindFunctionsStripped(elfF *elf.File, relevantFuncs map[string]interface{})
 		return nil, err
 	}
 
+	if len(pclndat) <= 8*2*8 {
+		return nil, errors.New(".gopclntab section too small")
+	}
+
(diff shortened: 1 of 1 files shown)


internal/pkg/process/binary/funcs_stripped.go
```diff
@@ -22,6 +22,10 @@ func FindFunctionsStripped(elfF *elf.File, relevantFuncs map[string]interface{})
 		return nil, err
 	}
 
+	if len(pclndat) <= 8*2*8 {
```
RonFed, line 25: Can you please add a comment on why this is the minimum size?

RonFed (approved):

internal/pkg/process/binary/funcs_stripped.go
grcevski: Added thanks!

MrAlias (approved):
:star:

PR #1732, opened 2025-02-03
Automate the update of `libbpf`

### LINKED ISSUE #961: Add libbpf update Github workflow
Using the `update.sh` script added in https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/957, add a workflow that will check for newer versions of libbpf, and update everything appropriately if found.

This needs to be manually triggered as well as run periodically similar to the [offsets action](https://github.com/open-telemetry/opentelemetry-go-instrumentation/actions/workflows/offsets.yml).

MrAlias:
Resolve #961

- Replace `internal/include/libbpf/update.sh` with the `internal/tools/synclibbpf` go utility.
- Add a make target to manage updates locally.
- Add a Github workflow to automate the update.
- Document the update process in CONTRIBUTING.md
- "Update" from `v1.4.5` to `v1.4.7` (this is a no-op).


internal/tools/synclibbpf/main.go
```diff
@@ -0,0 +1,169 @@
...
+	"fmt"
+	"io"
+	"log"
+	"net/http"
+	"os"
+	"path/filepath"
+	"sort"
+	"strings"
+
+	"github.com/hashicorp/go-version"
```
kakkoyun, line 21: nit: Since you decided to remove `github.com/hashicorp/go-version` package in https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/1766, should we also use the new package here for the sake of consistency?

kakkoyun (approved):
LGTM.

Only added a comment, a nitpick ⛏️

internal/tools/synclibbpf/main.go
MrAlias: Good catch. I'll be sure to update here or there when either of these PRs merge. Thanks :+1:

RonFed (approved):

PR #1766, opened 2025-02-06
Replace use of `github.com/hashicorp/go-version`

### LINKED ISSUE #1753: Replace use of `github.com/hashicorp/go-version`
We currently use [github.com/hashicorp/go-version](https://pkg.go.dev/github.com/hashicorp/go-version) to represent versions throughout the project. There are many limitations and unexpected behavior that that this project contains that warrant a move away from its use.

### No way to construct a new value without also returning an error.

If you want to create a new `version.Version` you can only do so by passing a string that is parsed with `version.NewVersion`. This means an error type is always returned with the parsed version. If you know the version is valid (i.e. a static value in a test), you should be able to create a version directly

### Comparisons are not evaluated according to semantic conventions

We use `version.Constraints` to match package versioning. However, github.com/hashicorp/go-version does not evaluate these as [specified by semantic versioning](https://semver.org/#spec-item-11).

- https://github.com/hashicorp/go-version/issues/130
- https://github.com/hashicorp/go-version/issues/36

The package takes its own approach to handling prereleases.

This means if we add a package constraint like `> 1.34.0` and use a package with a version like `"1.34.1-0.20250205150531-85fab8be9320"` it will evaluate to false even though according to semantic versioning this should be true

### `Version` is not comparable

The [`Version` type](https://github.com/hashicorp/go-version/blob/ec1a6b6f9ca05cd26710bf07c9a58c1809a4cb05/version.go#L36-L43) is defined as such:

```go
// Version represents a single version.
type Version struct {
	metadata string
	pre      string
	segments []int64
	si       int
	original string
}
```

Based on the way the segments field is defined, this type is not comparable.

An `Equal` method is provided, but it would be more idiomatic if we could compare equality with the `=` Go operator.

Also, this mean the `Version` cannot be used directly as a map key. This results in us having to manually make the parts comparable:

https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/9f23c8918881992acdd2d883fba9068b643e3dc5/internal/pkg/structfield/structfield.go#L359-L363

### The `version` package name sits on a useful variable name

Commonly using `version` for a variable name will be descriptive of what we are trying to hold. However, this will conflict with the package name.

### Code health

Over the last 11 years there have been 73 commits: https://github.com/hashicorp/go-version/graphs/contributors

There are 23 open Pull Requests (some > 10 years old): https://github.com/hashicorp/go-version/pulls

There are 18 open issues and 30 closed issues: https://github.com/hashicorp/go-version/issues

There are 8 releases: https://github.com/hashicorp/go-version/releases

### Licensing

`github.com/hashicorp/go-version` is licensed under the [Mozilla Public License, version 2.0](https://github.com/hashicorp/go-version/blob/main/LICENSE). This is a weak copyleft license that prohibits trademark use, and requires source disclosure, license and copyright notice, and same license inheritance.

This is not a blocking issue as we are complying with all the terms currently, however there are more permissively licensed projects we could consider.

## Proposal

Switch to using [`github.com/Masterminds/semver/v3`](https://pkg.go.dev/github.com/Masterminds/semver/v3)

- versions can be constructed directly without needed to handle an error: https://pkg.go.dev/github.com/Masterminds/semver/v3#New
- the [`Version type`](https://github.com/Masterminds/semver/blob/1558ca3488226e3490894a145e831ad58a5ff958/version.go#L48-L54) is comparable
- comparison is done based on semantic versioning: https://pkg.go.dev/github.com/Masterminds/semver/v3#hdr-Basic_Comparisons
- the package name is not `version`, it is `semver`
- code health
  - 157 commits over the last 10 years
  - 11 open PRs (oldest 3 years old)
  - 13 opne issues, 84 closed issues
  - 22 releases
- Licensed under the MIT License which is a permissive license with conditions only requiring preservation of copyright and license notices.

Auditing the package, there appears to be a 100% feature overlap with how we currently use `github.com/hashicorp/go-version`

- There is a `Version` type that represents Go versions
- The `Version` type supports comparison
- There is a `Constraint` type that supports matching versions
- Supports sorting versions.
- Supports JSON and other text marshaling
- Supports Go >= 1.21

Additionally ...

- the `Version` type can be used as a map key given its comparability.
- [wildcard validation](https://pkg.go.dev/github.com/Masterminds/semver/v3#hdr-Wildcards_In_Comparisons) is supported.
- has [no dependencies](https://github.com/Masterminds/semver/blob/1558ca3488226e3490894a145e831ad58a5ff958/go.mod) other than the Go standard library

MrAlias:
Resolve #1753

Replace with `github.com/Masterminds/semver/v3`


MrAlias:
Looking into CI failures

MrAlias:
The CI failure looks to be coming from offsets not being correctly identified. The offsets are stored as "X.Y.Z" without a prefix of "v" (i.e. not "vX.Y.Z"). The versions being matched to these in the index all use a "v" prefix for the "original" and are therefore not matching.

I'm looking into a solution.

kakkoyun (approved):
LGTM.

I have used both of the packages before and gone through similar migration. The changes look straight forward.

internal/pkg/inject/consts_test.go
```diff
@@ -6,7 +6,7 @@ package inject
 import (
 	"testing"
 
-	"github.com/hashicorp/go-version"
+	"github.com/Masterminds/semver/v3"
```
pellared, line 9: There is also https://pkg.go.dev/golang.org/x/mod/semver

But I have not checked if it would be better. Possibly it is lighter.

pellared (approved):

internal/pkg/inject/consts_test.go
pellared: I guess you mean https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/1753. Resolving

PR #1815, opened 2025-02-13
Remove support for Go < 1.19 instrumentation

### LINKED ISSUE #1812: Drop instrumentation support for Go < 1.19
As was pointed out [here](https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/1771#issuecomment-2643842278), we have not supported Go 1.19 ...

> [...] for 9 months: https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/790
> That is 8 releases without it being reported.

[In our last SIG meeting](https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/1771#issuecomment-2651885429) we reached an agreement that we will drop instrumentation support for Go < 1.19 to clean up the code-base and reduce development burden going forward.

Backwards compatible support can be added back in the future. However, dropping this once we stabilize will be more problematic. Based on this the consensus was to remove this support prior to the stabilization of this project.

- [ ] Update our `inspect` tooling to not generate offsets for Go < 1.19
- [ ] Remove context tracking support for Go < 1.17 in our included C libraries
- [ ] Update our compatibility documentation

MrAlias:
Resolve #1812


internal/include/arguments.h
```diff
@@ -71,42 +68,13 @@ static __always_inline void *get_argument_by_reg(struct pt_regs *ctx, int index)
...
-    if (is_registers_abi)
-    {
-        return (void *)GOROUTINE(ctx);
-    }
-
-    return contextContext;
-}
-
-static __always_inline bool is_register_abi()
+static __always_inline void *get_consistent_key(struct pt_regs *ctx)
```
RonFed, line 75: Do you think we should keep this function, or just switch to getting the goroutine pointer directly?

RonFed (approved):
🚀

internal/include/arguments.h
MrAlias: Good question. I figured keeping it reduced the review burden, but I'm also on board with replacing it with directly with just the direct goroutine pointer access.

MrAlias: Removed.

PR #1890, opened 2025-02-27
Move process discovery out of `auto`

### LINKED ISSUE #1887: Move process discovery from `auto` into `auto/cli`
From recent discussions, the plan is make the `auto.Instrumentation` have no process discovery capabilities. It will accept a PID to target.

The existing process discovery needs to be moved into the CLI to preserve functionality of our distributed executable.

MrAlias:
Resolve #1887

- Have the CLI manage process discovery instead of the `auto.NewInstrumentation`.
  - Parse `OTEL_GO_AUTO_TARGET_EXE` in the CLI instead of `auto.NewInstrumentation`
  - Remove the `WithTarget` option
- Add support for the `-target-exe` CLI flag to configure the target executable path
- Add support for the `-target-pid` CLI flag to configure the target pid
- Add support for parsing `OTEL_GO_AUTO_TARGET_PID` in the CLI
- Add tests for process discovery


damemi (commented):
@MrAlias wdyt about moving the entire `internal/pkg/process` to its own top-level package, such as `go.opentelemetry.io/auto/process`?

This is coming from looking at the Probe api, which is currently tied to the `process` package through the [targetdetails arg](https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/3fe6adc72f328b85e03c4ee679be2087ea2560f1/internal/pkg/instrumentation/probe/probe.go#L43)

Given that not all probes will be tied to a process, and process detection and management is going to move up to the consumer level, I think it makes sense to not have it be tied in that way. But it could be useful for the consumers who are implementing process management for us to provide that code as a helper package. Curious what you think?

MrAlias:
> @MrAlias wdyt about moving the entire `internal/pkg/process` to its own top-level package, such as `go.opentelemetry.io/auto/process`?
> 
> This is coming from looking at the Probe api, which is currently tied to the `process` package through the [targetdetails arg](https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/3fe6adc72f328b85e03c4ee679be2087ea2560f1/internal/pkg/instrumentation/probe/probe.go#L43)
> 
> Given that not all probes will be tied to a process, and process detection and management is going to move up to the consumer level, I think it makes sense to not have it be tied in that way. But it could be useful for the consumers who are implementing process management for us to provide that code as a helper package. Curious what you think?

I'm looking at refactoring the `process` package right now. There is a lot of clean up needed there.

I agree we will need some top-level package to represent a package for the probe API. I'm not in favor of just moving the whole package as is given things I'm seeing there, but we can surely add something at the top level that makes sense when we add the probe API.

That said, I do not plan to move the `process` package in this PR. I see this as part of the larger refactor of that package and a step towards moving parts of it.

damemi:
@MrAlias sounds good, and yeah I assumed that there are refactoring changes needed (didn't mean that we could just literally move the entire package wholesale as-is). My main point is having something external that can be used with the external Probe api. I think that will be a prerequisite for that refactor

MrAlias:
Gotcha, sounds like we're on the same page 😀

RonFed (approved):

PR #1906, opened 2025-02-28
Move probe initialization to `auto`

MrAlias:
Instead of initializing all probes within the manger, have Instrumentation handle the creation of all wanted probes.

This cleans up unneeded context being passed to the NewManager call (i.e. globalImpl and Version()) and these values can be used where they are defined.

This also is a first step in accepting probes as options to Instrumentation which is a goal of #1105.


damemi (approved):
Looks great

> This also is a first step in accepting probes as options to Instrumentation which is a goal of https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/1105.

Just so I understand, you mean eventually passing the probes to `NewInstrumentation`? That's what I had in mind

PR #1935, opened 2025-03-05
remove .PHONY for sample-app docker builds

RonFed:
Currently, using either the `fixture` targets or the `sample-app` targets directly results in a "nothing to do" message from make.
Removing the .PHONY fixes that.
local docker caching makes sure that nothing is rebuilt if not necessary.


grcevski (approved):
LGTM!

MrAlias:
Removing this and having it run every time the make target or one with this as a dependency is run removes the usefulness of the target. It should be removed and re-integrated into the `fixture` target if we want to go in this direction.

RonFed:
> Removing this and having it run every time the make target or one with this as a dependency is run removes the usefulness of the target. It should be removed and re-integrated into the `fixture` target if we want to go in this direction.

Updated, PTAL

MrAlias (approved):

pellared (approved):

PR #1936, opened 2025-03-05
fix get_go_string_from_user_ptr to add null byte at the end of a string

RonFed:


grcevski (approved):
LGTM! Nice catch!

MrAlias (approved):

MrAlias:
Can you add a changelog describing the fix?

skandragon:
> LGTM! Nice catch!

You're welcome :)

We knew we did not have any kafkaesque topics like we were seeing, and from my C days, I pretty much knew what the problem was, just not where.

skandragon:
What can I do to help move this into completion?

rafaelroquetto:
@RonFed @MrAlias are we able to update the clang version in the generator image to something like clang-19? It seems to be using the fairly ancient clang-14. Clang-19 is able to to generate compliant code to make the tests pass on 5.14, but otherwise, there's an issue with the checking of the bonds - the verifier is not seing that `0xFFFFF`, looking into it.

rafaelroquetto:
Ok, so there are a few problems here:

the verifier is not being able to see through the integer narrowings (in particular the ternary operator). Doing something such as:
```

    u32 size = user_str.len & 0xffff;
    u32 mlen = max_len & 0xffff;

    if (size > mlen) {
        size = mlen;
    } else {
        dst[size] = '\0';
    }

    success = bpf_probe_read_user(dst, size, user_str.str);
    if (success != 0) {
        return false;
    }
```
fixes most of the cases, but then there's the complaint about about `variable stack access`. That is the `dst[size_to_read] = '\0'` part when `dst` is on the stack. `size_to_read` is a variable, and older kernels forbid that - see https://lore.kernel.org/lkml/20210426072819.147955757%40linuxfoundation.org/ - this limitation got dropped from 5.10. For some reason, clang-19 is able to generate code that works around that.

In any case, the workaround is to index on the "constant", i.e. `max_len`, by the means of `__builtin_memset(dst, 0, max_len)`. I will add a suggestion to the PR with that, that should make all tests pass on 5.4

internal/include/go_types.h
```diff
@@ -161,26 +161,29 @@ static __always_inline void append_item_to_slice(void *new_item, u32 item_size,
...
+    u32 size = size_to_read & 0xFFFFF;
+    success = bpf_probe_read_user(dst, size, user_str.str);
+    if (success != 0) {
         return false;
     }
 
+    if (size_to_read < max_len) {
+        dst[size_to_read] = '\0';
+    }
+
```
rafaelroquetto, line 186:
```suggestion
    __builtin_memset(dst, 0, max_len)
    
    success = bpf_probe_read_user(dst, size_to_read, user_str.str);
    if (success != 0) {
        return false;
    }
```

RonFed:
Thanks @rafaelroquetto!, this did the trick.

PR #1937, opened 2025-03-05
zero each kafka span before populating

RonFed:


internal/pkg/instrumentation/bpf/github.com/segmentio/kafka-go/producer/bpf/probe.bpf.c
```diff
@@ -144,13 +144,24 @@ int uprobe_WriteMessages(struct pt_regs *ctx) {
...
+    if (zero_kafka_request == NULL)
     {
-        bpf_printk("uuprobe/WriteMessages: kafka_request is NULL");
+        bpf_printk("uuprobe/WriteMessages: zero_kafka_request is NULL");
         return 0;
     }
+
+    u32 actual_id = 1;
+    // Zero the span we are about to build, eBPF doesn't support memset of large structs (more than 1024 bytes)
+    bpf_map_update_elem(&kafka_request_storage_map, &actual_id, zero_kafka_request, 0);
```
grcevski, line 157: Very clever trick! should the update have BPF_ANY instead of 0? I don't think it would ever fail, but maybe it's worth checking if the update succeeded.

RonFed: updated to use `BPF_ANY`.
regarding error handling, not sure if we should drop the span if we failed to zero the memory or not.

MrAlias (approved):

PR #1941, opened 2025-03-05
Replace `Analyzer` with `NewInfo` function

MrAlias:
Replace the `Analyzer` with a `NewInfo` function in the process package.


internal/pkg/process/info.go
```diff
@@ -0,0 +1,120 @@
...
+	"runtime/debug"
+
+	"github.com/Masterminds/semver/v3"
+
+	"go.opentelemetry.io/auto/internal/pkg/process/binary"
+)
+
+// Info are the details about a target process.
+type Info struct {
+	ID         ID
```
RonFed, line 19:
```suggestion
	PID         ID
```

RonFed (approved):

internal/pkg/process/info.go
MrAlias: I want to match the type name here given we are still in the same package.

RonFed (approved):

PR #1943, opened 2025-03-05
Move process Info creation into NewManager

MrAlias:
The Manager manages probes for a process. Have the manager handle the process info creation and coordination instead of extraneously passing it from the `Instrumentation`.


internal/pkg/instrumentation/manager.go
```diff
@@ -55,19 +55,43 @@ type Manager struct {
 }
 
 // NewManager returns a new [Manager].
-func NewManager(logger *slog.Logger, otelController *opentelemetry.Controller, cp ConfigProvider, probes ...probe.Probe) (*Manager, error) {
+func NewManager(logger *slog.Logger, otelController *opentelemetry.Controller, pid process.ID, cp ConfigProvider, probes ...probe.Probe) (*Manager, error) {
```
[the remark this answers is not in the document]
damemi: Also this function signature is growing, maybe an options struct here would be useful?

internal/pkg/instrumentation/manager.go
```diff
@@ -55,19 +55,43 @@ type Manager struct {
...
+		if err := m.registerProbe(p); err != nil {
+			return nil, err
+		}
+
+		for _, s := range p.Manifest().Symbols {
+			funcs[s.Symbol] = nil
+		}
+	}
+
+	pa := process.NewAnalyzer(logger, pid)
```
damemi, line 77: do you think eventually the process analysis could be done outside of the manager and passed into it? thinking along the lines of making process totally public

damemi: process analysis identifies what probes should be loaded, and filters unused probes. For a tool that's implementing its own process management, couldn't that analysis/filtering be done outside the manager? The single process ID can still be passed to the manager, but why should filtering unused probes be embedded inside the manager?

damemi: I guess I am trying to define the line between process analysis and probe management. Not proposing moving probe management outside of the project, I'm saying *process* management and analysis could be a top level package of this project used in conjunction with the probe manager instead of as a dependency to the manager

internal/pkg/instrumentation/manager.go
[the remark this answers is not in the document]
damemi: > the process ID only contains an integer type that represents the PID

I meant that `process.ID` is able to provide the path and buildinfo too, so maybe it's semantic (since they're all derived from the PID) but that feels like it covers more than just PID

For passing it as an argument, how will this be handled for probes that don't attach to a process like TC probes that attach to a network interface?

internal/pkg/instrumentation/manager.go
MrAlias: If I understand you correctly, your question is more about the function filtering done here? The "build a function set", "interrogate the process for its functions", "filter the process functions", and then "filter the probes based on functions found" workflow, right?

If so, and if your point is this is overly complex, then we're on the same page.

I've given this some though, but I'm not ultimately sure how we can clean it up and keep the same performance behavior. I had considered just having `process.NewInfo` not take a function filter set and just hold all the found functions. Then when that `Info` is passed to the `Manager` the manager would only need to see what was there and filter its probes accordingly. This means though, that we will hold in memory all functions listed within a process instead of just a subset. This doesn't see ideal.

I don't see a way to achieve this decoupling without performance degradation. Do you?

internal/pkg/instrumentation/manager.go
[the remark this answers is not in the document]
MrAlias: I thought the conclusion from the custom probe discussion is that non-Go instrumentation probes are not meant to be handled by this project?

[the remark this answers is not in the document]
MrAlias: > I meant that `process.ID` is able to provide the path and buildinfo too, so maybe it's semantic (since they're all derived from the PID) but that feels like it covers more than just PID

Sure, I'm not sure how that affects the changes being proposed in this PR. Is this just a comment about #1916? Are you saying we should just pass an `int` here?

internal/pkg/instrumentation/manager.go
MrAlias: @damemi are your suggesting that we keep the process `Info` creation in `NewInstrumentation`?

RonFed (approved):

internal/pkg/instrumentation/manager.go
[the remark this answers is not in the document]
damemi: It doesn't really affect the changes here, it was more for my understanding of how we plan to use/define `process.ID` and get your thoughts on that. Not a blocker for this

internal/pkg/instrumentation/manager.go
damemi: @MrAlias just to be clear, I'm not trying to block this PR on this topic, just trying to talk about it conceptually. Maybe this isn't the place for that discussion, sorry about that.

I do agree with you about the complexity, but my main goal is to get that through decoupling. Ideally, I think the `Info` creation could happen even outside of `NewInstrumentation`, with `NewInstrumentation` accepting whatever info was created. That's what I mean by fully exporting the `process` package, making it a top level package, and removing all the process stuff from the manager.

Maybe that's not possible or efficient, but that's what I was trying to get at to see if you were thinking along those lines too

damemi (approved):

internal/pkg/instrumentation/manager.go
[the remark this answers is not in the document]
MrAlias: Gotcha, happy to discuss the envisioned uses of `ID`. I've added a topic to this weeks SIG meeting to talk more about the design.

internal/pkg/instrumentation/manager.go
MrAlias: I think we can decouple them a bit more, but I'm not entirely sure we will drop `Info` creation entirely from the `Manager` given we want to scope its contents by probes in the end.

I added a topic to today's SIG meeting to discuss more there.

PR #1945, opened 2025-03-05
Lazy allocate memory for process 

MrAlias:
Only allocate memory for a process if a probe requires the allocation.

Move the allocation functionality to be a method of the process.Info. This is intrinsically linked to a single process, therefore, encapsulate the functionality as a method.

### DIFF
--- a/internal/pkg/instrumentation/manager_load_test.go
+++ b/internal/pkg/instrumentation/manager_load_test.go
@@ -80,6 +80,13 @@ func TestLoadProbes(t *testing.T) {
 			ProbesLoad(t, info, p.(TestProbe))
 		})
 	}
+
+	// The grpcClient, grpcServer, httpClient, dbSql, kafkaProducer,
+	// kafkaConsumer, autosdk, and otelTraceGlobal all allocate. Ensure it has
+	// been called.
+	a, err := info.Alloc(logger)
+	require.NoError(t, err)
+	assert.NotEmpty(t, a.StartAddr, "memory not allocated")
 }
 
 const mainGoContent = `package main
--- a/internal/pkg/instrumentation/manager_load_test.go
+++ b/internal/pkg/instrumentation/manager_load_test.go
@@ -48,16 +48,11 @@ func TestLoadProbes(t *testing.T) {
 	// Reset Info module information.
 	info.Modules = make(map[string]*semver.Version)
 
-	logger := slog.Default()
-	info.Allocation, err = process.Allocate(logger, pid)
-	if err != nil {
-		t.Fatalf("failed to allocate for process %d: %v", id, err)
-	}
-
 	ver := utils.GetLinuxKernelVersion()
 	require.NotNil(t, ver)
 	t.Logf("Running on kernel %s", ver.String())
 
+	logger := slog.Default()
 	probes := []probe.Probe{
 		grpcClient.New(logger, ""),
 		grpcServer.New(logger, ""),
--- a/internal/pkg/process/info.go
+++ b/internal/pkg/process/info.go
@@ -7,9 +7,12 @@ import (
 	"debug/elf"
 	"errors"
 	"fmt"
+	"log/slog"
 	"regexp"
 	"runtime/debug"
 	"strings"
+	"sync"
+	"sync/atomic"
 
 	"github.com/Masterminds/semver/v3"
 
@@ -25,9 +28,24 @@ type Info struct {
 	// Experimental and build information included in the version is dropped.
 	// If a development version of Go is used, the commit hash will be included
 	// in the metadata of the version.
-	GoVersion  *semver.Version
-	Modules    map[string]*semver.Version
-	Allocation *Allocation
+	GoVersion *semver.Version
+	Modules   map[string]*semver.Version
+
+	allocOnce onceResult[*Allocation]
+}
+
+// Alloc allocates memory for the process described by Info i.
+//
+// The underlying memory allocation is only successfully performed once for the
+// instance i. Meaning, it is safe to call this multiple times. The first
+// successful result will be returned to all subsequent calls. If an error is
+// returned, subsequent calls will re-attempt to perform the allocation.
+//
+// It is safe to call this method concurrently.
+func (i *Info) Alloc(logger *slog.Logger) (*Allocation, error) {
+	return i.allocOnce.Do(func() (*Allocation, error) {
+		return allocate(logger, i.ID)
+	})
 }
 
 // NewInfo returns a new Info with information about the process identified by
@@ -151,4 +169,39 @@ func (i *Info) GetFunctionReturns(name string) ([]uint64, error) {
 	}
 
 	return nil, fmt.Errorf("could not find returns for function %s", name)
+}
+
+// onceResult is an object that will perform exactly one action if that action
+// does not error. For errors, no state is stored and subsequent attempts will
+// be tried.
+type onceResult[T any] struct {
+	done atomic.Bool
+	mu   sync.Mutex
+	val  T
+}
+
+// Do runs f only once, and only stores the result if f returns a nil error.
+// Subsequent calls to Do will return the stored value or they will re-attempt
+// to run f and store the result if an error had been returned.
+func (o *onceResult[T]) Do(f func() (T, error)) (T, error) {
+	if !o.done.Load() {
+		// Outlined complex-path to allow inlining here.
+		return o.do(f)
+	}
+	return o.val, nil
+}
+
+func (o *onceResult[T]) do(f func() (T, error)) (T, error) {
+	o.mu.Lock()
+	defer o.mu.Unlock()
+
+	if !o.done.Load() {
+		v, err := f()
+		if err != nil {
+			return v, err
+		}
+		o.val = v
+		o.done.Store(true)
+	}
+	return o.val, nil
 }
--- a/internal/pkg/process/info.go
+++ b/internal/pkg/process/info.go
@@ -31,7 +31,9 @@ type Info struct {
 	GoVersion *semver.Version
 	Modules   map[string]*semver.Version
 
-	allocOnce onceResult[*Allocation]
+	aDone atomic.Bool
+	aMu   sync.Mutex
+	a     *Allocation
 }
 
 // Alloc allocates memory for the process described by Info i.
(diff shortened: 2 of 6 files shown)


RonFed:
What do you think of handling this in the manager level? since it is something that happens once, instead of having each probe try to allocate and the first one win, have the manager check once if any of the probes require allocation and if so, do it once.

MrAlias:
> What do you think of handling this in the manager level? since it is something that happens once, instead of having each probe try to allocate and the first one win, have the manager check once if any of the probes require allocation and if so, do it once.

I like this approach better given it puts the responsibility on the thing that needs the allocation. Having a manger perform additional checks is not only unneeded overhead in the form of an additional "check" loop, it decouples a needed step for the probe initialization away from the probe itself.

This decoupling error prone and adds cognitive development burden. It introduces the possible error of when a manager makes a change in its handling that will result in breakages to a probe. This means that developers need to implicitly be aware of this non-transparent synchronous interaction.

The alternative presented here does not have these problems. When a probe needs an allocation it asks for it. The developer immediately knows this by looking at the probe alone and there are no side-effects that can cause errors.

RonFed:
@MrAlias from the perspective of a developer writing a probe, this design might lead them to think that each probe will allocate its own memory.

> When a probe needs an allocation it asks for it. The developer immediately knows this by looking at the probe alone and there are no side-effects that can cause errors.

This is still true with the current design - each probe declares whether it needs allocation using the `const` - and the manager will take care of the rest.

MrAlias:
> This is still true with the current design - each probe declares whether it needs allocation using the `const` - and the manager will take care of the rest.

Right. And this is looking to fix that by removing side-effects.

MrAlias:
> @MrAlias from the perspective of a developer writing a probe, this design might lead them to think that each probe will allocate its own memory.

Is that a problem? Why wouldn't they think this is allocating the memory needed for their probe?

RonFed (commented):
I still think that doing the allocation in the manager init is more straight forward.
What do you think about having the `Allocation` const in each probe setting a boolean to the `Info` (or some other common struct) indicating the allocation is required.
Then the manager will examine this value and allocate if necessery.
This approach has the advantages that:
1. It's idempotent  when the probes are requesting allocation.
2. It is guaranteed to happen once without to worry about concurrency.
3. Simpler.

internal/pkg/process/analyze.go
```diff
@@ -104,3 +122,36 @@ func (a *Analyzer) findFunctions(elfF *elf.File, relevantFuncs map[string]interf
...
+	val   T
+}
+
+// Do runs f only once, and only stores the result if f returns a nil error.
+// Subsequent calls to Do will return the stored value or they will re-attempt
+// to run f and store the result if an error had been returned.
+func (o *onceResult[T]) Do(f func() (T, error)) (T, error) {
+	if o.done.Load() {
+		o.mutex.Lock()
+		defer o.mutex.Unlock()
```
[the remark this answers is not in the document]
MrAlias: This is added to ensure the setting function call has returned.

internal/pkg/process/analyze.go
```diff
@@ -104,3 +122,36 @@ func (a *Analyzer) findFunctions(elfF *elf.File, relevantFuncs map[string]interf
...
+type onceResult[T any] struct {
+	done  atomic.Bool
+	mutex sync.Mutex
+	val   T
+}
+
+// Do runs f only once, and only stores the result if f returns a nil error.
+// Subsequent calls to Do will return the stored value or they will re-attempt
+// to run f and store the result if an error had been returned.
+func (o *onceResult[T]) Do(f func() (T, error)) (T, error) {
```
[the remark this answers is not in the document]
MrAlias: Sure, let me look into simplifying.

MrAlias:
> 1. It's idempotent when the probes are requesting allocation.
> 3. Simpler.

I like this proposed approach for all the reasons you mentioned here. The call uses the Go syntax and stdlib primitives to explicitly ensure idempotence and there is not implicit and obfuscated side-effect complexity associated with the initialization.

> It is guaranteed to happen once without to worry about concurrency.

This is not true. There are currently no once-and-only-once guarantees built into calls to `process.Allocate`. There is nothing stopping more than one allocation per process other than developer tradition.

internal/pkg/instrumentation/testutils/testutils.go
```diff
@@ -32,10 +32,6 @@ func ProbesLoad(t *testing.T, p TestProbe, libs map[string]*semver.Version) {
 
 	info := &process.Info{
 		ID: 1,
-		Allocation: &process.Allocation{
-			StartAddr: 140434497441792,
-			EndAddr:   140434497507328,
-		},
```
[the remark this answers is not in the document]
MrAlias: This shouldn't be zero values though. If a probe needs allocated memory it will be fully verified against allocated memory for process with ID `1`. This uses the full code-path in testing.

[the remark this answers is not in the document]
RonFed: I was thinking we need some kind of mock for these constants (since the actual allocate operation isn't performed during the test - we don't have a remote process to allocate for in the test, right ?). 
Hence the actual constants injected need to have mocks values.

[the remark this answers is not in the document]
MrAlias: Sure, as it currently is written we are relying on a process with PID `1` and a specific address space to exist and be allocate-able. I'll look into creating a fake process to mock these tests if we don't want to rely on this.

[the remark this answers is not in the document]
MrAlias: WIP: https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/1991

internal/pkg/process/info.go
```diff
@@ -118,3 +136,38 @@ func (i *Info) GetFunctionReturns(name string) ([]uint64, error) {
 
 	return nil, fmt.Errorf("could not find returns for function %s", name)
 }
+
+// onceResult is an object that will perform exactly one action if that action
+// does not error. For errors, no state is stored and subsequent attempts will
+// be tried.
```
RonFed, line 142: I'm not sure if we should re-try on an error, or maybe this should only try once regardless if the operation succeeded or not.
I can't think of a cause we could recover from an error on `allocate`

RonFed:
> This is not true. There are currently no once-and-only-once guarantees built into calls to process.Allocate. There is nothing stopping more than one allocation per process other than developer tradition.

That's a good point. My assumption is that each process has one Instrumentation and one Manager that get initialized once - hence, if the `allocate` is called in that init phase we know it happens once per process. Using the go primitives has a stronger guarantee as you said.

internal/pkg/process/info.go
MrAlias: The error would possibly come from the external system. Meaning that it could experience a temporary failure due to hardware or other externalities. I think we should follow the principle that gives the caller the opportunity to respond however they think is appropriate to the error. Not allowing them to retry means we are making that decision for them.

internal/pkg/process/info.go
```diff
@@ -152,3 +170,38 @@ func (i *Info) GetFunctionReturns(name string) ([]uint64, error) {
 
 	return nil, fmt.Errorf("could not find returns for function %s", name)
 }
+
+// onceResult is an object that will perform exactly one action if that action
```
pellared, line 174: nit: maybe rename it to something like `eventualResult` just to indicate that it can actually run more than once?

internal/pkg/process/info.go
```diff
@@ -152,3 +170,38 @@ func (i *Info) GetFunctionReturns(name string) ([]uint64, error) {
...
+// does not error. For errors, no state is stored and subsequent attempts will
+// be tried.
+type onceResult[T any] struct {
```
pellared, line 177: I would not make it as a generic if we need it only for allocations.
It is a lessons learned from C# that when generics where added then everyone was adding it to places like here where it was not necessary.
With making it for allocations only we already show where it is needed + makes the code a little simpler for those not familiar with generics.

This could also help us with coming up with a simpler name. E.g. `eventualResult[Allocation]` is more complex than `eventualAlloc`.

internal/pkg/process/info.go
```diff
@@ -152,3 +170,38 @@ func (i *Info) GetFunctionReturns(name string) ([]uint64, error) {
...
+	done atomic.Bool
+	mu   sync.Mutex
+	val  T
+}
+
+// Do runs f only once, and only stores the result if f returns a nil error.
+// Subsequent calls to Do will return the stored value or they will re-attempt
+// to run f and store the result if an error had been returned.
+func (o *onceResult[T]) Do(f func() (T, error)) (T, error) {
```
pellared, line 186: I think it would be good to add unit tests for this.

pellared (commented):
Overall LGTM.

internal/pkg/process/info.go
MrAlias: Based on https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/1945#discussion_r2006130255, given this is going to only be used for a single type, I'm going to rename to `allocOnce`.

internal/pkg/process/info.go
MrAlias: This can be flattened into the type if we don't need it to be generic. Updating.

MrAlias: Updated. PTAL.

internal/pkg/process/info.go
MrAlias: Updated. PTAL.

internal/pkg/instrumentation/manager_load_test.go
```diff
@@ -48,16 +48,11 @@ func TestLoadProbes(t *testing.T) {
 	// Reset Info module information.
 	info.Modules = make(map[string]*semver.Version)
 
-	logger := slog.Default()
-	info.Allocation, err = process.Allocate(logger, pid)
-	if err != nil {
-		t.Fatalf("failed to allocate for process %d: %v", id, err)
-	}
```
RonFed, line 55: nit: can we add an assert that the allocation field in the `info` is set as expected? we can do that in `ProbesLoad`

RonFed (approved):

internal/pkg/instrumentation/manager_load_test.go
MrAlias: I've asserted after all the probes have loaded: https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/1945/commits/232bfee3bd3903716b6621edb13dd4f148bb7372.

It is not guaranteed the first probe allocate memory so it is not done within `ProbesLoad`.

PR #1962, opened 2025-03-07
Add a dockerignore

MrAlias:
Stop copying the whole project when building the auto docker image. This causes unrelated changes to the parts of the codebase to invalidate the docker cache

This also stops using the `make` utility to build the project. Doing so brings in many steps outside of building the target binary (i.e. linting and running `go mod tidy`). Instead, run the go commands that are needed directly.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,9 +17,15 @@ FROM --platform=$BUILDPLATFORM base AS builder
 COPY . .
 
 ARG TARGETARCH
+ARG CGO_ENABLED=0
+ARG BPF2GO_CFLAGS="-I/usr/src/go.opentelemetry.io/auto/internal/include/libbpf -I/usr/src/go.opentelemetry.io/auto/internal/include"
 RUN --mount=type=cache,target=/root/.cache/go-build \
     --mount=type=cache,target=/go/pkg \
-    GOARCH=$TARGETARCH make build
+    GOARCH=$TARGETARCH \
+	CGO_ENABLED=$CGO_ENABLED \
+	BPF2GO_CFLAGS=$BPF2GO_CFLAGS \
+	go generate ./... \
+	&& go build -o otel-go-instrumentation ./cli/...
 
 FROM gcr.io/distroless/base-debian12@sha256:125eb09bbd8e818da4f9eac0dfc373892ca75bec4630aa642d315ecf35c1afb7
 COPY --from=builder /usr/src/go.opentelemetry.io/auto/otel-go-instrumentation /
--- a/Dockerfile.dockerignore
+++ b/Dockerfile.dockerignore
@@ -1,0 +1,13 @@
+*
+
+# must-have toplevel files
+!/go.sum
+!/go.mod
+!*.go
+!/LICENSE
+
+# directories
+!/cli
+!/sdk
+!/internal/include
+!/internal/pkg


Dockerfile
```diff
@@ -17,9 +17,15 @@ FROM --platform=$BUILDPLATFORM base AS builder
...
+ARG CGO_ENABLED=0
+ARG BPF2GO_CFLAGS="-I/usr/src/go.opentelemetry.io/auto/internal/include/libbpf -I/usr/src/go.opentelemetry.io/auto/internal/include"
 RUN --mount=type=cache,target=/root/.cache/go-build \
     --mount=type=cache,target=/go/pkg \
-    GOARCH=$TARGETARCH make build
+    GOARCH=$TARGETARCH \
+	CGO_ENABLED=$CGO_ENABLED \
+	BPF2GO_CFLAGS=$BPF2GO_CFLAGS \
+	go generate ./... \
+	&& go build -o otel-go-instrumentation ./cli/...
```
RonFed, line 28: I think I prefer calling the makefile here,
The main reason being we don't need to duplicate the env vars setup and having a single source of truth for building.
If we want to avoid go-mod tidy and linting we can probably add a target without those dependencies?
Although I don't think it adds a lot of latency to the build.

MrAlias: I prefer we avoid the circular dependency of make calling this and then this calling make.

damemi: I think I agree with @RonFed, calling `make build` reduces the need to make sure changes are duplicated

RonFed: > Additionally, it is adding the command to a development tooling file, but it is not meant to be run by developers. What happens when we get an issue asking why the new make target fails because their repo was not correctly linted before-hand? That seems like a valid complaint. They are running developer tooling, its just they don't have the domain history of the project to know that this "one special target" isn't meant for them to run.

This can be addressed by improving the Makefile.
For example when the build target is triggered, it can check if it is running on linux, otherwise return an informative error.

MrAlias: @RonFed I don't follow. I develop on a >inux system, why would I want it to succeed for me but not for a Mac dev?

Also, that wouldn't address the docker cache invalidation I am trying to address.

RonFed: I'm working on a Mac so I use `make docker-build`, If I'd use `make build` we can have an error like `building the executable is only supported on Linux system, try make docker-build`.

Regarding the cache invalidation - the Makefile doesn't seems to change a lot I think.

MrAlias: > Regarding the cache invalidation - the Makefile doesn't seems to change a lot I think.

I guess that is relative. It changes all the time for me working on things like the testing pipeline and fixing the Makefile.

MrAlias: > I'm working on a Mac so I use `make docker-build`, If I'd use `make build` we can have an error like `building the executable is only supported on Linux system, try make docker-build`.

Gotcha. I'm still not in favor of adding a target to the Makefile that will not work on any system and isn't intended for any developer to use.

MrAlias: Please see https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/2031 which adds a make target to run the project with a docker container. The container has the full project mounted there and supports running make commands.

RonFed (approved):

PR #1971, opened 2025-03-11
auto/sdk: use stack allocated span context passed to eBPF

### LINKED ISSUE #1961: auto sdk: avoid passing heap allocated variables to ebpf program
Doing some tests on my setup I consistently encountered an error
```
             app-658961  [002] ....1 151003.148202: bpf_trace_printk: write_span_context: failed to write trace ID: -14
             app-658961  [002] ....1 151003.148204: bpf_trace_printk: failed to write span context: -2
```

which means the memory we tried to write to was not paged in yet (see https://github.com/torvalds/linux/blob/00a7d39898c8010bfd5ff62af31ca5db34421b38/mm/maccess.c#L150).

since we fully control the pointers we provide to that program, we could make sure that these pointers are paged in and are on the stack using the following change:

```Diff
diff --git a/sdk/tracer.go b/sdk/tracer.go
index 131b3ac8..d147e412 100644
--- a/sdk/tracer.go
+++ b/sdk/tracer.go
@@ -23,14 +23,15 @@ type tracer struct {
 var _ trace.Tracer = tracer{}
 
 func (t tracer) Start(ctx context.Context, name string, opts ...trace.SpanStartOption) (context.Context, trace.Span) {
-       var psc trace.SpanContext
+       var psc, sc trace.SpanContext
        sampled := true
        span := new(span)
 
        // Ask eBPF for sampling decision and span context info.
-       t.start(ctx, span, &psc, &sampled, &span.spanContext)
+       t.start(ctx, span, &psc, &sampled, &sc)
 
        span.sampled.Store(sampled)
+       span.spanContext = sc
 
        ctx = trace.ContextWithSpan(ctx, span)
```

Tested this and verifies it solves the issue.
@MrAlias

RonFed:
resolves #1961


MrAlias (approved):
We need to copy this to opentelemetry-go to resolve the issue still.

RonFed:
> We need to copy this to opentelemetry-go to resolve the issue still.

@MrAlias see https://github.com/open-telemetry/opentelemetry-go/pull/6434

PR #1985, opened 2025-03-13
Use exact Go version from debug for `process.runtime.version`

### LINKED ISSUE #1984: `process.runtime.version` incorrect
From the [semantic conventions](https://github.com/open-telemetry/semantic-conventions/blob/ac2f0778479bd0d46daa0a6da17447f86ba7cc60/docs/resource/process.md#go-runtimes):

> `process.runtime.version` - Fill in the exact value returned by `runtime.Version()`, i.e. `go1.17`.

Meaning this needs to be removed:

https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/b9a7efe80f7d539656b7cbdfa5f55798e6007583/internal/pkg/process/id.go#L91-L95

MrAlias:
Fix #1984


internal/pkg/process/info.go
```diff
@@ -35,30 +42,57 @@ func NewInfo(id ID, relevantFuncs map[string]interface{}) (*Info, error) {
...
-	result.Modules, err = findModules(goVersion, bi.Deps)
+	result.Modules, err = findModules(result.GoVersion, bi.Deps)
 	return result, err
 }
 
+func goVer(raw string) (*semver.Version, error) {
+	if strings.HasPrefix(raw, "devel") {
+		return goDevVer(raw)
+	}
+	raw = strings.TrimPrefix(raw, "go")
```
RonFed, line 70: I'm a bit confused, if we remove the `go` prefix, how does it end up in the trace results?

MrAlias: This is parsing the version from `runtime.Version` to be used as a `semver.Version`. This is used for the module map and `GoVersion` field. It is not used by the types that generate resources. The `BuildInfo` is used directly for by those, which contains the exact version of Go.

RonFed: I see, thanks.
Do you think we can do the `buildinfo.ReadFile` once and use the result both for the resource and for the `semver`? not a blocker just think its nice to have since I assume that function has some non negotiable cost

RonFed (approved):

internal/pkg/process/info.go
MrAlias: I think we could change to doing that, but we would need to change the underlying `ID` type to not be an `int` but a `struct`.

I'm not sure it is needed given these are called during startup. If a caller wants to cache this they can do so. But, it is something we can consider going forward. Especially when we start moving the `process` package externally.

PR #1965, opened 2025-03-09
Run e2e tests on arm

RonFed:

PR #1986, opened 2025-03-13
Stop adding `process.runtime.description` to resource

### LINKED ISSUE #1983: `process.runtime.description` should not be set
This attribute is not recommended for the Go runtime: https://github.com/open-telemetry/semantic-conventions/blob/ac2f0778479bd0d46daa0a6da17447f86ba7cc60/docs/resource/process.md#go-runtimes

MrAlias:
Fix #1983


RonFed (approved):
we can also remove the redact addition from #1965 since we won'y have the arch in the resource

PR #1991, opened 2025-03-14
Test probe load with mocks

MrAlias:
- Use a real compiled Go binary running process to simulate real allocation and process identification behavior
- Remove the build flag so tooling finds the test file during normal development (skip the test if not able to run)
- Flatten the `testutils` package into the only place it is used


RonFed (approved):
🥇

PR #2001, opened 2025-03-17
Instrument the OTel trace API's auto SDK

MrAlias:
Part of #974 

Depends on https://github.com/open-telemetry/opentelemetry-go/pull/6456.


internal/test/e2e/nonrecording/main.go
RonFed, line 1: This looks very similar to the `autosdk` I think we should consider to unify them to one test where we could check both functionalities.

instrumentation.go
```diff
@@ -87,6 +88,7 @@ func NewInstrumentation(ctx context.Context, opts ...InstrumentationOption) (*In
 		kafkaProducer.New(c.logger, Version()),
 		kafkaConsumer.New(c.logger, Version()),
 		autosdk.New(c.logger),
+		otelTrace.New(c.logger),
```
RonFed, line 91: Don't we want this to be opt-in same as we have below the check for 
```Go
if c.globalImpl {
		p = append(p, otelTraceGlobal.New(c.logger))
	}
```
I think it should be the same opt-in flag - or do we want to remove that flag entirely and have this on by default (both the global and the non-recording)

internal/pkg/instrumentation/bpf/go.opentelemetry.io/otel/trace/bpf/probe.bpf.c
```diff
@@ -0,0 +1,235 @@
...
+
+// This instrumentation attaches uprobe to the following function:
+// func (noopSpan) tracerProvider(autoEnabled *bool) TracerProvider
+// https://github.com/open-telemetry/opentelemetry-go/blob/2e8d5a99340b1e11ca6b19bcdfcbfe9cd0c2c385/trace/noop.go#L98C1-L98C65
+SEC("uprobe/tracerProvider")
+int uprobe_tracerProvider(struct pt_regs *ctx) {
+    if (wrote_flag) {
+        // Already wrote flag value.
+        return 0;
+    }
```
RonFed, line 95: This is kind of an invalid state - if we already wrote we should have been unloaded I guess.
It also can take some time for the unload to happen so this covers this, right?

internal/pkg/instrumentation/bpf/go.opentelemetry.io/otel/trace/bpf/probe.bpf.c
```diff
@@ -0,0 +1,235 @@
...
+// TODO: tune this. This is a just a guess, but we should be able to determine
+// the maximum size of a span (based on limits) and set this. Ideally, we could
+// also look into a tiered allocation strategy so we do not over allocate
+// space (i.e. small, medium, large data sizes).
+#define MAX_SIZE 2048
+
+// Injected const.
+volatile const u64 span_context_trace_id_pos;
+volatile const u64 span_context_span_id_pos;
+volatile const u64 span_context_trace_flags_pos;
```
RonFed, line 23: we could share this in a header with auto/sdk probe that uses the exact same consts.

RonFed (commented):
overall, looks great.
wonder if we can share a lot of this between the auto/sdk probe and this new one in:
* C code - we can have a common header with the common functionality.
* `probe.go` also has common logic of parsing the perf events.
* e2e tests can be unified.

internal/test/e2e/nonrecording/main.go
MrAlias: They are similar code paths, but different origins that test different code. I would prefer to not combine testing of two different SDKs into the same test.

internal/pkg/instrumentation/bpf/go.opentelemetry.io/otel/trace/bpf/probe.bpf.c
MrAlias: I'm fine if you would like to refactor this after this PR is merged. I would prefer to keep the complexity of this PR to a minimum and making changes to other probes to support a new header is not something I want to include here.

instrumentation.go
MrAlias: I think we can remove the flag. It was originally intended to set user expectations about our initial partial support for the OTel tracing API. At this point we provide full coverage of the API (for recent versions of otel-go).

I can track the removal of this opt-in flag in an issue an tackle it in a follow-on PR.

internal/pkg/instrumentation/bpf/go.opentelemetry.io/otel/trace/bpf/probe.bpf.c
MrAlias: FWIW, this is how the `auto/sdk` probe also handles this.

internal/pkg/instrumentation/bpf/go.opentelemetry.io/otel/trace/bpf/probe.bpf.c
MrAlias: SIG meeting feedback. Refactor this PR to handle dedup.

internal/test/e2e/nonrecording/main.go
MrAlias: Resoling based on SIG meeting agreement.

instrumentation.go
MrAlias: Resolved by #1859.

internal/pkg/instrumentation/bpf/go.opentelemetry.io/otel/trace/bpf/probe.bpf.c
MrAlias: Updated. PTAL.

internal/include/sdk.h
```diff
@@ -0,0 +1,36 @@
...
+
+struct {
+    __uint(type, BPF_MAP_TYPE_HASH);
+    __type(key, void*);
+    __type(value, struct otel_span_t);
+    __uint(max_entries, MAX_CONCURRENT);
+} active_spans_by_span_ptr SEC(".maps");
+
+// Function declaration
+static __always_inline long write_span_context(void *go_sc, struct span_context *sc);
```
RonFed, line 34: I think we can put the function implementation here too. As it looks identical to me in both cases

RonFed (approved):

PR #2028, opened 2025-03-20
Fix demo sql syntax

minimAluminiumalism:
In demo`httpPlusdb`, the definition of `tableInsertion` is:

```
tableInsertion = `INSERT INTO 'contacts'
	('first_name', 'last_name', 'email', 'phone') VALUES
	('Moshe', 'Levi', 'moshe@gmail.com', '052-1234567');`
```

SQL typically uses backticks (`) or double quotes (") for identifiers (table/column names), not single quotes which are for string values, which will cause the sqlparser to fail parsing.

截屏2025-03-20 11 49 06

The original variable definition prevents us from setting `db.query.summary` as the span name(Originally using default field `DB`).

after:
Clipboard_Screenshot_1742460781

### DIFF
--- a/examples/httpPlusdb/main.go
+++ b/examples/httpPlusdb/main.go
@@ -24,9 +24,9 @@ const (
 	email TEXT NOT NULL,
 	phone TEXT NOT NULL);`
 
-	tableInsertion = `INSERT INTO 'contacts'
-	('first_name', 'last_name', 'email', 'phone') VALUES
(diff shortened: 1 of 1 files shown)


minimAluminiumalism:
I suppose this change don't need to be added to changelog.md, and it seems hat the failed k8s e2e-tests is unrelated to this PR, rerun the check? @MrAlias

MrAlias (approved):

PR #2029, opened 2025-03-20
(wip) shift process management from Probe to Manager

damemi:
Part of setting up for exposing the Probe API -- the main goal of this PR is to shift usage of the internal `process` package from the base Probe implementation to the Manager.

https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/1943 moved process analysis to the manager, so this follows a similar pattern of consolidating process handling into the manager.

Functionally, that means moving a lot of the Probe initialization to the manager. So now, instead of thinking "the Probe sets itself up" the flow is more "the Manager sets up the Probe"

This is a WIP right now showing the basic idea, the tougher part will be decoupling Process info from Consts


internal/pkg/instrumentation/probe/probe.go
```diff
@@ -36,17 +33,36 @@ type Probe interface {
...
+	Init(*sampling.Config) error
 
 	// Run runs the events processing loop.
 	Run(func(ptrace.ScopeSpans))
 
 	// Close stops the Probe.
 	Close() error
+
+	// GetLogger returns the *slog.Logger associated with the Probe.
+	GetLogger() *slog.Logger
```
MrAlias, line 46: Why would we ever need to get the logger from a probe? Doesn't the Manager have its own logger it can use for its own operations?

internal/pkg/instrumentation/probe/probe.go
```diff
@@ -36,17 +33,36 @@ type Probe interface {
...
+
+	// Spec returns the *ebpf.CollectionSpec for the Probe.
+	Spec() (*ebpf.CollectionSpec, error)
+
+	// SetCollection sets the *ebpf.Collection for the Probe.
+	SetCollection(*ebpf.Collection)
+
+	// GetCollection returns the *ebpf.Collection for the Probe.
+	GetCollection() *ebpf.Collection
```
MrAlias, line 55: This has a smell to it. Information and processing should flow in one direction. Why would we need to get and set this value. Can it be avoided?

internal/pkg/instrumentation/probe/probe.go
```diff
@@ -36,17 +33,36 @@ type Probe interface {
...
+
+	// GetUprobes returns a list of *Uprobes for the Probe.
+	GetUprobes() []*Uprobe
+
+	// GetConsts returns a list of Consts for the Probe.
+	GetConsts() []Const
```
MrAlias, line 61: These are duplicates of the `Manifest` method which contains these values.

internal/pkg/instrumentation/probe/probe.go
```diff
@@ -36,17 +33,36 @@ type Probe interface {
...
+
+	// UpdateClosers updates the closers for the Probe to the io.Closers passed to it,
+	// and returns the new list of io.Closers for the Probe.
+	UpdateClosers(...io.Closer) []io.Closer
```
MrAlias, line 65: This as well has a smell. It implies that we are externalizing an operation inherently related to a probe and then providing a setter to save the results.

If an operation is going to update the state of a probe that operation should be encapsulated within the probe itself.

MrAlias (commented):
I like the idea of decoupling the probe from process management, but it seems like this is just shuffling existing patterns out of a probe without refactoring the details deep in the functionality. I think we need to drill deeper into what the process interactions are for a probe and likely redesign around that.

internal/pkg/instrumentation/probe/probe.go
damemi: after the sig call, I am moving the collection tracking to the manager, so these aren't necessary for the probe anymore

internal/pkg/instrumentation/probe/probe.go
damemi: Moved closers to the manager as well, this is no longer needed

internal/pkg/instrumentation/probe/probe.go
damemi: True, and it wasn't clear to me because we change the name of them in the Manifest: `Uprobes`->`FunctionSymbol` and `Consts`->`StructFieldIDs`

The latter I think is because the Manifest doesn't actually store all the Consts in the Probe, just the consts for fields. If we're loading the uprobes in the manager now and injecting the consts there, we'll need all of the consts.

Do you think we should refactor the Manifest struct a bit to align better with the Probe definition?

damemi:
@MrAlias I did some more refactoring based on the last sig call. Basically, the Manager will now track a reference to the "initialized" probe, ie the probe+its collection+its closers. Then all collection/closer interaction can be done in the manager.

It's just a step, but let me know what you think

damemi:
@MrAlias and now that I think about it, the collection/closer stuff could be set in the Manifest, but is the manifest intended to be mutable like that (ie, updated once the Probe is initialized) or just a static representation of the Probe's defintion?

MrAlias:
> just a static representation of the Probe's defintion?

My intent was for it to be static.

internal/pkg/instrumentation/probe/probe.go
MrAlias: We should definitely unify these types respectively.

Given they are static to a probe, keeping their definitions in the `Manifest` seems appropriate.

internal/pkg/instrumentation/manager.go
```diff
@@ -398,11 +419,189 @@ func (m *Manager) loadProbes() error {
...
+			logFn("failed to load uprobe", "probe", i.probe.Manifest().ID, "symbol", up.Sym, "error", err)
+			continue
+		}
+
+		i.closers = append(i.closers, up)
+	}
+	return nil
+}
+
+func (m *Manager) loadUprobe(u *probe.Uprobe, c *ebpf.Collection) error {
```
MrAlias, line 531: It looks like this is no longer used and a can be removed.

internal/pkg/instrumentation/manager.go
```diff
@@ -398,11 +419,189 @@ func (m *Manager) loadProbes() error {
...
+		}
+	}
+	if err != nil {
+		return err
+	}
+
+	return inject.Constants(spec, opts...)
+}
+
+func (m *Manager) loadUprobesFromProbe(i ProbeReference) error {
```
MrAlias, line 471: Similar here, this can be removed.

internal/pkg/instrumentation/manager.go
```diff
@@ -398,11 +419,189 @@ func (m *Manager) loadProbes() error {
...
+	if err != nil {
+		return errors.Join(err, m.cleanup())
+	}
+
+	err = m.InjectProbeConsts(i.probe, spec)
+	if err != nil {
+		return errors.Join(err, m.cleanup())
+	}
+
+	c, err := utils.InitializeEBPFCollection(spec, m.collectionOpts)
```
MrAlias, line 435: Should this function be moved to this package/file. This is the only place it is used.

internal/pkg/instrumentation/manager.go
```diff
@@ -398,11 +419,189 @@ func (m *Manager) loadProbes() error {
...
+	reader, err := i.probe.InitStartupConfig(c, cfg.SamplingConfig)
+	if err != nil {
+		return errors.Join(err, m.cleanup())
+	}
+	i.closers = append(i.closers, reader)
+
+	return nil
+}
+
+func (m *Manager) InjectProbeConsts(i probe.Probe, spec *ebpf.CollectionSpec) error {
```
MrAlias, line 450: Should this be exported? It should only ever be called by the manager and never external to this package, right?

internal/pkg/instrumentation/manager.go
```diff
@@ -54,6 +58,15 @@ type Manager struct {
 	probeMu         sync.Mutex
 	state           managerState
 	stateMu         sync.RWMutex
+	collectionOpts  *ebpf.CollectionOptions
+}
+
+// ProbeReference is used by the Manager to track an initialized reference
+// to a Probe and its related resources such as its ebpf.Collection and io.Closers.
+type ProbeReference struct {
```
MrAlias, line 66: This can be unexported. It is only used internal to this package.

internal/pkg/instrumentation/manager.go
```diff
@@ -54,6 +58,15 @@ type Manager struct {
 	probeMu         sync.Mutex
 	state           managerState
 	stateMu         sync.RWMutex
+	collectionOpts  *ebpf.CollectionOptions
```
MrAlias, line 61: This doesn't seem to have a need as a field. It is only ever used in a single call site and always set to the same value. It should just be a local var to https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/2029/files#diff-8008933194d32417cc135227b15916c4d1f575de0abda31bc0545b51d9783bebR435

internal/pkg/instrumentation/manager.go
```diff
@@ -128,7 +147,10 @@ func (m *Manager) registerProbe(p probe.Probe) error {
 		return err
 	}
 
-	m.probes[id] = p
+	m.probes[id] = ProbeReference{
+		probe:   p,
+		closers: make([]io.Closer, 0),
```
MrAlias, line 152: This is not needed. It is only allocating memory prior to need. `append` handles appends to empty values without issue.

```suggestion
```

internal/pkg/instrumentation/manager.go
```diff
@@ -398,11 +419,189 @@ func (m *Manager) loadProbes() error {
...
+	err = m.InjectProbeConsts(i.probe, spec)
+	if err != nil {
+		return errors.Join(err, m.cleanup())
+	}
+
+	c, err := utils.InitializeEBPFCollection(spec, m.collectionOpts)
+	if err != nil {
+		return errors.Join(err, m.cleanup())
+	}
+	i.collection = c
```
MrAlias, line 439: This is an unused write to `i.collection`

internal/pkg/instrumentation/manager.go
```diff
@@ -398,11 +419,189 @@ func (m *Manager) loadProbes() error {
 	return nil
 }
 
+func (m *Manager) LoadProbe(i ProbeReference, name probe.ID, cfg Config) error {
```
MrAlias, line 422: `LoadProbe` is only ever called internal to the package. This should be unexported.

internal/pkg/instrumentation/manager.go
MrAlias: It's also not clear why this is needed. The collection field is never used (it is ineffectively set, but never read), and the `closers` field could just be captured in a thin wrapper around a `Probe` implementation (given it is already an interface).

MrAlias (commented):
This doesn't looks to be a working example. I'm not sure how far I can review because of this.

internal/pkg/instrumentation/manager.go
MrAlias: Oh, I gotcha, just this is writing to a value, not a reference. When this function returns the passed argument remains unchanged. This assignment is to a copied value that is not used.

internal/pkg/instrumentation/probe/probe.go
MrAlias: Yes. Any logger that the manager passes should just be the manager's logger.

damemi: sgtm, I wasn't totally sure what this relationship was between them

internal/pkg/instrumentation/manager.go
damemi: this was a mistake on my part, it should be a pointer reference so we can access it when closing as Ron pointed out

internal/pkg/instrumentation/probe/probe.go
damemi: @MrAlias looking more at the Manifest, I think this change in particular makes it redundant and everything it provides can be in the Probe itself. Given that we are defining the Probe as a more static representation now, I pushed an update to remove the Manifest type entirely and just use the references from it directly in the Probe.

Let me know if there's some other reason for Manifest I'm missing

internal/pkg/instrumentation/manager.go
damemi: replied above that this should have been set on a pointer, sorry about that confusion. But I'm curious what you mean by the closers field being a wrapper around a Probe. If I understand you correctly, that sounds like what I'm basically trying to do here, but could you explain what you have in mind? If it simplifies this then awesome

damemi:
Hi all, thanks for the feedback so far and for your patience on my end getting back to you.

I pushed some more updates, notably:

* Unexported locally used types and methods in the manager
* Made sure probeReference fields are set by reference (the new Go testing framework actually helped catch this, thanks @MrAlias for that work)
* Collapsed the `Manifest` field into `Probe`. It seems like Manifest is mostly redundant at this point when we can get what that provided from the Probe itself (ID, uprobes, and struct fields). Downside is this adds methods to the Probe type, but I think they make sense (GetUprobes, GetConsts, GetID). 

For the consts, I made a type that has a method to just get the StructField consts to preserve the current behavior of Manifest. This looks like it's only used in a test though, so if we don't need it then I'm find dropping it.

This now passes the tests, but I'm sure there's more we can do so please take another look. I squashed the commits down to give us a clean slate at this point. Thanks!

internal/pkg/instrumentation/manager.go
MrAlias: We should avoid programming side effects. Instead, we should use standard Go programming practices and have any values we want to use from a function be returned.

.gitignore
```diff
@@ -47,3 +47,5 @@ examples/kafka-go/producer/producer
 examples/kafka-go/consumer/consumer
 
 go.work*
+
+opentelemetry-helm-charts
```
MrAlias, line 51: This is not needed. 

```suggestion
```

internal/pkg/instrumentation/probe/probe.go
```diff
@@ -17,39 +17,53 @@ import (
...
+func (id ID) String() string {
+	return fmt.Sprintf("%s/%s", id.InstrumentedPkg, id.SpanKind)
+}
+
 // Probe is the instrument used by instrumentation for a Go package to measure
 // and report on the state of that packages operation.
 type Probe interface {
-	// Manifest returns the Probe's instrumentation Manifest. This includes all
-	// the information about the package the Probe instruments.
-	Manifest() Manifest
```
MrAlias, line 40: Replacing this with individual mentions does not seem advantageous. 

Methods cannot be added to an interface in a backwards compatible manner. Fields and methods can be added to a strict in a backwards compatible manner.

There is also no way for default behavior for an interface other than compile time failures. A struct can be designed to apply sane defaults are expressive errors. 

I feel we should continue to hold static definitions about the probe in a manifest type of some sort instead of this flattening. Methods for the probe should be added for actions the probe makes and other components will can to initiate those actions.

damemi:
It seems unnecessary to have a separate struct nested in the Probe
definition whose entire purpose is just to be a carbon copy of existing
fields in the Probe. We know that every Probe is going to need an ID,
Uprobes, and Consts. So it’s reasonable to add those as basic interface
getters on the API. The Base Probe implementation can provide the default
behavior like it does here. I don’t see the point in a duplicate Manifest
type that has to be instantiated on each Probe anyway.

internal/pkg/instrumentation/manager.go
damemi: Can you please elaborate what you mean?

internal/pkg/instrumentation/probe/probe.go
damemi: @MrAlias I re-added the `Manifest()` method to the Probe interface after our discussion on the sig call. I tried taking this a little step further and now the Manifest just stores the raw `Consts` and `Uprobes`, with methods for getting the `StructFields` and `Symbols` specifically from that. We had talked before about trying to align these terms, and they do seem to have separate usages, so I'm curious what you think

Right now we precompute and sort these fields, which I could see being an advantage to not have to sort them every time they are called. So, maybe we store consts, uprobes, structfields, *and* symbols in the Manifest. The tradeoff there is just the redundancy. So either way could work imo

internal/pkg/instrumentation/manager.go
damemi: After the SIG discussion, I went back and refactored this to do less implicit updating of the passed argument from the caller, and instead I'm returning values like the `collection` and `closers` which are then set by the caller (the Manager) on the ProbeReference for the Manager to track

Doing this, I actually broke `loadProbe` up into more discrete functions based on what we are returning and setting. So previously, `loadProbe` did the following:

* initialize the collection spec
* load the uprobes (using the collection spec)
* set closers on the probe
* pass sampling config to the probe

These were all [done in one function](https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/1375c4779c060ba41029f63a47e08dbf61404905/internal/pkg/instrumentation/probe/probe.go#L121-L159) even before this shift to doing it in the Manager. Now that the Manager is an external caller, this means that output from the individual steps (like the collection) need to be handled and passed to the next step. So that's the reason for those changes, let me know what you think

damemi:
From last week's sig call, just summarizing the 2 changes I've made here:

* https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/2029#discussion_r2109417516 - re-add Manifest to the Probe interface, trying out the idea of Manifest storing raw consts/uprobes and providing functions to get structfields/symbols from that
* https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/2029#discussion_r2109428008 - make reference updating more explicit, involves breaking the current `loadProbe` into more discrete methods

rafaelroquetto:
After digging into this with @damemi, I think this change makes a lot of sense. It cleanly decouples the actual business logic by moving it into the manager, and turns the probe into a simple declarative shell - which feels like the right direction.

Encapsulating the machinery inside the manager makes it all internal detail, which not only simplifies the interface but also sets us up nicely for future extensions without breaking anything. I also appreciate how it improves clarity and maintainability - you no longer need to cross-reference the probe and manager code. It’s now self-contained within the manager, behind a clean, well-defined interface.

I also understand keeping the initial scope focused on uprobes, which is a solid starting point. This design gives us a clear path to revisit and expand once this lands.

On the "Manifest" type: my first instinct was that it might make more sense for things to either live in the probe directly, or in the manifest alone - with the probe simply returning the manifest. But then, that risks turning the manifest into a full-blown probe itself, which defeats the purpose. Mike explained the reasoning behind keeping the Manifest, and it makes sense to me, but in a way it would be good to have a single entity (although I understand this may not be possible due to API guarantees) - so consider this more of an open-ended thought than an objection, 

Finally, just a few things to keep in mind for future iterations beyond this PR - especially if we aim to integrate this more broadly across other OpenTelemetry eBPF projects like OBI. The future design should ideally accommodate:

- Support for different types of eBPF programs beyond uprobes. Ultimately, making this "program-type agnostic" would be ideal - but we should be cautious not to end up building a thin wrapper around cilium/ebpf. There's a balance to strike here, and happy to chat more when the time comes.
- Removing the one-ring-buffer-per-probe constraint. In OBI, multiple probes (or "tracers") share ring buffers.
- Providing mechanisms for features like BEYLA_PIN_INTERNAL maps. These are pinned maps that bypass the bpffs filesystem and are shared between probes using file descriptors. For reference: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/6fbf834be87fabd1de051d69bb65f1d976cc08f4/pkg/components/ebpf/tracer_linux.go#L69
- Allowing probes to be reused, i.e. having the ability of sharing a probe amongst N executables, rather than instantiating a probe per executable.

I think that sums it up, but @damemi can provide further context on the next SIG meeting (public holiday here, so I will be away).

Overall, really good stuff IMHO!

damemi:
Thanks @rafaelroquetto! Yeah, this PR is entirely meant to adjust the internal machinery -- not finalize any API details. By doing that I think it makes extending to support features like those you listed much easier, because the bulk of the change for those will be in the Manager.

Overall I'd like to keep the scope of this PR narrow, and just focused on moving the machinery to start decoupling the API from internal packages. There will be more PRs like this one, but I would rather do it in steps than try to get everything perfect in one giant PR. Like we discussed, I think the Manifest changes here are even a bit scope creepy for what I was trying to do (but I'm fine with that either way since it is related).

MrAlias:
Closing for now. We are looking at #2500 first.

PR #2090, opened 2025-04-03
Replace net/http Request.pat.str with Request.Pattern

txabman42:
`Request.pat.str` was made public via `Request.Pattern` in go1.23
- https://github.com/golang/go/commit/a523152ea1df8d39d923ed90d19662896eff0607
- [Proposal details](https://github.com/golang/go/issues/66405)

It's just a small change to not depend for internal properties if public ones could be used. A bit more simple and avoid being affected by some internal refactors.


internal/pkg/instrumentation/bpf/net/http/server/bpf/probe.bpf.c
```diff
@@ -297,11 +296,7 @@ int uprobe_serverHandler_ServeHTTP_Returns(struct pt_regs *ctx) {
     // Collect fields from response
     read_go_string(req_ptr, method_ptr_pos, http_server_span->method, sizeof(http_server_span->method), "method from request");
     if (pattern_path_supported) {
-        void *pat_ptr = NULL;
-        bpf_probe_read(&pat_ptr, sizeof(pat_ptr), (void *)(req_ptr + req_pat_pos));
-        if (pat_ptr != NULL) {
-            read_go_string(pat_ptr, pat_str_pos, http_server_span->path_pattern, sizeof(http_server_span->path), "patterned path from Request");
-        }
+        read_go_string(req_ptr, req_pattern_pos, http_server_span->path_pattern, sizeof(http_server_span->path_pattern), "pattern from Request");
```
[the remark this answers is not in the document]
txabman42: But `Request.pat` was added in go 1.22. Before 1.22, spans based on the net/http ServerMux will just have `<method>` as the span name. After 1.22, those span names will be `<method> <pattern>`.

With this change, those span names will be `<method> <pattern>` after 1.23 instead of 1.22.

AFAIU go1.22 will be the only version affected which will generate its net/http ServerMux spans with just `<method>` name. As it's not a huge breaking change, I wasn't sure how its being handling in this project.

But all versions <1.22 wont be affected, as in both cases the span name will be just `<metho>`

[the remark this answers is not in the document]
grcevski: I see, sorry I misunderstood the scope of the change, but I still think we should ensure 1.22 works as well as possible. We currently have a way to extract the pattern for 1.22, I don't think we should lose that ability. We have a way to pass down the Go version information at BPF load time, so one possibility is to add a constant saying we are on Go > 1.22 and then we can use the public interface field, otherwise can pull it out from the private field as it's done today.

RonFed (changes requested):
I agree with @grcevski comments.
There is no reason to drop support for go1.22 - we can use the private field when the public one is not available.

internal/pkg/instrumentation/bpf/net/http/server/bpf/probe.bpf.c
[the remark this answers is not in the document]
txabman42: Agree, it makes sense. TY for the feedback, I've added the changes to support also 1.22.

internal/pkg/instrumentation/bpf/net/http/server/probe.go
```diff
@@ -232,7 +252,7 @@ func processFn(e *event) ptrace.SpanSlice {
 	}
 
 	spanName := method
-	if isPatternPathSupported && isValidPatternPath {
+	if (isPatternPathPublicSupported || isPatternPathSupported) && isValidPatternPath {
```
RonFed, line 255: Not sure this condition should change - this code is agnostic to whether the field was read from the private or public field.

internal/pkg/instrumentation/bpf/net/http/server/bpf/probe.bpf.c
```diff
@@ -296,7 +299,9 @@ int uprobe_serverHandler_ServeHTTP_Returns(struct pt_regs *ctx) {
     bpf_probe_read(&url_ptr, sizeof(url_ptr), (void *)(req_ptr + url_ptr_pos));
     // Collect fields from response
     read_go_string(req_ptr, method_ptr_pos, http_server_span->method, sizeof(http_server_span->method), "method from request");
-    if (pattern_path_supported) {
+    if (pattern_path_public_supported) {
```
RonFed, line 302: nit: I think a more clear way of writing it might be:
```
if pattern_supported {
  if public_pattern {
      // read public
   } else {
     // read private
   }
}
```

CHANGELOG.md
```diff
@@ -31,6 +31,7 @@ OpenTelemetry Go Automatic Instrumentation adheres to [Semantic Versioning](http
 ### Changed
 
 - Upgrade OpenTelemetry semantic conventions to `v1.30.0`. ([#2032](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/2032))
+- Modify how the pattern is fetch from `net/http.Request`. Now it uses `Request.Pattern` instead of `Request.pat.str` ([#2090](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/2090)) unless it's go1.22, which continue using `Request.pat.str`.
```
MrAlias, line 34:
```suggestion
- Modify how the pattern is fetch from `net/http.Request`.
  Now it uses `Request.Pattern` instead of `Request.pat.str` unless using go1.22, which continue using `Request.pat.str`. ([#2090](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/2090))
```

internal/pkg/instrumentation/bpf/net/http/server/bpf/probe.bpf.c
txabman42: Agree, updated

internal/pkg/instrumentation/bpf/net/http/server/probe.go
txabman42: That's a good catch. No need to add `isPatternPathPublicSupported` as in the end is a subgroup of `isPatternPathSupported`. If the pattern path is supported, it will be added to the span, regardless its private or public.

RonFed (approved):

grcevski (approved):
LGTM! Thanks for iterating on this!

PR #2099, opened 2025-04-05
[WIP] - add go version to e2e test matrix

RonFed:


MrAlias:
The instability of our CI system needs to be addressed before we add additional test cases. The system is already very unstable and this will only make things worse.

PR #2105, opened 2025-04-06
fix: fallback to no context propagation in kafka-go instrumentation if `bpf_probe_write_user` is not enabled

RonFed:


grcevski (approved):
LGTM!

rafaelroquetto:
More like a suggestion than anything. Instead of directly relying on `#ifndef NO_HEADER_PROPAGATION`, it is better to do something like:
```
#ifdef NO_HEADER_PROPAGATION
enum { k_header_propagation_enabled = 0 };
#else
enum { k_header_propagation_enabled = 1 };
```

and then
```
if (k_header_propagation_enabled) { ... }
```

in the code. Using a constant instead of a macro ensures that the code is always sieved thru the compiler regardless of what is being defined/not defined, and helps errors to be caught invariably.

When `k_header_propagation_enabled == 0`, the compiler will strip out the dead code anyway, so in that sense it's pretty much equivalent.

RonFed:
@rafaelroquetto that's a nice trick.
I prefer the `#ifdef` because it allows to remove un needed definitions (e.g a map only used for context propagation).

PR #2118, opened 2025-04-09
Run integration tests with Go

### LINKED ISSUE #93: Refine e2e tests

MrAlias:
# Replace Integration Tests with Go `testing` implementation 

This replaces the existing integration test run on a `kind` server with ones run directly from Go on the host machine.

Resolves #93 

## Add `instrumentation_integration_test.go`

This is the integration testing file. It runs integration tests for all existing coverage:

- autosdk
- databasesql
- gin
- grpc
- kafka-go
- nethttp
- nethttp_custom
- otelglobal

Each test is run as a sub-test of the main `TestIntegration`. This allows users to run all test:

```go
go test -run=TestIntegration
```

Or just a single test for a packge. For example:

```go
go test -run=TestIntegration/KafkaGo
```

The test is skipped if being run in short mode (i.e. `go test -short`), or if the user does not have permissions to set memory limits on the host (the same as our eBPF verifier tests).

The test can be run from within a virtualized environment if permissions are not available by default with something like the following:

```shell
docker run --rm -it --privileged --network=host --user root -v /var/run/docker.sock:/var/run/docker.sock -v "$PWD":/usr/src/go.opentelemetry.io/auto -w /usr/src/go.opentelemetry.io/auto golang:1.24.1 sh -c "apt update; apt install docker.io bats jq -y; go test -v -run=TestIntegration"
```

Note, the `Makefile` has been updated (see below).

### Processing reuse where possible

Instead of writing every part of the end-to-end tests in Go, existing functionality is reused where possible. This means that the test results are still evaluated with BATS.

This is an area of improvement to address after this PR (see below).

### Event flow via signal

The end-to-end tests no longer rely on sleeps/timing to manage event flow. The integration tests signal the spawned process using Linux process signaling. Once signaled, the application knows it is being monitored and starts it processing.

### Kafka Go testing

The Kafka testing is unique. It needs to run a Kafka server to communicate with. This consolidates that process into the end-to-end testing application.

The Docker Go client is used to managed and run a Kafka container.

### Removed BATS verification of `traces.json`

This was talked about at the last SIG meeting and is accomplished in these changes. The BATS verification against a static JSON output is unstable. Instead of updating all `traces.json` to the lastest version of Go used in this change (and other spans added to the Kafka testing), the verification of `traces.json` is removed.

## Makefile

All `fixture` targets are removed from the `Makefile`.

The docker development target (`make docker-dev`) is updated. It now runs with the correct permissions and mounts to run integration tests.

The docker test target (`make docker-test`) is updated. It now runs with the correct permissions and mounts to run the full test suite, including integration tests.

## Kind Integration tests

The kind integration testing is removed along will all of the related artifacts.

## Follow-up items

- Compatibility testing. Now that these end-to-end tests are reasonably fast, Github actions can be added to test a matrix of supported versions.
  - Investigate running a unified Kafka server
- Replace BATS. There is no need to shell-out to evaluate the captured telemetry. This can be replaced with an entirely Go solution.


.github/workflows/checks.yml
```diff
@@ -65,13 +65,32 @@ jobs:
...
-          go-version: ${{ matrix.go-version }}
+          go-version: ${{ env.GO_VERSION }}
           check-latest: true
           cache-dependency-path: "**/go.sum"
       - name: Install build dependencies
         run: sudo apt-get update && sudo apt-get install -y clang llvm
       - name: Run tests
         run: make test-race
+  integration-test:
+    runs-on: ubuntu-latest
```
RonFed, line 76: We currently have coverage for amd and arm, so I think we should have the same runner matrix here.

.github/workflows/checks.yml
```diff
@@ -65,13 +65,32 @@ jobs:
...
+        with:
+          go-version: ${{ env.GO_VERSION }}
+          check-latest: true
+          cache-dependency-path: |
+            go.sum
+            **/go.sum
+      - name: Install dependencies
+        run: sudo apt-get update && sudo apt-get install -y clang llvm bats jq
+      - name: Initialize
+        run: make go-mod-tidy generate
```
RonFed, line 91: It is important to generate the object files in a known containerized environment to have consistent results. Doing it this way is
1. Different from objects we generate for the released image (those are generated on a Debian based container). 
2. Generating this way may lead to instability since we are depending on the GitHub runner environment and packages versions which might change.

RonFed (commented):
This is amazing.
Going to go over it more tomorrow, left a few comments regarding the CI.

.github/workflows/checks.yml
MrAlias: I'm not following. What are you suggesting?

We already run our other tests this way? Why is this different?

.github/workflows/checks.yml
MrAlias: Updated. PTAL.

.github/workflows/checks.yml
MrAlias: Are you suggesting that we run this from our Docker image?

If so, I can look into trying that out.

RonFed: The current e2e tests are building the `otel-go-instrumentation` image which has the generated files inside.
I'm suggesting to keep this behavior and build in a known containerized environment

pellared: @RonFed, do you think it makes sense to address this concern in a separate PR to not scope creep this PR and make the next one easier to review?

RonFed (commented):
I really like this.
1. Regarding the kafka case I'm wondering how it'll scale once we expand the tests to have more cases in the matrix (client version, go version etc') - we don't wan't to create the kafka broker for each case - so I think we need to have a way to register a "setup" function to our tests - this can be relevant also to the database/sql one and more future libraries.
2. Also relating to the first point - I think we can make use of https://github.com/testcontainers/testcontainers-go instead of directly using the docker api, I think that lib can provide some nice abstractions for us.

.github/workflows/checks.yml
RonFed: I think we can switch `make generate` with the existing `make docker-generate` so I don't think it would have an impact on the PR size.

MrAlias: `make docker-generate` will build the project docker container on each invocation. I'm not in favor this:

1. It is slow and ultimately unneeded.
2. This package is designed to be a library, not a delivered binary. It is already used by other projects in a way where they are already generating their own object files. This is the intended use-case. Testing the object files for the CLI is not the purpose of this test.

MrAlias:
> 1. Regarding the kafka case I'm wondering how it'll scale once we expand the tests to have more cases in the matrix (client version, go version etc') - we don't wan't to create the kafka broker for each case - so I think we need to have a way to register a "setup" function to our tests - this can be relevant also to the database/sql one and more future libraries.

I'm not seeing an issue running a broker per test given they will be run in parallel.

But more importantly, I do not see it being relevant to this PR. We can address that issue when we address running multiple of these tests in a matrix test. This is something out-of-scope for this PR and is already being tracked as a follow up task in the description. I'll be sure to include this feedback in the issue I create for that task when this PR merges.

.github/workflows/checks.yml
RonFed: it is needed for because of:

> Generating this way may lead to instability since we are depending on the GitHub runner environment and packages versions which might change

and it will also change the behavior of our current e2e testing

MrAlias:
> 2\. Also relating to the first point - I think we can make use of https://github.com/testcontainers/testcontainers-go instead of directly using the docker api, I think that lib can provide some nice abstractions for us.

I've tried switching to this but am not able to get this working. It looks to be due to my lack of knowledge about how to get the testcontainers to configure the network appropriately.

Given we have a working solution for the single kafka test case, and given the abstractions are not going to make a significant reduction of code here (my non-working prototype had 10s of less lines) , I'm going to keep the use of the Docker API the way it is.

It is something we can address in the future if we find a need or have the motivation to.

pellared:
> I think we can make use of https://github.com/testcontainers/testcontainers-go instead of directly using the docker api, I think that lib can provide some nice abstractions for us.

Based on experience I had so far (amount of issues, stability of API), I would rather look into https://github.com/ory/dockertest. We use it in https://github.com/signalfx/splunk-otel-go. But using Docker API seems also fine and I do not think it should block this PR.

.github/workflows/checks.yml
MrAlias: I can look into [running this job on a base image](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idcontainerimage).

However, I'm not convinced this is what we want. We are no longer focused on shipping a binary for this project. Instead, we are focused on providing this project as a library. It should work with whatever version of clang the user of this project is using. I expect they are using a variety of versions of the clang.

MrAlias: Updated to run on golang:1.24.1-bookworm@sha256:fa1a01d362a7b9df68b021d59a124d28cae6d99ebd1a876e3557c4dd092f1b1d

RonFed (approved):

Issue #2136, opened 2025-04-11
Add compatibility testing to CI


MrAlias:
Follow up to #2118

- [ ] Add matrix entries for stable and oldstable versions of Go
- [ ] Explore compatibility testing for different versions of the Go packages being instrumented
  - For the kafka-go testing, investigate running a unified Kafka server

### Prior Art

- https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/2099

daemon1024:
Hey @MrAlias I would like to take this up! 

I see that #2099 was closed due to
> The instability of our CI system needs to be addressed before we add additional test cases. The system is already very unstable and this will only make things worse.

Is the CI stable now? Is there a linked issue where we are discussing the stability of CI where I can pitch in?

PR #2138, opened 2025-04-11
Test multiple Go verions in integration-test

MrAlias:
Part of #2136

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -75,7 +75,12 @@ jobs:
   integration-test:
     strategy:
       matrix:
-        runner: [ubuntu-latest, ubuntu-22.04-arm]
+        runner: 
+          - ubuntu-latest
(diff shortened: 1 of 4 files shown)


RonFed:
Do you think we should add older versions? According to our compatibility doc we support go1.19 and above so maybe we should add go1.19 as well.

MrAlias:
> Do you think we should add older versions? According to our compatibility doc we support go1.19 and above so maybe we should add go1.19 as well.

That's going to be a bit harder. All of the project mod files require at least Go 1.23. We will not be able to compile the project with older versions.

To accomplish this type of testing we will need to update the integration tests to build the end-2-end test applications (i.e. `nethttp`, `nethttp_custom`) with different versions of Go.

Closing this as it is duplicating the verification that the project can be built with both versions of Go. A more comprehensive approach mentioned above is required instead.

Issue #2137, opened 2025-04-11
Replace BATS


MrAlias:
Follow up to #2118

Design and implement an evaluation testing framework that can be used to validate the OTLP received from the integration testing.

PR #2163, opened 2025-04-14
Replace BATS with Go testing for `autosdk` integration test

MrAlias:
- Add Go testing utilities to the `e2e` package
- Update the `autosdk` test to provide testing validation for new testing utilities
- Update integration test for `autosdk` to use new e2e testing utility
- Remove BATS definitions for `autosdk`

The plan is to do the same for all other e2e tests (replace BATS) once this merges. This is scoped to only `autosdk` to keep the size of this PR down.

Part of a resolution for #2137


MrAlias:
The integration tests need to be restructured (i.e. moved into `internal/test/e2e`) so there isn't a dependency on `e2e`.

MrAlias:
This needs to support tests like this: https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/f0f934fb672d5d060a9494fd4ba983861505f72d/internal/test/e2e/nethttp/verify.bats#L73-L77

MrAlias:
Replaced by #2178

PR #2169, opened 2025-04-15
Fail skipped integration tests

MrAlias:
Currently, the end-to-end tests are being skipped for the CI.

- Check the expected test is not skipped by the CI and fail if it is
- Fix the CI so the test is not skipped
- Fix the Kafka e2e test to work within the bridge network


RonFed (approved):
I saw the original error was:
```
    instrumentation_integration_test.go:38: cannot manage memory, skipping test.
--- SKIP: TestIntegration (0.00s)
```

Was it a permission error? I'm not sure what caused the test to get skipped and not throw an error?

I was looking for a way to tell go to fail on skip, but couldn't find any

MrAlias:
Yeah, from my investigation there isn't a way to tell Go to fail if test is skipped other than doing something similar to what is done here.

As for why the runner didn't have permissions, I was able to determine that by running this job in a container, that container didn't have elevated permissions like the base runner to run the test. Updating the container definition here to run in privileged mode fixed the issue.

PR #2178, opened 2025-04-17
Replace BATS with Go testing in `autosdk`

MrAlias:
- Add Go testing utilities to the `e2e` package
- Update integration test for `autosdk` to use new e2e testing utility
- Remove BATS definitions for `autosdk`

The plan is to do the same for all other e2e tests (replace BATS) once this merges. This is scoped to only `autosdk` to keep the size of this PR down.

Part of a resolution for #2137

Replaces #2163


grcevski (approved):
This looks good to me! LGTM!

PR #2206, opened 2025-04-24
Update database end-to-end program

MrAlias:
- Do not use HTTP. This is conflicting with other tests and is unneeded[^1].
  - Replace the HTTP parent span with one from the OTel global API.
- Run all DB queries after auto-instrumentation starts so there is not indeterminate results when the auto-instrumentation starts sooner than expected.
- Use error handling instead of panics.
- Update bats with new span indexes.

[^1]: https://github.com/open-telemetry/opentelemetry-go-instrumentation/actions/runs/14644532137/job/41095136897?pr=2178

### DIFF
--- a/internal/test/e2e/databasesql/cmd/main.go
+++ b/internal/test/e2e/databasesql/cmd/main.go
@@ -9,20 +9,17 @@ import (
 	"database/sql"
 	"flag"
 	"fmt"
-	"io"
 	"log/slog"
-	"net/http"
 	"os"
 	"os/signal"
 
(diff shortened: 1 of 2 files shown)


grcevski (commented):
I think keeping HTTP is useful to test that the SQL spans nest properly under the root HTTP server span.

MrAlias:
I've updated the description to capture this parent span switch.

grcevski (approved):
Ah cool, thanks I didn't catch that, it should be all the same then. Thanks for explaining it!

PR #2221, opened 2025-04-26
Replace BATS in nonrecording end-to-end test

MrAlias:
Part of a resolution for #2137.

Copies all the BATS test and adds test for the link and event on the child span.


internal/test/e2e/otelglobal/integration_test.go
```diff
@@ -7,9 +7,107 @@ package autosdk
...
+		e2e.AssertTraceID(t, parentSpan.TraceID(), "trace ID")
+		e2e.AssertSpanID(t, parentSpan.SpanID(), "span ID")
+
+		assert.Equal(t, ptrace.SpanKindServer, parentSpan.Kind(), "span kind")
+
+		status := parentSpan.Status()
+		assert.Equal(t, ptrace.StatusCodeOk, status.Code(), "status code")
+
+		attrs := e2e.AttributesMap(parentSpan.Attributes())
+		assert.Equal(t, int64(42), attrs["int_key"].Int(), "int_key")
```
pellared, line 60: Won't this panic if `attrs["int_key"]` does not exist?

internal/test/e2e/otelglobal/integration_test.go
```diff
@@ -7,9 +7,107 @@ package autosdk
...
+		assert.Equal(t, "forty-two", attrs["string_key"].Str(), "string_key")
+		assert.True(t, attrs["bool_key"].Bool(), "bool_key")
+		assert.Equal(t, 42.3, attrs["float_key"].Double(), "float_key")
+		assert.Equal(t, 42.3, attrs["float_key"].Double(), "float_key")
```
pellared, line 64:
```suggestion
```

pellared (changes requested):
I noticed some things that are probably good to improve after second look. Sorry.

internal/test/e2e/otelglobal/integration_test.go
MrAlias: It will, but this is also a test. The panic is validating the existence, just will poor representations of the source.

I'm fine addressing this, but would like to do so holistically for all end-to-end test once we get these migrated.

MrAlias: https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/2229

pellared (approved):

PR #2259, opened 2025-05-01
Add getlog package for leveled logs in end-to-end tests

MrAlias:
We should not output end-to-end application logs when tests are not run in verbose mode. This changes the gRPC end-to-end test to support this behavior and act as a template for the rest of the end-to-end tests.

The added `getlog` package is similar to the `trigger` package. It provides a `flag.Var` end-to-end test applications can use. The `Flag` type is configured with a logging level via a CLI flag or the `OTEL_LOG_LEVEL` environment variable. From this, a test can used the configured `*slog.Logger` to log information about the application.

The integration test runner is updated set OTEL_LOG_LEVEL to `debug` when the test is run in verbose mode.


grcevski (approved):
LGTM!

PR #2262, opened 2025-05-01
Fix gRPC end-to-end test

MrAlias:
The Heritage test for span relationship depends on the spans being received in deterministic order. This order is not guarnteed.

Instead, test based on the expected trace IDs for deterministic validation.

Example test failure:

- https://github.com/open-telemetry/opentelemetry-go-instrumentation/actions/runs/14759596480/job/41436726864?pr=2253


grcevski (approved):
Sweet!

PR #2271, opened 2025-05-03
Add a make target for compile-commands.json

MrAlias:
Use [bear](https://github.com/rizsotto/Bear) to generate compile_commands.json. This is used for editor's LSPs (i.e. clangd).

The compile-commands.json file itself is going to contain local absolute paths. Therefore, add a make target to generate it locally for each developer.


grcevski (approved):
LGTM!

RonFed:
can you please give some more context about this? what are we planning to use it for?

RonFed:
Got it. Do you think it's worth adding a mention to this in the contributing doc under `Development`?

MrAlias:
Sounds like a good idea to me. I'll add a note.

RonFed (approved):

PR #2272, opened 2025-05-04
 telemetry: add tests for trace id and span id encoding

yumosx:
add tests for trace id and span id encoding.


sdk/internal/telemetry/id_test.go
```diff
@@ -0,0 +1,52 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package telemetry
+
+import "testing"
+
+func TestTraceIdEncoding(t *testing.T) {
```
MrAlias, line 8:
```suggestion
func TestTraceIDEncoding(t *testing.T) {
```

sdk/internal/telemetry/id_test.go
```diff
@@ -0,0 +1,52 @@
...
+			Expect: []byte(`""`),
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.Name, runJSONEncodingTests(tc.Input, tc.Expect))
+	}
+}
+
+func TestSpanIdEncoding(t *testing.T) {
```
MrAlias, line 31:
```suggestion
func TestSpanIDEncoding(t *testing.T) {
```

sdk/internal/telemetry/id_test.go
```diff
@@ -0,0 +1,52 @@
...
+		},
+		{
+			Name:   "empty span id",
+			Input:  SpanID{},
+			Expect: []byte(`""`),
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.Name, runJSONUnmarshalTest(tc.Input, tc.Expect))
```
MrAlias, line 50: Why test only unmarshalling here instead of marshaling as well?

sdk/internal/telemetry/id_test.go
MrAlias, line 1: This is missing tests for the errors cases where the IDs are invalid lengths or they do not contain valid hex encoded data.

sdk/internal/telemetry/id_test.go
yumosx: Sorry, my mistake. I didn't check this carefully. Thank you for pointing it out.

sdk/internal/telemetry/id_test.go
MrAlias: @yumosx are you able to address the errors testing? Is this something you could add to this PR?

sdk/internal/telemetry/id_test.go
```diff
@@ -0,0 +1,114 @@
...
+			Input:     []byte(`"234567890abcdefgh1234567890abcdef"`),
+			ExpectErr: errors.New("cannot unmarshal ID from string"),
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.Name, func(t *testing.T) {
+			var id TraceID
+			err := json.Unmarshal(tc.Input, &id)
+			require.Error(t, err)
```
MrAlias, line 59: This is unneeded give we test the error on the next line.


```suggestion
```

The same suggestion applies below to the other test cases.

sdk/internal/telemetry/id_test.go
```diff
@@ -0,0 +1,114 @@
...
+			require.ErrorContains(t, err, tc.ExpectErr.Error())
```
MrAlias, line 60: This should use [`assert`](https://pkg.go.dev/github.com/stretchr/testify/assert), not `require`. More specifically, it can use [`ErrorIs`](https://pkg.go.dev/github.com/stretchr/testify/assert#ErrorIs).


```suggestion
			assert.ErrorIs(t, err, tc.ExpectErr)
```

Similar below.

yumosx: we can define the expected error message strings directly and use assert.ErrorContains:

like this:
```go
assert.ErrorContains(t, err, tc.ExpectErrMessage)
```

PR #2383, opened 2025-05-28
add distro version and name by default to otelsdk handler

RonFed:
Adding our distro version and name as resource attributes to the `otelsdk` handler.
This requires moving the version constant to a new location that can be consumed both by the handler and by the auto package and to avoid cyclic dependency.


damemi (approved):

MrAlias:
Can we update this in a way that [multimod](https://pkg.go.dev/go.opentelemetry.io/build-tools/multimod) will handle the update of the version during a release?

MrAlias:
> Can we update this in a way that [multimod](https://pkg.go.dev/go.opentelemetry.io/build-tools/multimod) will handle the update of the version during a release?

https://github.com/open-telemetry/opentelemetry-go-build-tools/blob/9733cfe3461f2d064cac4449103713fba47c775c/multimod/internal/prerelease/prerelease.go#L133-L187

MrAlias:
> > Can we update this in a way that [multimod](https://pkg.go.dev/go.opentelemetry.io/build-tools/multimod) will handle the update of the version during a release?
> 
> https://github.com/open-telemetry/opentelemetry-go-build-tools/blob/9733cfe3461f2d064cac4449103713fba47c775c/multimod/internal/prerelease/prerelease.go#L133-L187

https://github.com/open-telemetry/opentelemetry-go-build-tools/issues/994

RonFed:
@MrAlias do you think this is blocked by the multimod feature. or we can merge this and update the multimod config once that feature is merged?

MrAlias:
It is blocked. It will not currently work with our releasing process.

MrAlias:
The `version.go` file can now be updated to use the [new multimod version ref syntax](https://github.com/open-telemetry/opentelemetry-go-build-tools/pull/1002). Example: https://github.com/open-telemetry/opentelemetry-go-build-tools/blob/afc9eb5e8cc2fc3f1427d890a3fe99bc06a92861/multimod/docs/versions-example.yaml#L62-L66

RonFed:
@MrAlias updated this with the new version format, PTAL. Tested locally with the `prerelease` command and verified it worked correctly.

looks like a new go version was just released so the offsets check is failing.

CHANGELOG.md
```diff
@@ -27,6 +27,10 @@ OpenTelemetry Go Automatic Instrumentation adheres to [Semantic Versioning](http
   This fixes the bug where images for alternate architectures (e.g. `arm64`) were built using the `amd64` architecture. ([#2411](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/2411))
 - Do not fail run when a module has a version of `(devel)`. ([#2437](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/2437))
 
+# Fixed
+
+- Add `telemetry.distro.version` resource attribute to the `otelsdk` handler. ([#2383](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/2383))
+
```
MrAlias, line 33: This needs to be moved to the unreleased section.

internal/pkg/instrumentation/version.go
```diff
@@ -0,0 +1,13 @@
...
+
+// Package instrumentation provides functionality to manage instrumentation
+// using eBPF for Go programs.
+package instrumentation
+
+const (
+	// DistributionName is used for `telemetry.distro.name` resource attribute.
+	DistributionName = "opentelemetry-go-instrumentation"
+	// DistributionVersion is the current release version of OpenTelemetry Go auto-instrumentation in use.
+	DistributionVersion = "v0.22.1"
```
MrAlias, line 12: What do you think about dropping the "Distribution" part of the const names?


```suggestion
	// Name is used for `telemetry.distro.name` resource attribute.
	Name = "opentelemetry-go-instrumentation"
	// Version is the current release version of OpenTelemetry Go auto-instrumentation in use.
	Version = "v0.22.1"
```

This will change use from `instrumentation.DistributionName` and `instrumentation.DistributionVersion` to `instrumentation.Name` and `instrumentation.Version` which read a bit better and are shorter to type.

MrAlias (approved):

Issue #2434, opened 2025-06-19
Support zero-code configuration of resource detectors


ezhang6811:
The AWS Go SDK currently supports several [resource detectors](https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/detectors) for third party computing environments. It would be useful if customers could enable select detectors in the agent's TracerProvider, via an environment variable, to automatically add environment-specific attributes to the telemetry.

PHP instrumentation has a similar mechanism, see [here](https://github.com/open-telemetry/opentelemetry-php/pull/600).

**Describe the solution you'd like**

When auto-instrumentation is started it determines what resource detector to run based on configuration it is provided (i.e environment variables, configuration file).

To accomplish this, the [`autodetect`](https://github.com/open-telemetry/opentelemetry-go-contrib/tree/53949dd7bf6f273f3325c17ecb431f63f87aa3e3/detectors/autodetect) package should be used.

**Describe alternatives you've considered**

- Write our own `autodetect` package (i.e. https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/2438)

MrAlias:
Copied from [duplicate #2492](https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/2492#issuecomment-3049550558).

For environment variable configuration, if we plan to support this method of configuration, the variable name will need to be prefixed with `OTEL_GO_AUTO_`. This matches our other [configuration options](https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/50a97b86eb506b426bc23de567ffbbe3cf909a28/docs/configuration.md) for things that are not defined in the specification.

MrAlias:
Copied from [duplicate #2492](https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/2492#issuecomment-3049550558).

Supporting a configuration file for this will mean we need to integration the [OpenTelemetry declarative configuration](https://github.com/open-telemetry/opentelemetry-configuration). This is not stable yet, and [support is Go](https://pkg.go.dev/go.opentelemetry.io/contrib/otelconf@v0.17.0/v0.3.0) is [an ongoing work-in-progres](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/7268).

I propose we do not attempt to solve this problem when resolving this issue. Meaning we stick to just supporting environment variables.

PR #2438, opened 2025-06-21
feat: support resource detector auto-configuration with OTEL_GO_DETECTORS env var

ezhang6811:
This PR allows AWS resource detectors to be initialized upon startup via the OTEL_GO_DETECTORS environment variable. The known values are "all", "ec2", "ecs", "eks", and "lambda". See #2434 for more context.

Order of priority is similar to that of OTEL_PHP_DETECTORS variable in [PHP](https://github.com/open-telemetry/opentelemetry-php/pull/600). If "all" is present, return all detectors. Else, iterate through the list and add instances of recognized detectors. Unrecognized or empty values will be ignored.


MrAlias:
This PR introduces an unfair bias to support only AWS resource detectors. I do not think it should be accepted as proposed.

Why not just use the [`WithResourceAttributes`](https://pkg.go.dev/go.opentelemetry.io/auto@v0.22.0/pipeline/otelsdk#WithResourceAttributes) in your setup code that supplies attributes from any detector you want to use?

ezhang6811:
I'm not sure that WithResourceAttributes covers this use case, as it doesn't support dynamic resource detection and only adds static key-values, which should already be covered by OTEL_RESOURCE_ATTRIBUTES.

Our goal is to provide AWS customers running Go applications with a tracing experience similar to that of other languages which already support auto resource detection out of the box. Other vendors are welcome to add their own resource detectors on top of this implementation.

Would it help remove the "all" option as to not suggest that AWS resource detectors are exclusively covered by this env var?

MrAlias:
> I'm not sure that WithResourceAttributes covers this use case, as it doesn't support dynamic resource detection and only adds static key-values, which should already be covered by OTEL_RESOURCE_ATTRIBUTES.


Resource detection is done once and the attributes are static after `Detect` returns. Why doesn't this act the same way?

```
// detector is a composite resource.Detector defined by the user.
attr, err := detector.Detect(ctx)
// handle err
otelsdk.NewHandler(ctx, otelsdk.WithResourceAttributes(attrs...))
```

ezhang6811:
That's true, but why would the user initialize their own TraceHandler if they want to use the autoinstrumentation? Writing application code in order to manage resource detection defeats the purpose of zero-code instrumentation.

MrAlias:
I see. If you want to add a solution to the default that is shipped here it needs to follow the OpenTelemetry tenet of being vendor neutral.

If you would like to provide only a AWS focused solution, feel free to build your own distribution on top of this project as that is its intended design.

ezhang6811:
Thank you for your feedback! After discussing with my team, we are hesitant to build our own AWS distro for Go instrumentation that we'd then have to maintain and keep up to date with this one, just to support this brief code change.

I would still contend that this change is just integration code to enable zero code instrumentation for our customers, as the actual implementation for the AWS resource detectors already exists separately in the opentelemetry-go-contrib repo. I understand that it could be more modularized though. PHP uses a [registerResourceDetector](https://github.com/open-telemetry/opentelemetry-php/blob/33715fddc1c70c883a49d1217482f72106a57974/src/SDK/Registry.php#L122) function that's called by third party resource detectors to be automatically added to the list of recognized detectors. This option seems ideal for both sides but would still require changes in the core repo that aren't part of your current plans.

End of the day, we want our customers to be able to use this feature while not having to maintain our own copy of a package that's rapidly gaining new features. Let me know if you have any other ideas for an approach if the current change doesn't align with your goals (even if its temporary). If not, we will fall back to the forking option.

MrAlias:
See https://github.com/open-telemetry/opentelemetry-go-contrib/pull/7522

ezhang6811:
Thank you for contributing this! If accepted, it should fulfill my proposed use case for autoinstrumentation as well. We can close this PR and address further follow ups in that PR.

MrAlias:
I'm going to close this PR. We can discuss the proposal for how to solve this issue in the [tracking issue](https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/2492). When we have agreement there, we can update/re-open this or open a new PR to resolve that.

PR #2648, opened 2025-08-13
chore: enable rangeValCopy rule from go-critic

mmorel-35:
#### Description

Enables and fixes [rangeValCopy](https://go-critic.com/overview.html#rangevalcopy) rule from go-critic


MrAlias:
Unrelated offsets CI failure.

PR #2684, opened 2025-08-19
Introduce clang-format

### LINKED ISSUE #2288: Standardize our C formatting
Use [`clang-format`](https://clang.llvm.org/docs/ClangFormat.html) to ensure a consistent standard formatting across the project.

TODO:

- Determine what standard to apply
- Add tooling/docs for developers to check locally
- Add CI workflows to validate code against the standard adopted

rafaelroquetto:
Resolve #2288

Introduce clang-format to ensure all C code complies to the same formatting.

Namely, this PR brings:
 - a default `.clang-format` configuration (same as OBI)
 - a `Makefile` target - `clang-format` that will format all files
 - a convenience pre-commit hook that can be installed with `make install-hooks` - once installed, it will take care of formatting any C source code according to `.clang-format` before finalising the commit
 - a github workflow that will nag you if your C files are not properly formatted 

Note: I've formatted all .C files in the tree accordingly. We might want to discuss that before this PR is merged.


rafaelroquetto:
Looks like the clang-format workflow is working ;)

rafaelroquetto:
@MrAlias should this be part of the changelog?

MrAlias:
> @MrAlias should this be part of the changelog?

I don't think so. I've update with the "Skip Changelog" label.

MrAlias (approved):

rafaelroquetto:
@MrAlias I did some research, and I think it will be difficult to ditch the explicit  reference to clang-format-19, at least for now. If we want to be 100% consistent, we have to lock to a version because there were/are bugs affecting different major clang-format versions when it comes to the parsing of macros that are impossible to workaround only by manipulating the configuration file.

For instance: https://github.com/llvm/llvm-project/issues/100304

It looks like something that has been fixed on clang-format-20 and this is why we see different outcomes.

PR #2802, opened 2025-09-12
Fix 386 architectures

dmathieu:
See https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/2801

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ OpenTelemetry Go Automatic Instrumentation adheres to [Semantic Versioning](http
 
 ## [Unreleased]
 
+### Fixed
+
+- Fix compilation errors on 386 architecture. ([#2802](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/2802))
+
 ## [v0.23.0/v1.2.0] - 2025-09-10
 
 <!-- markdownlint-disable MD028 -->


dmathieu:
CI is confirmed passing on 386, when run with the commits on https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/2801
https://github.com/open-telemetry/opentelemetry-go-instrumentation/actions/runs/17669756731/job/50218699157

dmathieu:
That's something that looks like would need additional tests. How about doing it in another PR? (there could be an issue tracking that work)

grcevski:
> That's something that looks like would need additional tests. How about doing it in another PR? (there could be an issue tracking that work)

Yeah that would work, I'm just surprised the tests pass ...

We do this all over the place:

```
SEC("uprobe/Transport_roundTrip")
int uprobe_Transport_roundTrip(struct pt_regs *ctx) {
    u64 request_pos = 2;
    void *req_ptr = get_argument(ctx, request_pos);
```

So, this last line ` void *req_ptr = get_argument(ctx, request_pos); `, that's reading `#define GO_PARAM2(x) ((x)->bx)`. If it works, I'm assuming that the docs google has are outdated and Go >= 1.17 on 386 uses register based linkage, just like amd64.

The problem is the code in the Go SDK Global handling, which does ` void *span_context_ptr_val = get_argument(ctx, 7); `, that will pull register R9 `#define GO_PARAM7(x) ((x)->r9)` which doesn't exist on 386.

dmathieu:
@MrAlias I introduced the 386 tests in a separate PR: https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/2801

MrAlias:
> If the goal here is to satisfy the failing CI I think we should add stubs for `other` archs similar to what we've done for non linux systems. These changes seems to suggest we would support i386 arch where in practice it requires a more broad change in the eBPF side. I don't think we should make that kind of change right now if there is no requirement for it from users - I think we should go with the stubs approach for now.

Agreed.

We should look into just testing the `sdk` package and provide support there. Adding support for the `auto` package seems ambitious and unmotivated.

dmathieu:
Works for me. I'll let you handle it so the otel-go project can upgrade future versions.

PR #2981, opened 2025-10-30
fix: grpc client call panic

quanbisen:
I used Cursor to fix a bug in the gRPC client call that occasionally triggered panics. After the fix, it has been running stably for 3 days without any issues. Before the fix, the program would panic within a day. While this fix resolved the gRPC panic issue, I'm unsure if the changes might affect other functionalities – though my current tests show no problems.


RonFed (commented):
I'm not sure I understand the rationale for this fix. It seems like this changes some local variables to be stored on storage maps but these maps are accessed within the same function - so I'm not sure how that helps.

quanbisen:
> I'm not sure I understand the rationale for this fix. It seems like this changes some local variables to be stored on storage maps but these maps are accessed within the same function - so I'm not sure how that helps.

but it work for me now, I'm sure it fix grpc call panic.

internal/pkg/instrumentation/bpf/google.golang.org/grpc/client/bpf/probe.bpf.c
```diff
@@ -180,24 +196,48 @@ int uprobe_LoopyWriter_HeaderHandler(struct pt_regs *ctx) {
...
+    struct hpack_header_field hf = {
+        .name = {
+            .str = key_ptr,
+            .len = W3C_KEY_LENGTH,
+        },
+        .value = {
+            .str = val_ptr,
+            .len = SPAN_CONTEXT_STRING_SIZE,
+        },
+        .sensitive = false,
```
quanbisen: maybe this work`.sensitive = false,`

PR #2987, opened 2025-10-30
Update offsets workflow PR creation

MrAlias:
[The current workflow is failing](https://github.com/open-telemetry/opentelemetry-go-instrumentation/actions/runs/18952132159/job/54130033027). Fix by using [a similar approach to `obi`](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/6fc7f273771d415758743a0bfe34ef5e9fbcd375/.github/workflows/update-offsets.yml) and use the `gh` utility.

### DIFF
--- a/internal/tools/go.mod
+++ b/internal/tools/go.mod
@@ -5,7 +5,7 @@ go 1.24.0
 require (
 	github.com/Masterminds/semver/v3 v3.4.0
 	github.com/docker/docker v28.5.1+incompatible
-	github.com/golangci/golangci-lint/v2 v2.5.0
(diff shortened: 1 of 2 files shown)


MrAlias:
> Looks good! Do we need to run a make command before the commit? Seems like the CI failed again.

I'll have to run the workflow after this is merged to fix the CI. With this change though, it should be able to push a new PR to resolve the issue.

grcevski:
> > Looks good! Do we need to run a make command before the commit? Seems like the CI failed again.
> 
> I'll have to run the workflow after this is merged to fix the CI. With this change though, it should be able to push a new PR to resolve the issue.

Sounds good!

PR #3472, opened 2026-04-27
Run clang-format inside a pinned docker image

### LINKED ISSUE #2774: Run clang-format in a docker image
Update the makefile to run clang-format with a consistent version by using a docker image.

iyiola-dev:
Update the Makefile clang-format target to invoke clang-format inside a Docker image pinned in dependencies.Dockerfile
Resolves #2774

### DIFF
--- a/.github/workflows/clang-format-check.yml
+++ b/.github/workflows/clang-format-check.yml
@@ -19,9 +19,6 @@ jobs:
           fetch-depth: 0
           persist-credentials: false
 
-      - name: Install Clang-Format
-        run: sudo apt-get install clang-format-19
-
       - name: Check Clang Format
         run: |
           # Check for modified C/C++ files
@@ -31,8 +28,12 @@ jobs:
             exit 0
           fi
 
-          # Run clang-format and check for changes
-          clang-format-19 -i $files
+          # Run clang-format from the docker image pinned in dependencies.Dockerfile
+          # to keep the version consistent with `make clang-format` and the
+          # local pre-commit hook.
+          image=$(awk '$4=="clang-format" {print $2}' dependencies.Dockerfile)
+          docker run --rm -u "$(id -u):$(id -g)" -v "$PWD:/workdir" -w /workdir "$image" \
+            clang-format -i $files
           if ! git diff --exit-code $files; then
             echo "Error: Some files are not formatted correctly. Please run clang-format on the modified files or use the provided pre-commit hook."
             exit 1
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,7 +57,7 @@ OpenTelemetry Go Automatic Instrumentation adheres to [Semantic Versioning](http
 
 ### Changed
 
-- The `clang-format` `make` target now runs inside a pinned Docker image to ensure a consistent `clang-format` version across contributors and CI. ([#3472](https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/3472))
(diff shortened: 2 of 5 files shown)


MrAlias (commented):
This looks like a useful improvement. Moving `make clang-format` to a pinned Docker image makes that workflow more reproducible and removes one source of local version drift.

One thing I’d want to call out is scope: this change standardizes the `make clang-format` path, but it doesn’t yet bring the other formatter entry points onto the same toolchain. The CI check still installs and runs `clang-format-19` directly, and the pre-commit hook still calls the host `clang-format`, so we can still get different behavior depending on how formatting is run.

I don’t think that blocks this PR if the goal here is specifically to fix the Makefile workflow. But I do think it would help to leave a note about expected follow-up, so readers don’t assume this change fully unifies formatting across local development and CI.

iyiola-dev:
> This looks like a useful improvement. Moving `make clang-format` to a pinned Docker image makes that workflow more reproducible and removes one source of local version drift.
> 
> One thing I’d want to call out is scope: this change standardizes the `make clang-format` path, but it doesn’t yet bring the other formatter entry points onto the same toolchain. The CI check still installs and runs `clang-format-19` directly, and the pre-commit hook still calls the host `clang-format`, so we can still get different behavior depending on how formatting is run.
> 
> I don’t think that blocks this PR if the goal here is specifically to fix the Makefile workflow. But I do think it would help to leave a note about expected follow-up, so readers don’t assume this change fully unifies formatting across local development and CI.

Hi, will extend the scope of this to the ci and the pre-commit hook

iyiola-dev:
hi @MrAlias extended the PR to dockerize the remaining two entry points so all three (`make clang-format`, the `clang-format-check` workflow, and [hooks/pre-commit](cci:7://file:///Users/iyiolaakanbi/projects/opentelemetry-go-instrumentation/hooks/pre-commit:0:0-0:0)) now run from the same image pinned in [dependencies.Dockerfile](cci:7://file:///Users/iyiolaakanbi/projects/opentelemetry-go-instrumentation/dependencies.Dockerfile:0:0-0:0):
 
- [.github/workflows/clang-format-check.yml](cci:7://file:///Users/iyiolaakanbi/projects/opentelemetry-go-instrumentation/.github/workflows/clang-format-check.yml:0:0-0:0) drops the `apt-get install clang-format-19` step and calls `docker run` against the pinned image. The image reference is read from [dependencies.Dockerfile](cci:7://file:///Users/iyiolaakanbi/projects/opentelemetry-go-instrumentation/dependencies.Dockerfile:0:0-0:0) via the same `awk` lookup the Makefile uses, so Renovate updates all three entry points from the single image declaration.
- [hooks/pre-commit](cci:7://file:///Users/iyiolaakanbi/projects/opentelemetry-go-instrumentation/hooks/pre-commit:0:0-0:0) resolves the repo root via `git rev-parse --show-toplevel` (so it works from subdirectories) and formats all staged files in a single `docker run` invocation.
 
Updated the CHANGELOG entry to reflect the broader scope. Verified locally:
- The `docker run ... clang-format --version` call returns `19.1.7`, matching the `clang-format-19` previously installed by the workflow.
- The pre-commit hook correctly picks up staged `.c`/`.h` files, runs clang-format inside the container, and re-stages the formatted files.

iyiola-dev:
hi @MrAlias do you have any review here yet for this?

iyiola-dev:
@MrAlias please any review for this?

iyiola-dev:
hello @MrAlias please check this. is this still valid or it needs to be closed?

Issue #640, opened 2024-02-01
Reduce pointers usage in TargetDetails


RonFed:
[TargetDetails](https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/main/internal/pkg/process/analyze.go#L29) seems to use pointers without a good reason, and it might cause out-of-memory error seen in #619. For large binaries analysis, the slices and maps used can get large, and removing the pointers might be an improvement.

iblancasa:
Hi @RonFed. I'm trying to instrument a big executable running in Kubernetes and I got OOM Killed all the time in the `opentelemetry-go-instrumentation` container. I increased the resource limits a lot but it seems it is not enough. I think I'm facing this issue (I have been doing some debugging).

I would like to contribute but I need some hints. Can you provide some light? Thanks!

RonFed:
Hey @iblancasa, thank you for your interest. 
What is approximately the memory limit you saw exceeded?
This is an interesting topic, and I'd start by profiling memory usage (pprof)  in a local setup to get the root cause.
The TargetDetails struct looked like a good candidate for the problem but I didn't get the chance to confirm that.
Another place that might be relevant is the `structfield` package which stores an offset mapping of relevant structs for instrumentation.

iblancasa:
> What is approximately the memory limit you saw exceeded?

I was trying to do some experiments with the OpenTelemetry Operator to autoinstrument an OpenTelemetry Collector. So... I added 2Gb to the pod as limits and it is `OOMKilled`. I reduced the size of my collector reducing the number of components and I was able to execute some extra statements in the instrumentation but I was not able to load the probes https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/9882b86f52d8daf168efee68ddc4442d2acd821f/internal/pkg/instrumentation/manager.go#L207-L214

After reading the comments, I think the issue you described here can be related.

> Another place that might be relevant is the structfield package which stores an offset mapping of relevant structs for instrumentation.

I agree. But I have been printing the memory usage until reaching these lines:
https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/9882b86f52d8daf168efee68ddc4442d2acd821f/internal/pkg/instrumentation/manager.go#L207-L214
And it is around 25MB. After the load is done, the pod is killed by Kubernetes.

RonFed:
[...]

PR #3479, opened 2026-06-04
process: reduce pointer usage in Info to lower GC pressure

lazureykis:
## Summary

Replaces pointer-heavy fields in `process.Info` with value types to reduce heap object count and GC scan work for large instrumented binaries.

**Changes to `Info`:**
| Field | Before | After |
|-------|--------|-------|
| `Functions` | `[]*binary.Func` | `[]binary.Func` |
| `GoVersion` | `*semver.Version` | `semver.Version` |
| `Modules` | `map[string]*semver.Version` | `map[string]semver.Version` |

Also fixes two non-Linux build stub mismatches found during this work:

- `bpffs_other.go`: stale `TargetDetails` type reference → `Info`
- `cpu_other.go`: `cpuCount()` returned `int` instead of `uint64`

### DIFF
--- a/internal/pkg/instrumentation/bpffs/bpfsfs_other.go
+++ b/internal/pkg/instrumentation/bpffs/bpfsfs_other.go
@@ -9,14 +9,14 @@ import "go.opentelemetry.io/auto/internal/pkg/process"
 
 // Stubs for non-linux systems
 
-func PathForTargetApplication(target *process.TargetDetails) string {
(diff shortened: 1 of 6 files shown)


lazureykis:
Closing — after reviewing the issue history, this change doesn't meaningfully address #640. The OOM occurs during BPF probe loading (kernel-side), not during Info construction. The pointer reduction here saves ~50 allocations for a slice that has 20-50 elements in practice — negligible impact.

PR #3592, opened 2026-06-30
Add deprecation notice

damemi:
As discussed in the SIG meeting on [6/4/2026](google.com/url?q=https://docs.google.com/document/d/1P6am_r_cxCX1HcpDQlznrTrTOvwN2whshL0f58lXSWI/edit?usp%3Dsharing&sa=D&source=calendar&ust=1782822782137318&usg=AOvVaw36IpV9f9uG9FUmIOMXaarn), this project has entered a sunset mode where we plan to migrate maintainer effort and community focus toward OBI.

This adds a notice to the readme and encourages any contributions toward the migration, but as discussed it makes no guarantees for migration to be considered in the current OBI roadmap.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,9 +2,23 @@
 
 [![PkgGoDev](https://pkg.go.dev/badge/go.opentelemetry.io/auto)](https://pkg.go.dev/go.opentelemetry.io/auto)
 
-This repository provides [OpenTelemetry] tracing instrumentation for [Go] libraries using [eBPF].
+> [!WARNING]
+> **This project is deprecated**
+>
+> **OpenTelemetry Go Automatic Instrumentation is no longer actively maintained.**
+> End users and vendors should migrate to [OpenTelemetry eBPF Instrumentation (OBI)](https://opentelemetry.io/docs/zero-code/obi), the upstream OpenTelemetry project for zero-code, eBPF-based application observability.
+>
+> **Migrate to OBI**
+>
+> - [Documentation](https://opentelemetry.io/docs/zero-code/obi)
+> - [GitHub repository](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation)
+> - [CNCF Slack `#otel-ebpf-instrumentation`](https://cloud-native.slack.com/archives/C06DQ7S2YEP) ([join CNCF Slack](https://slack.cncf.io))
+>
+> If you are migrating from this project, we encourage you to share feedback, opinions, and feature gaps in the OBI [GitHub Discussions](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/discussions) or on Slack so they can inform the OBI roadmap.
+> Due to limited maintainer bandwidth, there are **no official plans to provide zero-change migrations** from this project to OBI,
+> but any contributions toward this goal are welcome.
 
-:construction: This project is currently work in progress.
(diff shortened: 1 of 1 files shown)


MrAlias (commented):
Looks great! Can we add this to `go.mod` as well to [integrate with Go tooling](https://go.dev/ref/mod#go-mod-file-module-deprecation)?

README.md
```diff
@@ -2,9 +2,23 @@
 
 [![PkgGoDev](https://pkg.go.dev/badge/go.opentelemetry.io/auto)](https://pkg.go.dev/go.opentelemetry.io/auto)
 
-This repository provides [OpenTelemetry] tracing instrumentation for [Go] libraries using [eBPF].
+> [!WARNING]
+> **This project is deprecated**
```
jpkrohling, line 6: This project was merged with OpenTelemetry eBPF Instrumentation (OBI).

README.md
```diff
@@ -2,9 +2,23 @@
...
+>
+> **OpenTelemetry Go Automatic Instrumentation is no longer actively maintained.**
```
jpkrohling, line 8: OpenTelemetry Go Automatic Instrumentation is being merged with OBI and this repository is no longer actively maintained.

README.md
```diff
@@ -2,9 +2,23 @@
...
+> End users and vendors should migrate to [OpenTelemetry eBPF Instrumentation (OBI)](https://opentelemetry.io/docs/zero-code/obi), the upstream OpenTelemetry project for zero-code, eBPF-based application observability.
```
jpkrohling, line 9: Current users of OTel Go Automatic Instrumentation can expect the current binaries to work, but should consider moving to OBI for continued support and maintenance. OBI has been enhanced to cover all use-cases you know and love from Go Automatic Instrumentation. If you are facing issues migrating, open an issue, describing your scenario, what you are doing, and the expected outcome. (add a link to "issue")

README.md
```diff
@@ -2,9 +2,23 @@
...
+>
+> **Migrate to OBI**
```
jpkrohling, line 11: Is there a specific migration doc for Go Auto -> OBI ?

NameHaibinZhang:
Consolidating the eBPF-based application monitoring projects together is a great decision. Will the Profiling project also be merged into OBI down the road?

README.md
damemi: I think one could split hairs on the semantics vs technical detail of if this was "merged" with OBI. To me, this project really wasn't merged and merging implies that there will be transition/continuity, which is explicitly not the case.

On the other hand, from a user's perspective "deprecated" seems like kind of a dead end whereas "merged" more gracefully points to the replacement. So I could go either way. what would you recommend from the community perspective @jpkrohling ?

README.md
damemi: No, it was specifically a non-goal of this deprecation to provide migration paths from go-auto to obi, at least in terms of technical functionality. I think a simple migration doc would be reasonable though to show comparisons for things that do exist, wdyt @MrAlias ?

MrAlias: I'm not opposed if somebody wants to take that task. 👍

README.md
jpkrohling: I'm fine stating that it's deprecated, and I can see arguments for "merged" as well. I'd go with what the community (maintainers) feels most comfortable with, while still providing guidance to current users on what they should be looking into instead.

